Issue #111, opened 2013-08-09
Click should only fire when mousedown/mouseup occur on the same item


arvind:
Register these event listeners on the simple bar chart examples:
```javascript
ved.view.on('mousedown', function(e, i) { console.log('mousedown', i.key); });
ved.view.on('mouseup', function(e, i) { console.log('mouseup', i.key); });
ved.view.on('click', function(e, i) { console.log('click', i.key); });
```
Mousedown on one bar, move over to another bar and mouseup. See that all three events fire. With the DOM, the click would only fire if mousedown/mouseup occurred on the same element. 

Although, semantically, we're registering the event listeners on the entire view, they don't fire unless they occur over a specific item. So, in effect, events are implicitly registered on each individual item in the scene, and thus the click should follow convention.

jheer:
Fixed in 4476e45e9c3f1d516963e3ede855ef9d321b096c.

PR #206, opened 2014-08-06
Fix for #111

arvind:
This brings click handling in line with the DOM: http://api.jquery.com/mouseup/

> If the user clicks outside an element, drags onto it, and releases the button, this is still counted as a mouseup event. This sequence of actions is not treated as a button press in most user interfaces, so it is usually better to use the click event unless we know that the mouseup event is preferable for a particular situation.


jheer:
Thanks! I've incorporated your fix in 4476e45e9c3f1d516963e3ede855ef9d321b096c.

PR #225, opened 2014-12-09
Implement service to render SVG and PNG from vega specification

domoritz:
To run the server, use `node bin/server`.

To use the server, run `url -v -X POST -d @./examples/spec/bar.json http://localhost:8888/vg2png`.

Currently, the data has to be included in the request.


domoritz:
Implemented in separate repository at https://github.com/uwdata/vegaserver.

PR #227, opened 2014-12-28
use mutator with formula transform, add d3.format to formula list.

kanitw:


jheer:
Thanks! Merged into v1.5.0 branch.

Issue #221, opened 2014-09-29
Memory Leak using headless renderer


codazzo:
Hi,

I'm using Vega in headless mode to render charts on the server with Node. I seem to have found an issue whereby every time I render a chart, some memory is leaked.  The example in https://gist.github.com/codazzo/5dd7310b8d2e612b59e7 exhibits this behaviour on my machine. It prints the memory usage after each time `render()` is called.

There seem to be some cleanup to be done with jsdom - the author suggests using `window.close()` although I haven't been able to find a suitable place in the source to do this.

Could anyone shed some light on this?
Cheers

jheer:
As of v1.5.0, JSDOM is no longer used for SVG rendering.

PR #228, opened 2014-12-28
Fix memory leak when rendering SVG's in headless mode (issue #221)

codazzo:
The leak is caused by the memoization cache in nwmatcher.
Every time vega calls d3.select(), d3 uses one of the available
DOM query methods exposed by jsdom [1], which in turn uses nwmatcher.
These queries are cached for efficiency [2] but if the cache is not
cleared, it grows with a jsdom node every time a new selector is used.
When vega does headless rendering, this means with every new render
as the ID's used in the selectors are incremental.
I discussed a possible fix with @dperini, who suggested calling
`NW.Dom.configure({ SIMPLENOT: true })` to empty the cache before
every new render ([3]).

[1] https://github.com/tmpvar/jsdom/blob/1.0.0/lib/jsdom/selectors/index.js
[2] https://github.com/dperini/nwmatcher/blob/v1.3.3/src/nwmatcher-noqsa.js#L692-L695
[3] https://github.com/dperini/nwmatcher/blob/v1.3.3/src/nwmatcher-noqsa.js#L348-L351


jheer:
Thanks for the PR! This issue is rendered moot by v1.5.0, which removes JSDOM in favor of a leaner headless SVG renderer that emits SVG text directly.

PR #233, opened 2015-01-29
Update interecative demo

khawkins98:
Adds a simple jQuery-powered tooltip. 

Might be a nice example to show how users can integrate jQuery and do tooltips. jQuery isn't strictly necessary to do this, but suspect most would want to use jQuery when heading down this path. If you think it's useful ...

### DIFF
--- a/examples/interact/index.html
+++ b/examples/interact/index.html
@@ -3,9 +3,11 @@
     <title>Vega Interaction Test</title/>
     <script src="../lib/d3.v3.min.js"></script>
     <script src="../../vega.js"></script>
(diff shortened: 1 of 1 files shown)


arvind:
Thanks for the PR! Vega 2.0 supports interaction design from within the specification itself (see [Signals](https://github.com/vega/vega/wiki/Signals) and [Predicates](https://github.com/vega/vega/wiki/Predicates)). We've also added a simple [tooltip](http://vega.github.io/vega-editor/?spec=tooltip) example. Hope this helps!

PR #237, opened 2015-02-10
Fix reverse

domoritz:
Copy the range because reverse is in place and calling domainMinMax multiple times undoes reverse.

In my spec that uses auto padding and reverse, the reverse does not have an effect because it is undone.


jheer:
Thanks! Merged into v1.5.0 branch.

PR #239, opened 2015-02-12
Undefined scale outerPadding should default to padding

### LINKED ISSUE #238: OuterPadding should default to Padding if undefined
A subtle bug was introduced in 4d6cc14 whereby if outerPadding is undefined, it defaults to 0. This differs from the D3 default (and ergo the Vega default before this commit) which is that if outerPadding is undefined, it defaults to the padding.

arvind:
Fixes #238


jheer:
Pushed a fix already. Also, the provided fix fails in the case where outerPadding is intentionally set to zero.

PR #242, opened 2015-02-20
data.load rewrite, Node support improvements

nyurik:
* NodeJS: Added vg.config.defaultProtocol for protocol-relative URLs
     //example.org/data.csv  =>  http://example.org/data.csv

* Rewrote vg.data.load.sanitizeUrl(url) to
  * properly join baseUrl with the url
  * handle relative protocol
  * on Node ensures only 'http' and 'https' protocols

* Switched to 'request' NPM module to handle HTTP downloading
  Current implementation has frequently crashed - better use a lib

* Removed all global vg_load_* functions

* Handle SVG's URLs in loadImage() via sanitizeUrl()

* Added a global utility function vg.startsWith()
  There is a number of inefficient code doing str.indexOf(search) === 0
  that needs to be fixed. The new code uses it, but other changes
  are out of scope for this patch, and should be changed separatelly.


jheer:
Merged into v1.5.0 as of commit cca509e86737ef8202be4b29f609a99fa73274c3.

PR #244, opened 2015-02-22
Preserve sort order in unique

domoritz:
The new unique function does not preserve sort order. This prevented me from using the sort transform to sort the scale domain. 

Also fix tabs/spaces.


domoritz:
I believe this is not the right way to go about this since scales will have a sort property and should ignore the order in the data.

PR #246, opened 2015-02-26
Allow functions in non-safeMode

domoritz:
If safeMode is `false`, all expressions should work. More specifically, I needed `d3.format`. 

Future work could address the issue that a utility function like `d3.format` should be accessible in safeMode as well. Maybe as a utility function `vg.format`.


jheer:
All expressions should now be routed through the "safe" expression parser. If you would like to perform specialized text formatting, please use the new `template` value reference option in v1.5.0.

PR #247, opened 2015-02-27
To string template filter

domoritz:
So mixed data can be handled more safely.


jheer:
I don't want to force users to think through type issues here. Instead, we can do some analysis and inject a cast to String where needed, which I added in 018bd64f4c91b2f76e40953c34b7774d8e04c0ea.

PR #248, opened 2015-02-28
bump canvas to 1.2

grahamscott:
This bumps the node canvas dependency to 1.2.x, which allows us to run on io.js

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
   "dependencies": {
     "d3": "^3.5.5",
     "d3-geo-projection": "^0.2.13",
-    "canvas": "~1.1.6",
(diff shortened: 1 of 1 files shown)


jheer:
Hi @grahamscott, thanks for the PR. I have intentionally avoided upgrading to node-canvas 1.2 due to rendering errors. To replicate: upgrade to canvas 1.2, run `tests/cli-test.sh`, then look at `output/image.png`. The image marks are incorrectly scaled and then clipped. I don't have time at the moment to do a deeper dive, but if you know an effective fix for this problem please share. Once node-canvas behaves correctly I would be happy to upgrade.

PR #250, opened 2015-03-03
Implement comparison with null values

domoritz:
They are always less than any value but equal to each other.


kanitw:
+1

domoritz:
Alternatively, we could convert null to the empty string. Should have pretty much the same effect with less code but not as clean as the case statements.

PR #287, opened 2015-07-09
Add title to Treemap schema

mprudhom:
Since all the other transforms have titles.


jheer:
Thanks! Fixed in b69ea28fe2ffe17afa7567fb673047aac150bda3.

PR #294, opened 2015-07-12
Update requirements

domoritz:
Improves ee5d7a1cdfaae3b1794d320562f22cfa57b1bc5d


domoritz:
Will not properly work until https://github.com/se-panfilov/bower-check-updates/pull/4

jheer:
Added the version pattern updates, but not the additional dependency watchers. Adding that many extra dependencies seems overkill for the task at hand.

PR #301, opened 2015-07-17
additionalProperties should not be false in propset def

mprudhom:
propset has "additionalProperties" set to false, which causes the linking.json spec to fail validation with:

`node scripts/schema.js > vega-schema.json && z-schema vega-schema.json test/spec/linking.json`

### DIFF
--- a/src/parse/properties.js
+++ b/src/parse/properties.js
@@ -565,8 +565,9 @@ properties.schema = {
         "fontWeight": {"$ref": "#/refs/stringValue"},
         "fontStyle": {"$ref": "#/refs/stringValue"}
       },
-
-      "additionalProperties": false
+      "additionalProperties": {
+        "$ref": "#/refs/value"
+      }
     }
   }
 };


arvind:
Yes, this is by design. Although property sets can take custom properties, the json schema does not allow them by default. Instead, it can be extended via the `vg.schema(opt)` method, where `opt` is an object that takes the following properties:
* _url_ - A URL to the pre-compiled JSON schema (`vega-schema.json`).
* _properties_ - An object where keys are the names of custom properties, and values are either `number`, `string`, `boolean`, or `array`. 
* _propsets_ - An object where keys are the names of custom property sets. 

You'll see an example of this is in the `schema.test.js` file, which is how the linking example passes the unit tests. 

I selected this approach to validating property sets because I imagine that the vast majority of specifications do _not_ use custom properties or property sets. As a result, setting `additionalProperties` to anything but `false` would not catch typos and other errors, as the schema would parse them to be an additional property. With `vg.schema`, the schema will always explicitly enumerate the properties and property sets deemed valid. 

Happy to debate whether this is the right approach or not, but I think that is a better discussion to be had on the [Vega Google Group](https://groups.google.com/forum/#!forum/vega-js).

mprudhom:
Continuing discussion at https://groups.google.com/forum/#!topic/vega-js/v0O6PUTojWY

PR #302, opened 2015-07-19
Group sub-marks and sub-scales are not validated against the schema

mprudhom:
The "anyOf" schema for the "marks" array is problematic since "anyOf" implies that a conforming object will be one *or more* schemas, which means that a group mark would be both a group and a non-group mark. Since a group itself has a mark field, that means that the mark information (e.g., properties) would be duplicated.

A more correct implementation is for the "marks" array to contain items that are *either* a group mark or a non-group mark. This PR implements that change by making the "marks" array contain "oneOf" a group mark or a mark that doesn't conform to group.

### DIFF
--- a/src/parse/mark.js
+++ b/src/parse/mark.js
@@ -43,7 +43,6 @@ parseMark.schema = {
             "mark": {"type": "string"},
             "transform": {"$ref": "#/defs/transform"}
           },
-          "oneOf":[{"required": ["data"]}, {"required": ["mark"]}],
           "additionalProperties": false
         },
 
--- a/src/parse/marks.js
+++ b/src/parse/marks.js
@@ -52,9 +52,11 @@ parseRootMark.schema = {
 
     "nonGroupMark": {
       "allOf": [
-        {"$ref": "#/defs/mark"},
-        { "not": {"$ref": "#/defs/groupMark"} }
-      ]
+        {
+          "not": { "properties": { "type": {"enum": ["group"]} } },
+        },
+        {"$ref": "#/defs/mark"}
+      ]   
     }
   }
 };
--- a/src/parse/marks.js
+++ b/src/parse/marks.js
@@ -33,7 +33,7 @@ parseRootMark.schema = {
         },
         "marks": {
           "type": "array",
-          "items": {"oneOf":[{"$ref": "#/defs/groupMark"}, {"$ref": "#/defs/nonGroupMark"}]}
+          "items": {"oneOf":[{"$ref": "#/defs/groupMark"}, {"$ref": "#/defs/visualMark"}]}
         }
       }
     },
@@ -50,7 +50,7 @@ parseRootMark.schema = {
       ]
     },
 
-    "nonGroupMark": {
+    "visualMark": {
       "allOf": [
         {
           "not": { "properties": { "type": {"enum": ["group"]} } },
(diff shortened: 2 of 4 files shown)


arvind:
Thanks for the PR. I'm not sure I understand the problem with defining it using anyOf. Do you have an example specification that produces either a false positive or false negative? Thanks!

mprudhom:
Both the "anyOf" and "oneOf" approaches will validate (pretty much) the same. The issue comes from how a data model that implements the schema will represent the state of the model.

Say you have a data model that implements the mark item as an "anyOf". In pseudo-code, you'd have something like:

```java
class Container {
    List<MarkItem> marks;
    List<Scale> scales;
    List<Axis> axes;
    List<Legend> legends;
}

// A Group is "allOf" a mark and a container
class Group {
    Mark mark;
    Container container;
}

class Mark {
    PropSet enter;
    PropSet update;
    PropSet exit;
    PropSet hover;
}

interface MarkItem { }

// A plain (non-group) mark
class MarkMarkItem implements MarkItem {
    Mark mark;
}

// A "groupMark" is "anyOf" a group or a mark
class GroupMarkItem implements MarkItem {
    Group group;
    Mark mark;

    Map<String, Object> makeJSON() {
        Map<String, Object> json = HashMap<String, Object>();
        if (group != null) {
            json.putAll(group.makeJSON());
        }
        if (mark != null) {
            json.putAll(mark.makeJSON());
        }
        return json;
    }
}
```

Now, say I have an instance of a GroupMarkItem and I want to change one if its "update" properties. Since the group marks properties are duplicated in the "mark" field and in the "group.mark" field, I would need to do it in both places, or else re-serializing the GroupMarkItem would have some ambiguity about which mark properties should take precedence. For example:

```java
GroupMarkItem item = ...;
item.mark.update.fill = "red";
item.group.mark.update.fill = "green";
item.makeJSON(); // will the group mark be filled with red or green?
```

Making the type "oneOf" resolves this ambiguity. FTR, this isn't a hypothetical shortcoming; I'm having this exact issue with my own reification of the schema.

mprudhom:
Furthermore, here is an example of a spec with a perniciously subtle error ("fill" is set to a string literal) that validates successfully against the schema using the "anyOf" definition:

```json
{
  "marks": [
    {
      "type": "group",
      "marks": [
        {
          "type": "rect",
          "properties": {
            "update": { 
              "fill": "#AABBCC"
            }
          }
        }
      ]
    }
  ]
}
```

mprudhom:
After poking around a bit further, it seems that many of the sample specs in the `test/spec/` directory shouldn't be passing validation, but because group sub-marks are not being validated due to this issue, those failures weren't showing up. 

For example, if you move all of the group's sub-marks into top level of `index_chart.json`, you see that the "from" property fails schema validation because it doesn't contain a "data" or "mark" property (which oughtn't be required for group sub-marks).

I imagine that the other validation failures are along these same lines, although I don't have time right now to dig into them.

src/parse/mark.js
```diff
@@ -43,7 +43,6 @@ parseMark.schema = {
             "mark": {"type": "string"},
             "transform": {"$ref": "#/defs/transform"}
           },
-          "oneOf":[{"required": ["data"]}, {"required": ["mark"]}],
```
arvind: Why was this removed? If a mark's `from` property is specified, it should contains either a `data` or `mark` property.

arvind: Oh, I see the explanation in your extended commit message, thanks!

arvind:
How interesting! We've run into the same issues of the JSON schema validating errors with nested group marks/scales. Did not anticipate this to be the error, so thanks for tracking it down!

PR #313, opened 2015-07-30
Update formatting, add locale support.

jheer:
This PR updates all axis, legend and template formatting to pass through datalib's formatting support (^1.4.2). As a result Vega/Datalib now control all automatic formatting decisions, with no reliance on (inconsistent) D3 scale `tickFormat` methods. Most importantly, these facilities support runtime changes to current number and time locale formatting.

For example, to change the number formatting to German (DE-DE):
```javascript
vg.util.format.numberLocale({
  decimal: ",",
  thousands: ".",
  grouping: [3],
  currency: ["", "\xa0€"]
});
```

To change the time formatting to German (DE-DE):
```javascript
vg.util.format.timeLocale({
  dateTime: "%A, der %e. %B %Y, %X",
  date: "%d.%m.%Y",
  time: "%H:%M:%S",
  periods: ["AM", "PM"], // unused
  days: ["Sonntag", "Montag", "Dienstag", "Mittwoch", "Donnerstag", "Freitag", "Samstag"],
  shortDays: ["So", "Mo", "Di", "Mi", "Do", "Fr", "Sa"],
  months: ["Januar", "Februar", "März", "April", "Mai", "Juni", "Juli", "August", "September", "Oktober", "November", "Dezember"],
  shortMonths: ["Jan", "Feb", "Mrz", "Apr", "Mai", "Jun", "Jul", "Aug", "Sep", "Okt", "Nov", "Dez"]
});
```

For more about locale settings, see the [d3-format](https://github.com/d3/d3-format) and [d3-time-format](https://github.com/d3/d3-time-format) module repositories.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,12 +34,12 @@
   "dependencies": {
     "d3": "^3.5.6",
     "d3-geo-projection": "^0.2.15",
-    "datalib": "^1.3.2",
+    "datalib": "^1.4.2",
     "topojson": "^1.6.19",
-    "vega-dataflow": "^1.0.4",
+    "vega-dataflow": "^1.1.0",
     "vega-expression": "^1.0.2",
     "vega-logging": "^1.0.1",
-    "vega-scenegraph": "^1.0.5",
+    "vega-scenegraph": "^1.0.6",
     "yargs": "^3.15.0"
   },
   "devDependencies": {
--- a/src/parse/axes.js
+++ b/src/parse/axes.js
@@ -41,6 +41,7 @@ function parseAxis(config, def, index, axis, group) {
   axis.tickValues(def.values || null);
   // axis label formatting
   axis.tickFormat(def.format || null);
+  axis.tickFormatType(def.formatType || null);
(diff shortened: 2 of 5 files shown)


Fil:
What is the correct method with Vega3, and where is it documented? (Sorry to ask here but it's the only reference I found.)

Inspecting the `vega` object I found the `vega.formatLocale` method, and a way to get the result I want — but it seems I have to call it _before_ parsing the spec, is that correct?

jheer:
Yes, Vega 3 exports the `formatLocale` and `formatTimeLocale` methods for localization settings. These are actually [direct exports of D3 v4's `formatDefaultLocale` and `timeFormatDefaultLocale` methods](https://github.com/vega/vega-parser/blob/master/index.js). These set the global default locales for number and time formatting, respectively.

You will want to invoke these methods _prior_ to instantiating a Vega View and thus before any formatting functions are generated. You should be able to _parse_ a spec ahead of time, but you will need to set the locale before invoking the View constructor for your vis.

PR #317, opened 2015-08-04
Add CountPattern and Wordcloud transforms.

jheer:


jheer:
This PR adds the transforms + schemas. However, we will need to add test cases down the line.

randyzwitch:
Thanks for adding the wordcloud back!

PR #318, opened 2015-08-05
CountPattern and Lookup are not intrinsically batch operations.

arvind:
This preserves the logic of the two operators, but removes the BatchTransform requirement. If parameters change in response to signal values, or if the secondary dataset is pulsed, existing tuples reflow as part of `input.mod`. Thus, we should always be able to run the operators on `input.add`, and only run them when appropriate on `input.mod` and `input.rem`.

### DIFF
--- a/src/transforms/CountPattern.js
+++ b/src/transforms/CountPattern.js
@@ -2,11 +2,10 @@ var dl = require('datalib'),
     df = require('vega-dataflow'),
     Tuple = df.Tuple,
     log = require('vega-logging'),
-    Transform = require('./Transform'),
-    BatchTransform = require('./BatchTransform');
+    Transform = require('./Transform');
 
 function CountPattern(graph) {
-  BatchTransform.prototype.init.call(this, graph);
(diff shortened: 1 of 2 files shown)


src/transforms/CountPattern.js
```diff
@@ -19,47 +18,45 @@ function CountPattern(graph) {
...
-prototype.batchTransform = function(input, data) {
+prototype.transform = function(input) {
   log.debug(input, ['countpattern']);
 
   var get = this.param('field'),
       pattern = this.param('pattern'),
       tcase = this.param('case'),
       stop = this.param('stopwords'),
-      batch = 0;
+      reset = false, run = false;
```
jheer: I think we also need to set reset to true if the `field` parameter changes.

arvind: You're right. The `Transform` parent class provides a `reset` parameter as part of the `transform` method that is true if any dependent signals re-evaluate. I'll switch over to using + augmenting that, which should take care of this case and simplify CountPattern some more.

PR #365, opened 2015-09-17
[WIP] Update & Transition Control

jheer:
Experimental branch for adding more control over animated transitions and update property sets.
- Adds top-level `duration` and `ease` property to Vega spec.
- Adds `update` hash property to signal stream handlers. These hashes may include the properties `duration`, `ease` and `props` (for indicating a property set to invoke instead of `"update"`).

One open question for stream handlers is how to handle event cascades implicating multiple `update` hashes. The current strategy is to use the most recent hash, though an alternative would be to use the very first. Either way, this could lead to program behaviors that users have trouble reasoning about.

This also introduces new logic for requested property sets. If a request is included, the normal enter/update/exit cycle still applies. However, if a property set with a name matching the request is found, it will be evaluated in lieu of the update set. If not found, the normal update set is run.

### DIFF
--- a/src/core/Model.js
+++ b/src/core/Model.js
@@ -4,6 +4,7 @@ var dl = require('datalib'),
     Base = df.Graph.prototype,
     Node  = df.Node, // jshint ignore:line
     GroupBuilder = require('../scene/GroupBuilder'),
+    Transition = require('../scene/Transition'),
     visit = require('../scene/visit'),
     config = require('./config');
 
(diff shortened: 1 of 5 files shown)


arvind:
Thinking through this some, one alternative is to put the `update` property within each signal definition rather than a signal's stream handlers. That way, animation is part of the "interaction logic" and signals continue to be an abstraction barrier between it and low-level input events. Doesn't alleviate the problem of update cascades though...

jheer:
Closing as this branch is now completely out of sync. Techniques here may be applied to a future PR.

PR #370, opened 2015-09-24
Update d3-cloud.

jasondavies:
This includes a fix for using d3-cloud in “headless” mode with Node.js.  Related: #369.


jheer:
Thanks @jasondavies!

@arvind, I'll look into this. We may want to use our scenegraph canvas utils here instead of canvas-browserify (or modify scenegraph accordingly).

PR #383, opened 2015-10-09
Production rule valuerefs as arrays

RussellSprouts:
This adds a shortcut for defining production rules. Previously, a rule was defined as

    "fill": {
        "rule": [
            {"predicate":...},
            ...
        ]
    }

Now, you can just do this:

    "fill": [
        {"predicate":...},
        ...   
    ]

The old syntax still works. The schema has been updated to allow either one.

### DIFF
--- a/src/parse/properties.js
+++ b/src/parse/properties.js
@@ -362,7 +362,7 @@ function scaleRef(ref) {
 
 module.exports = properties;
 
-function valueSchema(type) {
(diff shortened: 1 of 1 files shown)


arvind:
Thanks @RussellSprouts. We will consider the `rule` property deprecated, and will remove it in the subsequent minor version.

PR #388, opened 2015-10-14
Updated Bin transform.

jheer:
This PR updates the bin transform in two ways:
 1. Perform automatic calculation of min/max if not specified.
 2. Output both bin start and end values to simplify later operations.

Item 1 is implemented directly using datalib's `extent` method rather than a streaming aggregator. To allow this, `Bin` is now a `BatchTransform` instance. This seems simpler and lower overhead than instantiating an aggregator and all its associated state (which would include another full array with all the tuples in it). Simple min/max scans should be pretty fast, and can also be cached more effectively in future updates.

Item 2 extends the binning output with an `"end"` parameter that defaults to `"bin_end"`. For internal consistency, it would be nice to have `"start"` and `"end"` parameters that default to `"bin_start"` and `"bin_end"`. However, for backwards-compatibility I've left the bin_start parameter as simply `"bin"`. Any thoughts on whether we should make a breaking change here? It appears that the crossfilter example is currently the only public example that uses the bin transform, so that is what we would need to update, but any changes here could adversely impact end users.

### DIFF
--- a/src/transforms/Bin.js
+++ b/src/transforms/Bin.js
@@ -19,8 +19,8 @@ function Bin(graph) {
   });
 
   this._output = {
-    bin: 'bin',
-    end: 'bin_end'
+    start: 'bin_start',
+    end:   'bin_end'
   };
   return this.mutates(true);
 }
@@ -65,14 +65,14 @@ prototype.batchTransform = function(input, data) {
     var v = get(d);
     v = v == null ? null
       : b.start + b.step * ~~((v - b.start) / b.step);
-    Tuple.set(d, output.bin, v);
+    Tuple.set(d, output.start, v);
     Tuple.set(d, output.end, v + b.step);
   }
   input.add.forEach(update);
   input.mod.forEach(update);
   input.rem.forEach(update);
 
-  input.fields[output.bin] = 1;
+  input.fields[output.start] = 1;
   input.fields[output.end] = 1;
   return input;
 };
@@ -141,7 +141,7 @@ Bin.schema = {
       "type": "object",
       "description": "Rename the output data fields",
       "properties": {
-        "bin": {"type": "string", "default": "bin"},
+        "start": {"type": "string", "default": "bin_start"},
         "end": {"type": "string", "default": "bin_end"}
       },
       "additionalProperties": false
--- a/src/transforms/Bin.js
+++ b/src/transforms/Bin.js
@@ -1,10 +1,11 @@
-var bins = require('datalib').bins,
+var dl = require('datalib'),
     Tuple = require('vega-dataflow').Tuple,
     log = require('vega-logging'),
-    Transform = require('./Transform');
+    Transform = require('./Transform'),
+    BatchTransform = require('./BatchTransform');
 
 function Bin(graph) {
-  Transform.prototype.init.call(this, graph);
+  BatchTransform.prototype.init.call(this, graph);
   Transform.addParameters(this, {
     field: {type: 'field'},
     min: {type: 'value'},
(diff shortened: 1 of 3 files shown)


jheer:
Here is an example spec for a histogram annotated with median and quartile boundaries, demonstrating how the new bin start/end outputs simplify scale construction and mark composition, all along a shared numerical axis. @domoritz @kanitw 

```
{
  "width": 400,
  "height": 200,
  "data": [
    {
      "name": "source",
      "values": [
      ],
      "transform": [
        { "type": "bin", "field": "y", "maxbins": 10,
          "output": {"bin": "bin_start", "end": "bin_end"} }
      ]
    },
    {
      "name": "bins",
      "source": "source",
      "transform": [
        { "type": "aggregate", "groupby": ["bin_start", "bin_end"],
          "summarize": {"*": "count"} }
      ]
    },
    {
      "name": "stats",
      "source": "source",
      "transform": [
        { "type": "aggregate",
          "summarize": {"y": ["q1", "median", "q3"]} }
      ]
    }
  ],
  "scales": [
    {
      "name": "x",
      "type": "linear",
      "range": "width",
      "zero": false, "round": true,
      "domain": {"data": "bins", "field": ["bin_start", "bin_end"]}
    },
    {
      "name": "y",
      "type": "linear",
      "range": "height",
      "domain": {"data": "bins", "field": "count"},
      "nice": true
    }
  ],
  "axes": [
    {"type": "x", "scale": "x"},
    {"type": "y", "scale": "y"}
  ],
  "marks": [
    {
      "type": "rect",
      "from": {"data": "stats"},
      "properties": {
        "enter": {
          "x": {"scale": "x", "field": "q1_y"},
          "x2": {"scale": "x", "field": "median_y"},
          "y": {"value": -5},
          "height": {"field": {"group": "height"}, "offset": 5},
          "fill": {"value": "lightgray"},
          "fillOpacity": {"value": 0.6}
        }
      }
    },
    {
      "type": "rect",
      "from": {"data": "stats"},
      "properties": {
        "enter": {
          "x": {"scale": "x", "field": "median_y"},
          "x2": {"scale": "x", "field": "q3_y"},
          "y": {"value": -5},
          "height": {"field": {"group": "height"}, "offset": 5},
          "fill": {"value": "lightgray"}
        }
      }
    },
    {
      "type": "rect",
      "from": {"data": "bins"},
      "properties": {
        "enter": {
          "x": {"scale": "x", "field": "bin_start"},
          "x2": {"scale": "x", "field": "bin_end", "offset": -1},
          "y": {"scale": "y", "field": "count"},
          "y2": {"scale": "y", "value": 0}
        },
        "update": {
        },
        "hover": {
          "fill": {"value": "red"}
        }
      }
    },
    {
      "type": "rect",
      "from": {"data": "stats"},
      "properties": {
        "enter": {
          "x": {"scale": "x", "field": "median_y"},
          "width": {"value": 2},
          "y": {"value": -5},
          "height": {"field": {"group": "height"}, "offset": 5},
          "fill": {"value": "crimson"}
        }
      }
    }
  ]
}
```

arvind:
My vote would be to rename `bin` to `start`. We can bump the minor version and message the release notes to call attention to and mitigate adverse effects to end-users.

jheer:
I agree. I think we constitute a quorum :) I will make the change and update the PR.

kanitw:
+1 on rename `bin` to `start`.

jheer:
OK, updated bin outputs and updated crossfilter test spec. The crossfilter example in vega-editor will also need to be updated once this is merged and released. I will now update the above comment with the working example spec. We also need to remember to update the documentation upon release.

jeffbaumes:
:clap: I had run into this issue too on building a generic histogram component, thanks for fixing!

PR #393, opened 2015-10-20
Add impute transform.

jheer:
**Work-in-progress** PR for adding basic missing value imputation.
TODOs: (1) write unit tests, (2) finalize design / iron out any bugs.

### DIFF
--- a/src/transforms/Impute.js
+++ b/src/transforms/Impute.js
@@ -1,0 +1,152 @@
+var dl = require('datalib'),
+    log = require('vega-logging'),
+    Tuple = require('vega-dataflow').Tuple,
+    Transform = require('./Transform'),
+    BatchTransform = require('./BatchTransform');
+
+function Impute(graph) {
+  BatchTransform.prototype.init.call(this, graph);
+  Transform.addParameters(this, {
+    groupby: {type: 'array<field>'},
+    orderby: {type: 'array<field>'},
+    field:   {type: 'field'},
+    method:  {type: 'value', default: 'value'},
+    value:   {type: 'value', default: 0}
+  });
+
+  return this.router(true).produces(true);
+}
+
+var prototype = (Impute.prototype = Object.create(BatchTransform.prototype));
+prototype.constructor = Impute;
+
+prototype.batchTransform = function(input, data) {
+  log.debug(input, ['imputing']);
+
+  var groupby = this.param('groupby'),
+      orderby = this.param('orderby'),
+      method = this.param('method'),
+      value = this.param('value'),
+      field = this.param('field'),
+      get = field.accessor,
+      name = field.field,
+      prev = this._imputed || [], curr = [],
+      groups = partition(data, groupby.accessor, orderby.accessor),
+      domain = groups.domain,
+      group, i, j, n, m, t;
+
+  function getval(x) {
+    return x == null ? null : get(x);
+  }
+
+  for (j=0, m=groups.length; j<m; ++j) {
+    group = groups[j];
+
+    // determine imputation value
+    if (method !== 'value') {
+      value = dl[method](group, getval);
+    }
+
+    // add tuples for missing values
+    for (i=0, n=group.length; i<n; ++i) {
+      if (group[i] == null) {
+        t = tuple(groupby.field, group.values, orderby.field, domain[i]);
+        t[name] = value;
+        curr.push(t);
+      }
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


jheer:
The basic impute transform in this PR appears to be working. However, updates do not always propagate correctly. @arvind, try the spec below, in which I drive the imputation value from a constant set in a signal. When I set the `missing` signal value to zero via the View API, the scales update correctly but the area does not. Is there perhaps a problem with dirty tracking? (Note that, upon signal change, the impute transform is not mod'ing a tuple, it rems the prior tuple and adds a new tuple.)

```
{
  "width": 500,
  "height": 200,
  "signals": [
    {"name": "missing", "init": 200}
  ],
  "data": [
    {
      "name": "table",
      "values": [
        {"x": 0, "y": 28, "c":0}, {"x": 0, "y": 55, "c":1},
        {"x": 1, "y": 43, "c":0}, {"x": 1, "y": 91, "c":1},
        {"x": 2, "y": 81, "c":0}, {"x": 2, "y": 53, "c":1},
        {"x": 3, "y": 19, "c":0}, 
        {"x": 4, "y": 52, "c":0}, {"x": 4, "y": 48, "c":1},
        {"x": 5, "y": 24, "c":0}, {"x": 5, "y": 49, "c":1},
        {"x": 6, "y": 87, "c":0}, {"x": 6, "y": 66, "c":1},
        {"x": 7, "y": 17, "c":0}, {"x": 7, "y": 27, "c":1},
        {"x": 8, "y": 68, "c":0}, {"x": 8, "y": 16, "c":1},
        {"x": 9, "y": 49, "c":0}, {"x": 9, "y": 15, "c":1}
      ],
      "transform": [
        {
          "type": "impute",
          "groupby": ["c"],
          "orderby": ["x"],
          "field": "y",
          "method": "value",
          "value": {"signal": "missing"}
        },
        { "type": "sort", "by": ["x"]}
      ]
    },
    {
      "name": "stats",
      "source": "table",
      "transform": [
        {
          "type": "aggregate",
          "groupby": ["x"],
          "summarize": [{"field": "y", "ops": ["sum"]}]
        }
      ]
    }
  ],
  "scales": [
    {
      "name": "x",
      "type": "ordinal",
      "range": "width",
      "points": true,
    },
    {
      "name": "y",
      "type": "linear",
      "range": "height",
      "nice": true,
      "domain": {"data": "stats", "field": "sum_y"}
    },
    {
      "name": "color",
      "type": "ordinal",
      "range": "category10",
      "domain": {"data": "table", "field": "c"}
    }
  ],
  "axes": [
    {"type": "x", "scale": "x"},
    {"type": "y", "scale": "y"}
  ],
  "marks": [
    {
      "type": "group",
      "from": {
        "data": "table",
        "transform": [
          {"type": "stack", "groupby": ["x"], "sortby": ["c"], "field": "y"},
          {"type": "facet", "groupby": ["c"]}
        ]
      },
      "marks": [
        {
          "type": "area",
          "properties": {
            "enter": {
              "interpolate": {"value": "monotone"},
              "y": {"scale": "y", "field": "layout_start"},
              "y2": {"scale": "y", "field": "layout_end"},
              "fill": {"scale": "color", "field": "c"},
              "fillOpacity": {"value": 1}
            }
          }
        }
      ]
    }
  ]
}
```

arvind:
Haven't had time to debug this in-depth yet but one thing that jumps out at me -- area properties are only being set on `enter`. The area updates when I switch it to `update` but updates seem to get really mangled. If I remove from the spec all the values that come after the one that should be imputed (i.e., `x >= 4`) and then manipulate the signal, the area seems to update correctly. So, I wonder if sorting is going wrong somewhere?

jheer:
There should be a forehead slap emoji to use for the `enter` / `update` mixup... I'm seeing the same problem as you, namely, tuples are not being sorted properly in the scenegraph. Note that `stack` will produce the same result regardless of sort order - the area interpolation is dependent on the scenegraph items being property sorted.

arvind:
Fixed in 89bbf1e (and vega/vega-dataflow@e0f4d66). We weren't setting `_input` in `Facetor` (our `dl.Aggregator` subclass). And so `sort` was not being applied to the cell DataSources produced by the Facet transform.

jheer:
OK, this should be good to go on top of vega-dataflow 1.2.7.

kanitw:
stack transform has a property called `sortby` while impute has `orderby`  — should we rename it to be consistent?

Issue #392, opened 2015-10-20
Scale Invert Issue


jhoffswell:
I noticed a problem with the `.invert` function on scales. This example shows itself in the `linking` example for the `gy` scale. While `gx.invert` produces the correct set of results when converting from a pixel value, `gy.invert` often produces `undefined`.

PR #394, opened 2015-10-21
Ensure ordinal range is ascending for inversion. 

arvind:
I wanted to do this via PR, in case you had a cleverer way of fixing #392.

### DIFF
--- a/src/scene/Scale.js
+++ b/src/scene/Scale.js
@@ -115,7 +115,7 @@ function ordinal(scale, rng, group) {
       outer = def.outerPadding == null ? pad : signal.call(this, def.outerPadding),
       points = def.points && signal.call(this, def.points),
       round = signal.call(this, def.round) || def.round == null,
-      domain, str;
+      domain, str, inv;
   
   // range pre-processing for data-driven ranges
   if (dl.isObject(def.range) && !dl.isArray(def.range)) {
@@ -162,17 +162,23 @@ function ordinal(scale, rng, group) {
     scale.rangeBands(rng, pad, outer);
   }
 
-  if (!scale.invert) {
-    scale.invert = function(x, y) {
+  prev.range = rng;
+  this._updated = true;
+
+  if (!(inv=scale.invert)) {
+    inv = scale.invert = function(x, y) {
       if (arguments.length === 1) {
-        return scale.domain()[d3.bisect(scale.range(), x) - 1];
+        if (!dl.isNumber(x)) {
+          throw new Error('Ordinal scale inversion is only supported for numeric input ('+x+').');
+        }
+        return inv._domain[d3.bisect(inv._range, x) - 1];
       } else if (arguments.length === 2) {  // Invert extents
         if (!dl.isNumber(x) || !dl.isNumber(y)) {
-          throw Error('Extents to ordinal invert are not numbers ('+x+', '+y+').');
+          throw new Error('Extents to ordinal invert are not numbers ('+x+', '+y+').');
         }
 
         var points = [],
-            rng = scale.range(),
+            rng = inv._range,
             i = 0, len = rng.length, r;
 
         for(; i<len; ++i) {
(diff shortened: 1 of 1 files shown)


src/scene/Scale.js
```diff
@@ -162,17 +162,23 @@ function ordinal(scale, rng, group) {
     scale.rangeBands(rng, pad, outer);
   }
 
-  if (!scale.invert) {
-    scale.invert = function(x, y) {
+  prev.range = rng;
```
jheer: Not sure I follow this line. If `prev` is the previous scale reference, this is overwriting its `range` method. But maybe `prev` is something else?

src/scene/Scale.js
```diff
@@ -187,8 +193,34 @@ function ordinal(scale, rng, group) {
     };
   }
 
-  prev.range = rng;
-  this._updated = true;
+  // Ordinal scale inversion requires range in ascending order.
+  // Shuffle domain/range to ensure this is true, and cache.
+  if (this._updated) {
```
jheer: `this._updated` is set to `true` above. I don't see how it could ever be `false` at this point. So why the `if` statement?

src/scene/Scale.js
```diff
@@ -187,8 +193,34 @@ function ordinal(scale, rng, group) {
...
+      if ((shuffle = invRange[i] > invRange[i+1])) break;
+    }
+
+    if (shuffle) {
+      for (i=0, n=invDomain.length; i<n; ++i) {
+        map[invRange[i]] = invDomain[i];
+      }
+
+      invDomain = [];
+      invRange  = dl.duplicate(invRange).sort(dl.cmp);
```
jheer: Why `duplicate` here, rather than `slice`?

src/scene/Scale.js
jheer: It looks like the primary strategy is to hold on to copies of the domain and range, but properly reversed if needed, in order to compute the inverse. Could we reduce data structure duplication here by instead maintaining a flag and then modifying the search procedure accordingly?

src/scene/Scale.js
arvind: `prev` is just an arbitrary object here, not a previous scale reference. We use it to determine if the scale should be identified (within the changeset) as updated. So this shouldn't be a problem.

src/scene/Scale.js
arvind: Ah shoot. Your comment on `_updated` points out a control flow flaw here. If we do an early return, the inverse structures will never be recomputed. 

Re: modifying the search procedure, I was hoping to do that initially but haven't been able to think of a way that generalizes for arbitrarily unordered numeric ranges that doesn't amount to just walking the arrays on every invert call. So, given the typical space/time tradeoff, I figured that since scale inversion demands interactive performance, the additional space was worth it. 

Let me know if you think of a better strategy that I've overlooked!

```json
{
  "scales": [
    {
      "name": "s",
      "type": "ordinal",
      "domain": ["a", "b", "c"],
      "range": [50, 3, 25]
    }  
  ]
}
```

src/scene/Scale.js
arvind: Good call!

arvind:
Ok, comments addressed in 39de562. Let me know if you think of a better search strategy that doesn't require maintaining separate domain/range structures.

jheer:
I think I may have found a better search strategy; see snippet below. Also, I think I would prefer to have the scale invert "polyfill" extracted out of Scale.js internals into a proper utility function (even if it still lives elsewhere inside of the Scale.js module as a top-level function). This could result in cleaner, easier to read code, as opposed to its current home injected into the midst of the `ordinal` method.
```
var bisect = d3.bisector(dl.numcmp).right,
    findAsc = function(a, x) { return bisect(a,x) - 1; },
    findDsc = d3.bisector(function(a,b) { return -1 * dl.numcmp(a,b); }).left;

scale.invert = function(x, y) {
  var rng = this.range(),
      asc = rng[0] < rng[1];

  if (arguments.length === 1) {
    if (!dl.isNumber(x)) {
      throw Error('Ordinal scale inversion is only supported for numeric input ('+x+').');
    }

    var find = asc ? findAsc : findDsc;
    return this.domain()[find(rng, x)];

  } else if (arguments.length === 2) {  // Invert extents
    if (!dl.isNumber(x) || !dl.isNumber(y)) {
      throw Error('Extents to ordinal invert are not numbers ('+x+', '+y+').');
    }

    var points = [], i = 0, n = rng.length-1, r;
    if (y < x) { r = x; x = y; y = x; } // ensure x <= y

    for (; i<=n; ++i) {
      r = rng[asc ? n-i : i];
      if (x <= r && r <= y) points.push(r);
    }
    return points.map(function(p) { return scale.invert(p); });
  }
}
```
In the two-argument case, the use of the `asc` flag to control iteration order may not be strictly necessary: the points array should have the same members regardless. That said, including it ensures a particular ordering of values that may be preferred.

jheer:
Note that my search strategy does not handle arbitrarily ordered numbers, but is that a case we need to support? If so, we might combine our strategies and support binary search over either ascending or descending sorts, and duplicate data only when it has an arbitrary ordering.

arvind:
Thanks for this! Yeah, that's something I went back and forth on. The use case I was thinking of was arbitrarily arranging domain values (eg a dashboard or annotations).

But I think there are enough other ways to express that in the language (eg positions as part of your data model). Moreover, considering we don't support inverting non-numeric ranges (i.e. arbitrary ordinal mappings), I think we can document that a numeric ordinal range must be ordered if you wish to invert it.  

I'll merge your snippet once I get into the office. Thanks!

-Arvind
-----
Arvind Satyanarayan <http://arvindsatya.com/>

arvind:
Merged with e60b61b, and added a unit test as well.

PR #402, opened 2015-10-29
Report parsing errors to the caller

nyurik:
This PR includes the cleanup PR https://github.com/vega/vega/compare/master...nyurik:cleanup?expand=1

It exposes the logging interface to allow for complex error processing

Alternative spec.parse() with error handling
* Handle the case when a user-overriden logger throws an error
* In case of an internal error, report it via callback(error, model)

TODO:
This is a fairly bad workaround, and could really use promises.
The error messages should be replaced with objects to allow
complex error processing.

### DIFF
--- a/index.js
+++ b/index.js
@@ -15,5 +15,6 @@ module.exports = {
   schema: require('./src/core/schema'),
   config: require('./src/core/config'),
   util:  require('datalib'),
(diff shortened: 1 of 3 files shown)


nyurik:
This PR is superseded by #431

PR #405, opened 2015-11-02
Call `View` with `this` instead of `null`

joneshf:
This call was blowing up in the browser when used. I'm not sure how it wasn't causing an issue in node.

`View` is expecting to set properties on the `this` context,
and you can't set properties on `null`.

I apologize for not performing a proper fork and testing and whatnot, but I cannot for the life of me get `vega` to build locally, so I cannot run the tests.

### DIFF
--- a/src/core/HeadlessView.js
+++ b/src/core/HeadlessView.js
@@ -4,7 +4,7 @@ var sg = require('vega-scenegraph').render,
     View = require('./View');
 
 function HeadlessView(width, height, model) {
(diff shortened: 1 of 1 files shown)


arvind:
Nice catch, thanks @joneshf! I think the reason it wasn't throwing any errors is because Vega's invocations of `HeadlessView` are routed through `View.factory`. The constructor doesn't serve much of a purpose as we set properties through subsequent chained method calls.

PR #411, opened 2015-11-04
Fix SVGRenderer errors when calling View.padding("auto")

Jerrythafast:
[EDIT: title has changed, see 3rd comment]

Adds a new option `autopad` to the `opt` argument of `View.update()`. If it is `true`-ish, Vega will re-run the `autopad` routine. This is useful if the size of the visualisation changes after the `update()` call.

### DIFF
--- a/src/core/View.js
+++ b/src/core/View.js
@@ -166,8 +166,8 @@ prototype.padding = function(pad) {
       this._padding = pad;
       this._strict = false;
     }
-    if (this._renderer) this._renderer.resize(this._width, this._height, pad);
-    if (this._handler)  this._handler.padding(pad);
+    if (this._renderer) this._renderer.resize(this._width, this._height, this._padding);
+    if (this._handler)  this._handler.padding(this._padding);
   }
   return (this._repaint = true, this);
 };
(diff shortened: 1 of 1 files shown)


arvind:
Hi @Jerrythafast, thanks for the PR. I'm curious, is there a reason `view.padding("auto").update();` doesn't work instead?

Jerrythafast:
Hey @arvind, thanks for the reply. With the visualisation I needed this for, Vega actually crashed when I did `view.padding("auto")`. So I kind of assumed that it was not possible to do it this way. I can try it again next Monday and then report back, either with a crash report or with apologies if it turns out I messed up ;)

If this is supposed to be possible, it might be advisable to update the [Runtime](https://github.com/vega/vega/wiki/Runtime) Wiki page as well. Currently it clearly states you need to send in an object with particular properties.

Jerrythafast:
Hey @arvind, I have checked what the issue was that I was having with this before. Turns out the crash I was having was actually unrelated to the padding call, but it was buried under three errors coming from the SVGRenderer that confused me. I have now changed this pull request to fix the `View.padding("auto")` call.

The crash I was facing was actually that `data` is null at [line 375 of Scale.js](https://github.com/vega/vega/blob/master/src/scene/Scale.js#L375). But that must be a completely different issue.

PR #427, opened 2015-11-10
Add automatic signals for width, height and padding.

jheer:
- Add named signals for the View `width`, `height` and `padding`.
- Add `"width"`, `"height"` and `"padding"` to the list of reserved signal names. Note that the `horizon` and `panzoom_points` examples require removal of explicit width/height signals to work.
- Add expression support (`{"expr": "..."}`) to transform parameters.
- Update layouts to use size signals as defaults for `size`, `translate` and `clipExtent` parameters.

The new signals have been added in a minimal fashion, with all existing width/height/padding processing otherwise unchanged. It may be that the use of signals allows for more elegant handling across the View, Model and scenegraph, but for now this PR focuses solely on adding the new functionality.

### DIFF
--- a/src/core/Model.js
+++ b/src/core/Model.js
@@ -19,6 +19,10 @@ function Model(cfg) {
 
   this.config(cfg);
   Base.init.call(this);
+
+  // create signals for width and height
+  this.signal('width', -1);
+  this.signal('height', -1);
 }
 
 var prototype = (Model.prototype = Object.create(Base));
--- a/src/core/Model.js
+++ b/src/core/Model.js
@@ -19,10 +19,6 @@ function Model(cfg) {
 
   this.config(cfg);
   Base.init.call(this);
-
-  // create signals for width and height
-  this.signal('width', -1);
-  this.signal('height', -1);
 }
 
 var prototype = (Model.prototype = Object.create(Base));
--- a/src/core/View.js
+++ b/src/core/View.js
@@ -170,6 +170,7 @@ prototype.padding = function(pad) {
     }
     if (this._renderer) this._renderer.resize(this._width, this._height, this._padding);
     if (this._handler)  this._handler.padding(this._padding);
+    this.signal('padding', this._padding);
   }
   return (this._repaint = true, this);
 };
@@ -197,6 +198,7 @@ prototype.autopad = function(opt) {
     this._model.width(this._width).height(this._height).reset();
     this.signal('width', this._width);
     this.signal('height', this._height);
+    this.signal('padding', pad);
 
     this.initialize().update({props:'enter'}).update({props:'update'});
   } else {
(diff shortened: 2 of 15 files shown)


jheer:
Here is a test spec for the horizon graph example. Open the console and change the chart height dynamically (e.g., `ved.view.height(100).update()`) to further test.
```
{
  "width": 500,
  "height": 50,

  "signals": [
    { "name": "layers", "init": 2,
      "streams": [
        {"type": "click", "expr": "1 + (layers % 4)"}
      ]
    },
    { "name": "vheight", "init": {"expr": "height * layers"},
      "streams": [
        {"type": "layers, height", "expr": "height * layers"}
      ]
    },
    { "name": "opacity", "init": {"expr": "pow(layers, -2/3)"},
      "streams": [
        {"type": "layers", "expr": "pow(layers, -2/3)"}
      ]
    }
  ],

  "data": [
    {
      "name": "layer_indices",
      "values": [0, 1, 2, 3],
      "transform": [
        {"type": "filter", "test": "datum.data < layers"},
        {"type": "formula", "field": "offset", "expr": "datum.data * -height"}
      ]
    },
    {
      "name": "table",
      "values": [
      ]
    }
  ],
  "scales": [
    {
      "name": "x",
      "type": "linear",
      "range": "width",
      "zero": false, "round": true,
    },
    {
      "name": "y",
      "type": "linear",
      "range": [{"signal":"vheight"}, 0],
      "nice": true,
      "domain": {"data": "table", "field": "y"}
    }
  ],
  "axes": [
    {"type": "x", "scale": "x", "ticks": 20}
  ],
  "marks": [
    {
      "type": "group",
      "properties": {
        "update": {
          "width": {"field": {"group": "width"}},
          "height": {"field": {"group": "height"}},
          "clip": {"value": true}
        }
      },
      "marks": [
        {
          "type": "group",
          "from": {"data": "layer_indices"},
          "properties": {
            "update": {
              "y": {"field": "offset"}
            }
          },
          "marks": [
            {
              "type": "area",
              "from": {"data": "table"},
              "properties": {
                "enter": {
                  "interpolate": {"value": "monotone"},
                },
                "update": {
                  "y2": {"scale": "y", "value": 0},
                  "fillOpacity": {"signal": "opacity"}
                }
              }
            }
          ]
        }
      ]
    }
  ]
}
```

jheer:
@arvind, can you please review [View.js#L351](https://github.com/vega/vega/blob/jh/size-signals/src/core/View.js#L351) and comment? I had to ensure re-evaluation of the graph _and_ processing of updated signals to get correct updates on dynamic width / height changes. The `_repaint` flag appeared to already correlate with the situations of interest. But, I'm not sure if this is the best way to proceed.

jheer:
OK, I believe this is now ready to merge, pending review.

@arvind, I made a modification to View.update that you should carefully examine. The key bit is now at [View.js#353](https://github.com/vega/vega/blob/jh/size-signals/src/core/View.js#L353). The motivating issue is that we now have cases where both signal updates and full model updates are needed on the same pulse. I've addressed this by adding the model node as a temporary dependency of the streamer node when the `_repaint` flag is true. This change may even fix potential (latent) update bugs (e.g., regarding auto-padding updates).

jeffbaumes:
These are great additions. I was finding that width and height signals were often needed and this will eliminate that code.

arvind:
Making my way through this PR. It's curious that the full model update is required, and that simply wiring up operators that depend on the new signals is insufficient. Need to think through that more. 

In the meanwhile, one thing missing from this PR is handling updates to width/height/padding via the `view.signal` API. For example, with your horizon graph example above, `ved.view.height(100).update()` correctly updates the area marks and resizes the canvas. However, `view.signal('height', 100).update()` only updates the area marks.

jheer:
Good point. I was implicitly treating them as "read-only" signals. The
easiest fix that comes to mind is to intercept signal setters for
width/height/padding and re-route those to their View setter methods. Not
very elegant, but should be functional...

jheer:
OK, updated in e48b160 to handle updates made through the signal API. Also rebased to follow the current master branch.

arvind:
Cool, merged with e48b160. For posterity, and to answer my question above: prior to this PR, repainting would inadvertently pulse the entire model due the branching in `View.update`. Thus, this is preserved now when repainting also sets signals (and ergo streamer listeners). Finally, why pulse the whole model? This is the only way to pulse the root scene graph node (and its scales etc.), which is hardwired to read width/height from the model. As the initial PR post notes, there is likely some more work needed to unify width/height/padding calculation across the view, model, and operators in an elegant fashion.

Issue #395, opened 2015-10-26
Crash: "c is undefined" in Cross+filter transform after Aggregate


Jerrythafast:
Dear Devs,

I got another interesting crash. I think I kind of caused this myself, because it seems to be related to the `filter` attribute of the `cross` transform.

Here it is:
```json
{
  "data": [
    {
      "name": "table",
      "values": [
        {"a": 1, "b": 1, "c": 1},
        {"a": 1, "b": 1, "c": 2},
        {"a": 1, "b": 2, "c": 1},
        {"a": 1, "b": 2, "c": 2},
        {"a": 2, "b": 1, "c": 1},
        {"a": 2, "b": 1, "c": 2},
        {"a": 2, "b": 2, "c": 1},
        {"a": 2, "b": 2, "c": 2}
      ]
    },
    {
      "name": "agcross",
      "source": "table",
      "transform": [
        {
          "type": "aggregate",
          "groupby": ["a"],
          "summarize": [{"field": "b", "ops": ["distinct"], "as": ["count"]}]
        },
        {
          "type": "cross",
          "with": "table",
          "filter": "datum.b.a < datum.a.a"
        }
      ]
    }
  ]
}
```

The issue is this: Vega crashes at [line 63 of Cross.js](https://github.com/vega/vega/blob/master/src/transforms/Cross.js#L63) with "TypeError: c is undefined". I am not exactly sure why, but I couldn't get it to crash without the `aggregate` transform so it seems to be related to that as well.

The expected output, which can be obtained by moving the `filter` to a separate `filter` transform, is:
```json
[
  {
    "a": {"a": 2, "count": 2},
    "b": {"a": 1, "b": 1, "c": 1}
  },
  {
    "a": {"a": 2, "count": 2},
    "b": {"a": 1, "b": 1, "c": 2}
  },
  {
    "a": {"a": 2, "count": 2},
    "b": {"a": 1, "b": 2, "c": 1}
  },
  {
    "a": {"a": 2, "count": 2},
    "b": {"a": 1, "b": 2, "c": 2}
  }
]
```

It's probably just an interesting corner case again ;)

Kind regards,
Jerry

EDIT: Small update. It does not seem to be dependent on the `aggregate` after all. The following spec, in which I included the `aggregate` output explicitly, crashes in exactly the same way:
```json
{
  "data": [
    {
      "name": "table",
      "values": [
        {"a": 1, "b": 1, "c": 1},
        {"a": 1, "b": 1, "c": 2},
[...]

PR #362, opened 2015-09-16
Add `filter` property to `cross` transform

Jerrythafast:
Implements/resolves vega/vega#338

PR #428, opened 2015-11-11
Self-cross transforms should process right changes

Jerrythafast:
If you cross data, that was filtered using a signal value, with itself (i.e. `"transform": [{"type": "cross"}]`), the `cross` transform did not behave correctly when more data passes the filter after the value of the signal changes. The new data objects were only crossed 'from the left' and would never appear on the right, except for the element on the diagonal.

### DIFF
--- a/src/transforms/Cross.js
+++ b/src/transforms/Cross.js
@@ -105,7 +105,7 @@ prototype.batchTransform = function(input, data) {
   input.rem.forEach(r);
   input.add.forEach(add.bind(this, output, true, wdata, diag, test));
 
-  if (!selfCross && woutput.stamp > this._lastWith) {
+  if (woutput.stamp > this._lastWith) {
     woutput.rem.forEach(r);
     woutput.add.forEach(add.bind(this, output, false, data, diag, test));
     woutput.mod.forEach(mod.bind(this, output, false));
--- a/src/transforms/Cross.js
+++ b/src/transforms/Cross.js
@@ -114,7 +114,7 @@ prototype.batchTransform = function(input, data) {
   }
 
   // Mods need to come after all removals have been run.
-  input.mod.forEach(mod.bind(this, output, true, data, diag, test));
+  input.mod.forEach(mod.bind(this, output, true, wdata, diag, test));
   upFields.call(this, input, output);
 
   return output;
--- a/src/transforms/Cross.js
+++ b/src/transforms/Cross.js
@@ -56,10 +56,10 @@ function add(output, left, data, diag, test, x) {
   }
 }
 
-function mod(output, left, x) {
+function mod(output, left, data, diag, test, x) {
   var cross = this,
       c = this._cache[x._id];
-  if (!c) return;
+  if (!c) return add(output, left, data, diag, test, x);
   
   if (this._lastRem > c.s) {  // Removed tuples haven't been filtered yet
     c.c = c.c.filter(function(y) {
@@ -105,16 +105,16 @@ prototype.batchTransform = function(input, data) {
   input.rem.forEach(r);
   input.add.forEach(add.bind(this, output, true, wdata, diag, test));
 
-  if (woutput.stamp > this._lastWith) {
+  if (!selfCross && woutput.stamp > this._lastWith) {
     woutput.rem.forEach(r);
     woutput.add.forEach(add.bind(this, output, false, data, diag, test));
-    woutput.mod.forEach(mod.bind(this, output, false));
+    woutput.mod.forEach(mod.bind(this, output, false, data, diag, test));
     upFields.call(this, woutput, output);
     this._lastWith = woutput.stamp;
   }
 
   // Mods need to come after all removals have been run.
-  input.mod.forEach(mod.bind(this, output, true));
+  input.mod.forEach(mod.bind(this, output, true, data, diag, test));
   upFields.call(this, input, output);
 
   return output;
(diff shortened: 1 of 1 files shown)


Jerrythafast:
For example, with the following spec in the Editor:
```json
{
  "signals": [{"name": "limit", "init": 3}],
  "data": [
    {
      "name": "table",
      "values": [1, 2, 3],
      "transform": [
        {"type": "filter", "test": "datum.data < limit"},
        {"type": "cross"}
      ]
    }
  ]
}
```
`ved.view.data("table").values()` yields
```json
[
  {"a": {"data": 1}, "b": {"data": 1}},
  {"a": {"data": 1}, "b": {"data": 2}},
  {"a": {"data": 2}, "b": {"data": 1}},
  {"a": {"data": 2}, "b": {"data": 2}}
]
```
But `ved.view.signal("limit", 4).update().data("table").values()` yields
```json
[
  {"a": {"data": 1}, "b": {"data": 1}},
  {"a": {"data": 1}, "b": {"data": 2}},
  {"a": {"data": 2}, "b": {"data": 1}},
  {"a": {"data": 2}, "b": {"data": 2}},
  {"a": {"data": 3}, "b": {"data": 1}},
  {"a": {"data": 3}, "b": {"data": 2}},
  {"a": {"data": 3}, "b": {"data": 3}}
]
```
Clearly, the following tuples are missing:
```json
[
  {"a": {"data": 1}, "b": {"data": 3}},
  {"a": {"data": 2}, "b": {"data": 3}}
]
```

Jerrythafast:
Oh, snap. The issue is more complicated than this. The `cross` transform does not listen for signal changes on the `with` (right) data set *at all*.

```json
{
  "signals": [{"name": "limit", "init": 3}],
  "data": [
    {
      "name": "other",
      "values": [1, 2, 3],
      "transform": [
        {"type": "filter", "test": "datum.data < limit"}
      ]
    },
    {
      "name": "table",
      "values": [10, 20, 30],
      "transform": [
        {"type": "cross", "with": "other"}
      ]
    }
  ]
}
```
This spec suffers from the exact same problem.

arvind:
Shoot, this goes back to #395. In 591e76a, I simply skip tuples that appear in `input.mod` if they don't have a cached value. This doesn't account for filter expressions with signal values -- when the signal value changes, all tuples are sent forward in `input.mod`. 

If you'd like to amend this PR, here is one possible fix to the `mod` function in `Cross`: if a cached value is found, continue with the current logic of `mod`; if a cached value isn't found, return with a call to the `add` function. This will re-run the filter and add new tuples if they pass.

Jerrythafast:
I think this is what you had in mind. I hope it works :)

Jerrythafast:
I see, you were right. Fixed now.

Jerrythafast:
Earlier today I thought to have reproduced this same issue when crossing with another data set, but by the looks of it I made a mistake there since my MWE for that case was functioning correctly.

Although I do have the impression that the `add` call *does* solve a similar problem I was facing (which I will definitely test tomorrow), it does not solve *this* particular problem. I really think the `!selfCross && ` part of the `if` statement in the `batchTransform` method should go.

arvind:
Hi @Jerrythafast, I was thinking about this some more and realized that there was a deeper bug with Cross' handling of signals in parameter values. I've pushed a fix with 2e8f4a8. With the introduction of the `filter` parameter, any time a change in signal value occurs, the simplest thing to do is to purge the current cached tuples and re-cross all input tuples. 

The earlier fix I suggested was still flawed: while it would handling the case where _all_ crossed tuples for a particular input tuple were filtered out (i.e., `cache` was `null`), it would fail in the case that only _some_ crossed tuples had been filtered out (i.e., `cache` was *not* `null`). 

Could you try the latest master branch, and let me know if it's resolved the issues discussed in this issue/PR. Thanks!

Jerrythafast:
Hey @arvind, unfortunately the fix you've pushed has not resolved the issue I was having. I could find no difference in behaviour with the MWE from the second comment. With this MWE I could not get the `reset` parameter to turn `true`, so perhaps the problem has something to do with how Vega is determining whether a `reset` is required.

arvind:
Apologies @Jerrythafast. I totally misread the initial PR/issue. You're absolutely right, the `!selfCross` was the bug there and prevent self crosses from processing both the left and right side. This should now be fixed with beee811.

arvind:
Aah, those params are left over from before. Will remove them in a linter pass, thanks. As you noticed, there no longer is a call to "add" because "mod" should now only be called if the properties of input tuples have been modified. In this case, Cross simply needs to forward the corresponding crossed tuples (if they exist). 

If signal values change, Cross receives tuples in "input.mod" but instead calls the new "purge" logic. 

Hope this makes sense. And yes, this process was very helpful giving Cross  a more thorough review. Thanks for filing all the issues! 

-Arvind
-----
Arvind Satyanarayan <http://arvindsatya.com/>

arvind:
Actually, scratch that. If tuple values change, we still need to re-run them against the filter. So not only do we need to call `add`, in the case cache doesn't exist, but for cached tuples, we need to re-test them. Look for (another) commit soon. Phew!

arvind:
Oh boy, `filter` was more complex addition to Cross than I originally realized back with #362 :) With cf1f0d3, we handle the above cases that I discussed. This back-and-forth has been helpful to beef up our tests for the cross transform, so hopefully Cross is working much more reliably now! Thanks for working with us on this, @Jerrythafast!

Jerrythafast:
I should thank you for investing all this time in implementing the `filter` option and then patiently fixing all the bugs that it caused when used with various other Vega features, @arvind! It's great to wake up in the morning, grab my phone and see there is a pile of emails from GitHub about stuff that you worked on :)

PR #431, opened 2015-11-12
Report parsing errors to the caller

nyurik:
This is the replacement pull request instead of https://github.com/vega/vega/pull/402

It exposes the logging interface to allow for complex error processing, and an alternative spec.parse() with error handling.

Handle the case when a user-overriden logger throws an error. In case of an internal error, report it via callback(error, model)

TODO:
* This is a fairly bad workaround, and could really use promises.
* The error messages should be replaced with objects to allow
complex error processing.
* There are 3 usage models: (1) ignore errors, (2) stop on the first error, and (3) report all errors. The (1) is the default, the (2) can be implemented by throwing an error on log.error() call or with a new  "config.throwOnError" parameter, but the (3)  requires that each log.error() call is accompanied by the context object, so that logging knows which graph is causing an error.

### DIFF
--- a/src/parse/spec.js
+++ b/src/parse/spec.js
@@ -3,11 +3,58 @@ var dl = require('datalib'),
     Model = require('../core/Model'),
     View = require('../core/View');
 
-function parseSpec(spec, callback) {
-  var vf = arguments[arguments.length-1],
+/**
+ * Backward compatibility wrapper that accepts callback without error handler
+ * @param spec (object)
+ * @param callback (model)
+ * @param config (optional object)
+ * @param viewFactory (optional callback)
+ * @returns {*}
+ */
+function parseSpec(spec, callback /* [, config] [, viewFactory] */) {
+  var cb = callback,
+      vf = arguments[arguments.length-1],
       viewFactory = arguments.length > 2 && dl.isFunction(vf) ? vf : View.factory,
-      config = arguments[2] !== viewFactory ? arguments[2] : {},
-      model = new Model(config);
+      config = arguments[2] !== viewFactory ? arguments[2] : {};
+
+  return module.exports.parse(spec, config, viewFactory, function(err, model) {
+    // For backward compatibility, the error is thrown even though it might never be caught
+    if (err) throw err;
+    cb(model);
+  });
+}
+
+module.exports = parseSpec;
+
+/**
+ * Parse graph specification
+ * @param spec (object)
+ * @param config (optional object)
+ * @param viewFactory (optional function)
+ * @param callback (error, model)
+ */
+parseSpec.parse = function (spec, /* [config,] [viewFactory,] */ callback) {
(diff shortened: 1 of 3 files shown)


src/parse/spec.js
```diff
@@ -3,11 +3,58 @@ var dl = require('datalib'),
...
+module.exports = parseSpec;
+
+/**
+ * Parse graph specification
+ * @param spec (object)
+ * @param config (optional object)
+ * @param viewFactory (optional function)
+ * @param callback (error, model)
+ */
+parseSpec.parse = function (spec, /* [config,] [viewFactory,] */ callback) {
```
arvind: Hi @nyurik, thanks for the PR. I think I have an alternate formulation of the goals of this PR but before I propose it, I had a couple of questions. First, was there a specific reason you want to swap the order of arguments here? I'm trying to better understand the motivation here.

src/parse/spec.js
```diff
@@ -3,11 +3,58 @@ var dl = require('datalib'),
...
+  var cb = arguments[arguments.length - 1],
+      model, argInd = 2,
+      viewFactory = View.factory;
+
+  function done(err, value) {
+    if (cb) {
+      cb(err, value);
+      cb = undefined;
+    }
+  }
```
arvind: Second, why is `done` needed? Why not call `cb` directly where needed? (And perhaps throw an error if no cb was specified?) Moreover, why set the cb as `undefined` at the end? Thanks!

arvind:
Thanks for the PR, @nyurik. I've left you some comments on specific lines. We can work on making error processing more consistent because we alternate between `log.error` and throwing errors across the codebase.

src/parse/spec.js
nyurik: Hi @arvind, this is so that it follows nodejs standard - function async results (both success and errors) are always returned via a callback that is passed as the last parameter.  This way when I use Promises, it can automatically convert this function into a standard promise (e.g.  bluebird lib's promisify function).
In general, I am a big proponent of promises in general, as they greatly simplify treating asinc code in a sync fashion, reduce clutter, and allow easy error handling.

src/parse/spec.js
nyurik: done() prevents multiple accidental callback() calls - in case of an error.

nyurik:
@arvind, is it related to https://groups.google.com/forum/#!topic/vega-js/jAVfCqm91ZM ?  We should devise a general strategy on how to deal with the errors.   It would be a painful anti-pattern to have a function that takes a callback, while also being capable of throwing an exception. Either one or the other, so that it is more predictable, and requires less code to call. I am a big proponent of promises in general, as they greatly simplify treating asinc code in a sync fashion, reduce clutter, and allow easy error handling, so maybe we should consider that.

Thanks!

jheer:
Closing in favor of #447, which includes the commits from this PR.

PR #441, opened 2015-11-18
Keep Cross._ids in sync with Cross._cache

Jerrythafast:
Without this fix, input objects that are re-added after being removed were not re-crossed.

You can see this bug in action with the same MWE as was used with #428. Setting the value of `limit` to `4` now adds the crossed objects like you would expect (because #428 is fixed). But changing `limit` back to `3` and then back to `4` again would not re-cross those objects.

With this PR, the re-added input objects are now re-crossed as you would expect. It does not, however, re-use the previously created crossed tuples - it is creating new ones instead. If the old crossed tuples are still kept in memory somewhere this is actually a memory leak.


arvind:
Thanks Jerry!

PR #447, opened 2015-11-23
Update parse method signatures and error handling.

jheer:
This is an extended version of pull request #431.

- Standardize spec parse signature. Callback is now always the last argument.
- Support `cb(error, result)` callback signature. For backwards compatibility, invoke callback as `cb(result)` if `cb.length === 1`. 
- Update test cases and command line utils to use the updated signatures.
- Add node-canvas as an optional dependency. This proved necessary for test cases to run on a clean npm install, in part due to d3-cloud's setup.

__Going forward, parsing callbacks with the signature (error, result) are the preferred method. Documentation and dependent libs should be updated in response upon the next release.__

### DIFF
--- a/src/parse/data.js
+++ b/src/parse/data.js
@@ -7,12 +7,22 @@ function parseData(model, spec, callback) {
   var config = model.config(),
       count = 0;
 
-  function loaded(d) {
+  function onError(error, d) {
+    log.error('PARSE DATA FAILED: ' + d.name + ' ' + error);
+    count = -1;
+    callback(error);
+  }
+
+  function onLoad(d) {
     return function(error, data) {
       if (error) {
-        log.error('LOADING FAILED: ' + d.url + ' ' + error);
-      } else {
-        model.data(d.name).values(dl.read(data, d.format));
+        onError(error, d);
+      } else if (count >= 0) {
+        try {
+          model.data(d.name).values(dl.read(data, d.format));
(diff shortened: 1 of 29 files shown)


jheer:
Should be ready for merge! Assigning to @arvind for review, but @nyurik should also look and chime in with any additional thoughts / concerns.

nyurik:
@jheer, thanks, i will review the code shortly. From the comment, why do you need the  `cb(result) if cb.length === 1` -- if you pass a function that only has one argument, the second argument will be ignored, so most of the time it should be ok.  But this is up to you.

src/parse/data.js
```diff
@@ -7,12 +7,22 @@ function parseData(model, spec, callback) {
...
-        model.data(d.name).values(dl.read(data, d.format));
+        onError(error, d);
+      } else if (count >= 0) {
+        try {
+          model.data(d.name).values(dl.read(data, d.format));
+        } catch (err) {
+          onError(err, d);
+        }
       }
       if (--count === 0) callback();
```
nyurik: this line should be moved inside the try{}, right after line 22.

jheer: Good catch. Will push update shortly.

nyurik:
Thanks! Looks good to me :)
Any thoughts about overall error handling and promises direction? its in the email

PR #454, opened 2015-11-27
Update bounder reevaluation check.

jheer:
Generalize bounds update check by testing if element lies along internal edge of group bounding box. Moved to vega from vega/vega-scenegraph#20.

PR #464, opened 2015-12-03
More granular bounds recalculation.

arvind:
In lieu of #454. This does a full recalculation of mark bounds for line and area marks. For others, it runs itemBounds for `input.add` and `input.mod`. Mark bounds are recalculated by union-ing only if an added item falls outside previously calculated mark bounds, or if a modified item aligns with them.


jheer:
LGTM.

PR #466, opened 2015-12-04
Implemented scale function in expression language.

RussellSprouts:
This adds the functions scale and iscale to the expression language. (i)scale takes 2 or 3 parameters. The first is a scale name, the second it the value to pass to the scale, and the third (optional) is a group object to perform the lookup on.

Previously, the expressions had no direct access to the model, so I had to add model as a parameter when you call an expression. I believe that I updated everywhere we call an expression.

This requires https://github.com/vega/vega-expression/pull/3 to be merged first, so we should bump the dependency on vega-expression to whatever the new value is before merging this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,10 @@
     {
       "name": "Jeffrey Heer",
       "url": "http://idl.cs.washington.edu"
+    },
+    {
+      "name": "Ryan Russell",
+      "url": "https://github.com/RussellSprouts"
     }
   ],
   "license": "BSD-3-Clause",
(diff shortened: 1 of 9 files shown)


jheer:
I did a pass over this PR, and added new facilities for formatting functions as well as some test cases. If we are happy with the design changes made, I think it's good to go. The function signature for Vega expressions has now changed to include `model` as the first parameter. This raises some questions:

1. Is this the right way to go? Is there another (better?) way to make the model available? FWIW, I'm comfortable with this change, but want to make sure we vet it.

2. Now that we include the model, should we make passing the signal values optional? For example, if explicit signal values are not passed in to the function, we could query the model for them (though potentially at the cost of another level of nested function invocation). Of course, we could always make this change later, so that needn't stop us from merging now.

RussellSprouts:
Another way I could see making the model available is to instantiate an expression compiler in parseSpec with a reference to the model object. Every visualization would have its own parser. parseSpec would pass that expression parser down to each of the other parsers.

This might be better. model could keep a property like _signalValues, and we would assign it to the expression object, so it would be available as this.signals in the expressions. Then we wouldn't have to create the signal value object every time we want to evaluate expressions. We could remove the model and signal parameters to the expression entirely -- you would only pass the datum and event.

jheer:
Another twist on that idea is the following: have the model itself instantiate it's own expression compiler, and have the compiler be accessed through the model. This allows us to have the model implicitly available to expressions and reduces the number of parameters (and also require imports across the codebase). We could also (as you note) find ways to streamline signal access to avoid the nasty business of marshalling values all over the place.

jheer:
Superseded by #467, which implements the ideas discussed here by @RussellSprouts and I.

PR #467, opened 2015-12-06
Expanded and streamlined expression parsing.

jheer:
Extends #466 with implicit model and signal value access.
Removes need to marshall signal values prior to expression invocation.

### DIFF
--- a/src/parse/expr.js
+++ b/src/parse/expr.js
@@ -1,4 +1,5 @@
-var expr = require('vega-expression'),
+var template = require('datalib').template,
+    expr = require('vega-expression'),
     args = ['model', 'datum', 'event', 'signals'];
 
 module.exports = expr.compiler(args, {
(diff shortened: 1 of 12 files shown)


arvind:
This looks good to me except for two minor points: (1) having expressions reach into model + signal private variables (though I can't think of anything better that doesn't incur a performance penalty...) (2) the fact that `this.defs` maps to `functionDefs` -- can we consistently name the variables (this would require an upstream change to vega-expression as well).

src/parse/expr.js
```diff
@@ -1,17 +1,91 @@
...
+    expr = require('vega-expression'),
     args = ['datum', 'event', 'signals'];
 
-module.exports = expr.compiler(args, {
+var compile = expr.compiler(args, {
   idWhiteList: args,
   fieldVar:    args[0],
-  globalVar:   args[2],
+  globalVar:   function(id) {
+    return 'this.sig["' + id + '"]._value';
```
arvind: Use `dl.str` to safeguard against crazy signal names (e.g., with quotes)?

jheer:
Yes, the signal private variable access breaks encapsulation but is fast. We could benchmark against method access as well, but that involves two method calls, one for signal lookup and another to access the value. I'm not too concerned with `this.defs` - unfortunately sometimes JS performance is influenced by name and method lengths, plus 'functionDefs' is a bit unwieldy.

arvind:
Re: `functionDefs`, I imagined simply renaming the property to `defs` for consistency with how its referred to within the codegen. But, it's a sufficiently minor point that I merged the PR anyways.

Issue #372, opened 2015-09-29
Deprecate predicates in favour of expressions


arvind:
Our expression language is almost sufficiently powerful to deprecate predicates. The remaining expressivity gap centers around functionality provided by `in` predicates: (1) ranges over ordinal scales, and the simplicity of specifying min/max for quantitative values/scales; (2) extensions/checking for membership within a dataset. 

One option would be to introduce functions to address each of these:

* `in_range(val, a, b, scale)` Tests is `val` is between the `a` and `b`. If `scale` is provided, performs the necessary scale inversion. In the case of ordinal scales, this would check if `val` exists in the set returned by the inversion. 

* `in_data(val, data, field)` Tests if the data set named `data` contains a record whose `field === val`.

arvind:
An open question: would we want to introduce a top-level `expressions` property to enable named expression reuse? Presumably yes. Expression properties in the rest of the specification (e.g., `filter.test`, `formula.expr`, `streams.expr`, etc.) could then be either defined inline or reference a named expression (e.g., `"expr": "datum.a + 1"`, `"expr": {"name": "isSelected"}`). 

Similarly, production rule branches could define the conditions for each branch inline, or invoke a named expression. For example, `"rule": [{"test": "datum.x > 5", "value": "grey"}, {"test": {"name": "isSelected"}, "value": "yellow"}]`.

arvind:
One downside: moving predicates into the expression language makes it more difficult to statically analyze them, e.g., as part of @jhoffswell's work. This could be ameliorated by exposing the AST of the parsed expressions.

kanitw:
>  One downside: moving predicates into the expression language makes it more difficult to statically analyze them, e.g., as part of @jhoffswell's work. This could be ameliorated by exposing the AST of the parsed expressions.

This might also make it a little harder (though not impossible) to develop UI for manipulating the expression.   Vega-lite team are facing the same issue whether we should have “aggregate” and “timeUnit” properties in the encoding spec.

arvind:
@kanitw, Expressions are pretty pervasive in a Vega specification already, so higher-level tools will likely already need to develop a UI to support them, independent of this issue. I'm not sure I understand why adding two new methods would make developing such a UI harder? Could you elaborate?

Similarly, not sure I understand the concern with `aggregate` and `timeUnit`? Do you use predicates or expressions in there?

kanitw:
[...]

PR #478, opened 2015-12-19
Add functionality to allow deprecation of predicates

RussellSprouts:
I've been working on implementing the changes necessary to have all the functionality of predicates using expressions (Issue #372 )

This adds an indata function, which takes indata(dataname, value, [field]). It returns true if a point in the data source has the given value under the given field. If field is not provided, it uses the entire data point.

I also added the property `test` which is an expression for a data source clear modification. This clears the data source when the expression becomes true.

For production rules, instead of a predicate property, you can provide a `test` property. This expression is evaluated for each datum instead of a predicate.

I still need to finish implementing the getIndex and hasElement functions in vega-dataflow. getIndex([field]) will create a map of all the values of the given field to the number of times they appear. hasElement returns true if the count is greater than 0. If the field is not provided, then the entire data point is used.

I still need to add the functionality of inrange for ordinal scales, and then I believe all the functionality of predicates will be implemented.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,8 +41,8 @@
     "d3-cloud": "^1.2.1",
     "datalib": "^1.5.7",
     "topojson": "^1.6.19",
-    "vega-dataflow": "^1.3.2",
-    "vega-expression": "^1.1.0",
+    "vega-dataflow": "^1.4.0",
+    "vega-expression": "^1.2.0",
     "vega-logging": "^1.0.1",
     "vega-scenegraph": "^1.0.16",
     "yargs": "^3.30.0"
--- a/src/core/Model.js
+++ b/src/core/Model.js
@@ -19,6 +19,7 @@ function Model(cfg) {
   this._builder = null; // Top-level scenegraph builder.
 
   this._reset = {axes: false, legends: false};
+  this._requestedIndexes = [];
 
   this.config(cfg);
   this.expr = compiler(this);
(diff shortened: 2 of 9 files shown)


RussellSprouts:
OK. Once [the vega-dataflow merge request](https://github.com/vega/vega-dataflow/pull/7) is merged, and we update vega's dependencies on vega-expression and vega-dataflow, this should be good to merge.

arvind:
Thanks @RussellSprouts. This LGTM for the most part. A few points:

1. I don't see any calls to a `hasElement` function you described above. So, with the current PR, I believe `field` would be a required parameter? 

2. Do you intend to submit a separate PR for `inrange`? Note, if a scale is specified, `inrange` should simply call its `invert` function. This is most useful for ordinal scales, but shouldn't be restricted to only them. 

3. In `properties.js`, can we attach `exprs` to the `encoder` function? I think this would be simpler, and would not create an unnecessary closure. We could also define a reference to it at the start of the encoder function (like we do with `this.tpl.set`) so that we don't have references to `this` littered across the encoder codegen. 

4. I now realize we never had unit tests to (statically) test that rules work. Do you mind adding either a few simple unit tests, or specs under `test/specs`, that verify that the new `test` property of rules works correctly? Thanks!

RussellSprouts:
I just realized that data sources will not be parsed when we are parsing an expression, so I queue the indexes to be created just after parsing is finished instead.

1. hasElement was a method on the datasource that checked the index. I removed it and just used getIndex instead. field is a required parameter now, because data points are always objects.

2. inrange already works for numeric ranges. I can add handling for scales.

3. Done

4. I can add some unit test.

PR #489, opened 2016-01-09
Themes

emilygu:
Done:
- pulled out x/y axis so that they can be separately specified in the config (config.axis for both, config.axis_y for only y axis, config.axis_x for only x axis)
- add 'baseline' to config.legend (top, middle, bottom)
- allow for config.legend.symbolShape to take an object in order to specify different shapes for different types of graphs, for example:
  {
    line: 'circle',
    symbol: 'diamond',
    rect: 'circle'
  }
- add tickPlacement to config.axis so that ticks for the x-axis can be placed such that the label is between ticks or on ticks (between, on) -- better name?
- can specify 'default' instead of a color in the spec for marks.  This takes the color given to the newly added field configProperty in config.marks (default is black if none specified) --better name?
- added width, height, padding and background as fields that can be specified in config
- added padding and outerPadding as fields that can be specified in config.scales
- added gridWidth, tickSizeEnd, grid and layer as field to config.axis

TODO:
- bug for group marks where setting the default color for marks will set it for all marks (like axis since all of these are also turned into marks)
- make sure current code passes existing test suite
- write new tests


emilygu:
Done:
- Added config files for excel, ggplot, vox, quartz
- Added specs for arc, bar, group_bar, line, scatter, scatter_matrix
- theme/output_(config types) contains screenshots of of different types of graphs for the given config
- Added symbolSize to config.marks
- Add grid on/off for different graph types
- Add grid dash
- Add markType to model to keep track of graph type

Differences:
- padding based on type of group charts
- grid based on orientation of bar charts
Vox
	- legend should be on bottom

Quartz
	- x axis should protrude from top
	- legend should be at top
	- scatter
		- dot with white circle
	- legend titles match color

GGPlot
	- extra padding between bars and axis on all sides
	- minor gridlines in between
	- legend need gray background
	- arc
		- legend squares larger and right next to each other

Todo:
- Determine whether modifying model._config is ok (must change for tests since change of model modifies subsequent test cases)
- For tests, find better way of passing in config in legend

src/scene/axis.js
```diff
@@ -407,19 +417,52 @@ function axisTicksExtend(orient, ticks, oldScale, newScale, size, offset) {
   } else {
     size = {value: sign * size, offset: offset};
   }
+
+  var tickPlacement = config_axis.tickPlacement;
```
arvind: @emilygu: Could you explain to me what this block of code is doing? I'm trying to understand why a particular axis needs to check other registered scales?

src/scene/axis.js
```diff
@@ -407,19 +417,52 @@ function axisTicksExtend(orient, ticks, oldScale, newScale, size, offset) {
...
+        currOuterPadding = config_scale.outerPadding;
+      }
+
+      var oldPadding = ((oldScale.offset * 2 * currPadding) / ((1 - currPadding) * 2));
+      var newPadding = ((newScale.offset * 2 * currPadding) / ((1 - currPadding) * 2));
+
+      updatedOldScale.offset = ((oldScale.offset - 1) * 2) + oldPadding;
+      updatedNewScale.offset = ((newScale.offset - 1) * 2) + newPadding;
+    }
+
```
arvind: @emilygu: Could you explain to me what this block of code above is doing? I'm trying to understand why a particular axis needs to check other registered scales?

src/parse/mark.js
```diff
@@ -23,6 +36,17 @@ function parseMark(model, mark) {
...
+function defaultColor(property, configProperty, prop, config) {
+  if (property in prop && 'value' in prop[property] && prop[property]['value'] === 'default') {
+    if (typeof config !== 'undefined' && 'marks' in config && configProperty in config.marks) {
+      prop[property]['value'] = config.marks[configProperty];
+    } else {
+      prop[property]['value'] = '#000000';
+    }
+  }
+}
+
```
arvind: @emilygu: A couple of questions here:

1. Is there a reason you named this property `config.marks.defaultFill` rather than simply `config.marks.fill`? 
2. Is there a reason they are only applied when the property is set to `{"value": "default"}` rather than automatically applied if no fill color is specified? I would think that if no fill color is specified, the default one is automatically used; and if users explicitly want _no_ fill color specified, and have a theme applied, they can set the fill to `transparent`?

src/scene/Scale.js
```diff
@@ -171,6 +173,13 @@ function ordinal(scale, rng, group) {
   if (!scale.invert && spatial) invertOrdinal(scale);
 }
 
+function setPadding(pad, config, defaultPad, parent_id, datum) {
+  if (!parent_id && config && config.scale && config.scale[pad] && !datum._facetID) {
+    return config.scale[pad];
+  }
+  return defaultPad;
+}
+
```
arvind: @emilygu: If I understand this, the test for `parent_id` and `datum._facetID` ensures that a theme's scale paddings are only applied to the scales defined at the top-level (i.e., not nested within groups). Is that correct? If so, was this a design pattern you frequently saw? If yes, then I would say we want to break these out to separate padding variables. It should still be possible to set the `padding` and `outerPadding` for _all_ scales in a theme.

src/scene/axis.js
emilygu: @arvind This chunk of the code allows the user to set ordinal tick marks to be in between the marks instead of on it.  
For example, this is the default
screen shot 2016-05-20 at 12 49 37 pm
But this chunk of code allows for 
screen shot 2016-05-20 at 12 52 09 pm

In order to figure out the offset of the ticks from the original position, I had to grab the current scale, since the padding and outerPadding gave me the information so I could calculate how far to offset ticks.

arvind: Thanks @emilygu, the screenshots are particularly helpful! I think this code block can be simplified as the `scale` variable is within the closure scope, so we don't need to iterate through all scales. Finally, I'm also curious about why this is within an `if` block that checks for `top` or `bottom` orientation? Can't the user specify a `tickPlacement` of `between` for left/right-oriented y-axes? Or is this a simple oversight? :)

emilygu: @arvind Oh yes, haha, thanks!  I was meaning to add that in but I guess I forgot.  Thanks for the catch!  I'll update the pull request with that fix

floribon: Take care of this `console.log` line 438 while you're there :-)

emilygu: @floribon Yup, just caught that one :)  Thank you!

src/parse/mark.js
emilygu: @arvind 
1. I named it defaultFill since I wanted to distinguish it from the regular fill that could be specified in the json spec.  I don't feel super strongly about having it be named defaultFill and not fill, it was more of a small decision I made along the way.  So if fill seems better to you, I can go ahead and change it!
2. I actually like that idea much more.  When I thought of default fill, my original thought was to have people specify 'default' which was why I went with that.  But agree that it makes more sense for users to be able to draw marks without specifying a color, in which case the default fill would be used instead.  Will work on changing that.

arvind: Yeah, let's go with `config.marks.fill`. That keeps it consistent with the property name it provides a default for (`fill`) if the user hasn't explicitly provided one. Thanks!

src/scene/Scale.js
emilygu: @arvind Yes, I believe so.  I had to add that in or else it was padding individual items in group charts.  This is the only place that this is used though.

src/parse/mark.js
```diff
@@ -1,10 +1,17 @@
...
   var props = mark.properties,
       group = mark.marks,
-      config = model._config;
+      config = model._config,
+      arglen = arguments.length,
+      argidx = 2,
+      isMark = false;
+
+  if (arglen > argidx && dl.isBoolean(arguments[arglen - 1])) {
+    isMark = true;
```
arvind: Hi @emilygu, if I follow, `isMark` is there so that we don't set any defaults for axis and legend items? If so, we'll want to make sure we're passing `true` in for the recursive call to `parseMark` in this file (line 59).

src/scene/Scale.js
arvind: I see, thank you! I am going to remove this for now because I think the behaviour it introduces is confusing -- two scale definitions that are otherwise identical but one at the top-level and one nested within a group would produce different visual output. More concretely, you may have a bar chart with a padded scale at the top-level, but if it were nested within a group (e.g., as part of a multi-view display), it would suddenly appear different. 

Could you open a new issue to track this for us? In particular, clearly describe what the motivation and goals are and include screenshots to illustrate examples. I wonder if there's an alternate mechanism that we can implement that provides more consistent behaviour.

arvind:
Good work @emilygu! 🎉🎉🎉

PR #495, opened 2016-01-12
Fix titleOffset property

floribon:
The extra `y` and `x` properties should be within a `{value: n}` object
rather than plain integer in order to work.

-----

To reproduce use the `bar` example of vega editor and add a `titleOffset: 50` to one of the axes. The parameter is not taken into account and the title has an offset forced to 0.


arvind:
Thanks @floribon!

Issue #477, opened 2015-12-16
Support color ramps for ordinal scale


kanitw:
From a conversation on slack:

> One option might be to support “named” color ramps (similar to Tableau and ColorBrewer). A Vega spec might include the named palette as the scale range. Then, based on the scale cardinality, an appropriate subdivision of colors could be chosen (e.g., 5/7/9). If the cardinality exceeds the largest pre-designed palette, we might fall back on interpolation.

For now, at least having interpolation would be great for VL!

domoritz:
I don't see how a rank transform alone is enough to support color ramps. 

Yes, we can use the rank transform to correctly set the colors of marks with a linear scale. However, if we also have a legend, we cannot use the linear scale because it produces the wrong kind of legend. 

```json
{
  "width": 200,
  "height": 200,
  "padding": "auto",
  "data": [
    {
      "name": "source",
      "values": [
        {"a": "A","x": 1,"y": 1},
        {"a": "B","x": 2,"y": 2},
        {"a": "C","x": 3,"y": 3},
        {"a": "A","x": 4,"y": 4},
        {"a": "B","x": 5,"y": 5},
        {"a": "C","x": 6,"y": 6}
      ],
      "transform": [
        {
          "type": "rank",
          "field": "a",
          "normalize": true
        }
      ]
    }
  ],
  "marks": [
    {
      "name": "root",
      "type": "group",
      "from": {"data": "layout"},
      "properties": {
        "update": {"width": {"value": 200},"height": {"value": 200}}
      },
      "marks": [
        {
          "type": "symbol",
          "from": {"data": "source"},
          "properties": {
            "update": {
              "x": {"scale": "x","field": "x"},
              "y": {"scale": "y","field": "y"},
              "size": {"value": 30},
              "shape": {"value": "circle"},
              "strokeWidth": {"value": 2},
              "opacity": {"value": 0.7},
              "stroke": {"scale": "color_rank","field": "rank"}
            }
          }
        }
      ],
      "scales": [
        {
          "name": "x",
          "type": "linear",
          "domain": {"data": "source","field": "x","sort": true},
          "rangeMin": 0,
          "rangeMax": 200,
          "round": true,
          "clamp": true,
          "nice": true
        },
        {
          "name": "y",
          "type": "linear",
          "domain": {"data": "source","field": "y","sort": true},
          "rangeMin": 200,
          "rangeMax": 0,
          "round": true,
          "clamp": true,
          "nice": true
        },
        {
          "name": "color",
[...]

PR #496, opened 2016-01-12
Adding rank transform.

arvind:
Implementing a simple `rank` transform to facilitate #477. Supported parameters include `start`, `step`, and `normalize`. The first two default to `1`. If the latter is true, the output `rank` field falls between [0, 1] (i.e., start = 0, step = 1/data.length).

### DIFF
--- a/src/transforms/Rank.js
+++ b/src/transforms/Rank.js
@@ -6,8 +6,6 @@ var Tuple = require('vega-dataflow').Tuple,
 function Rank(graph) {
   BatchTransform.prototype.init.call(this, graph);
   Transform.addParameters(this, {
-    start: {type: 'value', default: 1},
-    step:  {type: 'value', default: 1},
     field: {type: 'field', default: null},
     normalize: {type: 'value', default: false}
   });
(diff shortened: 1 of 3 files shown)


arvind:
Also added a `field` param with which ranking can be determined. This prevents an unnecessary aggregate+rank+lookup to wire up scales for Vega-Lite. If `field` is not specified, the tuples are simply ranked in observed order.

jheer:
Overall LGTM. My only question concerns the custom `start` and `step` parameters. Are those really needed? First, they create a bit of conditional logic that interacts with the `normalize` parameter. Second, they don't seem strictly necessary, as one could use a downstream formula instance to post-process the rank output. Third, in the case of normalization incrementing fractional steps is more prone to floating point error that doing per-element division. My leaning would be to drop the start/step parameters and perform per-element normalization. Feel free to merge once you've decided how you want to proceed and made any changes.

Note: for an example of the third point, try running this in the console in Chrome:
```
for (var x=0, y=0; y<=10; x+=0.1, y+=1) {
  console.log(y/10, x);
}
```

src/transforms/Rank.js
```diff
@@ -0,0 +1,74 @@
...
+prototype.batchTransform = function(input, data) {
+  log.debug(input, ['rank']);
+
+  var rank  = this._output.rank,
+      norm  = this.param('normalize'),
+      field = this.param('field').accessor,
+      l = {}, i = 0, len = data.length, r, v, d;
+
+  for (; i<len; ++i) {
+    r = norm ? (i+1)/len : i+1;
```
domoritz: Is this supposed to be divided by len or by the number of distinct values? If you have a dataset like [a,b,c,a,b,c], then the ranks are [0.166666, 0.3333333, 0.5, 0.166666, 0.3333333, 0.5]. I mean I can work with this, I just wonder whether this is intended behavior.

arvind: Good catch. Fixed with ecf225d.

PR #498, opened 2016-01-13
Fix error in panzoom example

eibanez:
Example won't work well with mice wheel that return a constant event.deltaY. In these cases, only the first time you roll the wheel in a direction is captured by the signal. This is easily fixed by adding a random number to the signal.

### DIFF
--- a/test/spec/panzoom_points.json
+++ b/test/spec/panzoom_points.json
@@ -49,7 +49,7 @@
       "name": "zoom",
       "init": 1.0,
       "streams": [
-        {"type": "wheel", "expr": "pow(1.001, event.deltaY)"}
+        {"type": "wheel", "expr": "pow(1.001, event.deltaY) + random()/100"}
       ]
     },
     {


RussellSprouts:
Good catch! You're right that the spec is broken, and this fix would work, but there is actually a simpler solution. I don't know if it is documented, but signals have a property "verbose." When verbose is set to true, signals fire even if the new value is the same as the previous. I noticed the same issue when I was testing panzoom in Firefox, which uses integers instead of Chrome's floats for deltas.

There's another issue with scroll events -- the [deltaMode](https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent/deltaMode) property. The delta value in a wheel event may be either in terms of pixels, lines, or pages, and the operating system/browser combination decides which.

The vega-editor project has an [updated version](https://github.com/vega/vega-editor/blob/master/app/spec/vega/panzoom_points.json) of the panzoom spec that fixes these issues. The version in vega proper is just here for the unit tests, but we can copy the new version to vega for consistency.

RussellSprouts:
OK.  4a475d7 uses the updated version of the spec from vega-editor.

PR #501, opened 2016-01-14
Don’t need to convert data to date time before formatting

domoritz:
Datalib takes care of converting timestamps.


domoritz:
@jheer, can you look at this when you have a minute?

jheer:
Merged via command line.

PR #511, opened 2016-02-02
Report invalid transformations

nyurik:
Report when transformation type is not valid (simplifies debugging)


arvind:
Thanks @nyurik!

Issue #513, opened 2016-02-04
JSON Schema is incomplete


okal:
Validating a spec against the schema fails where code uses `category20b` or `category20c` in a category ordinal scale.

PR #514, opened 2016-02-04
Add category20b and category20c to permissible range values

okal:
Closes https://github.com/vega/vega/issues/513


arvind:
Thanks @okal!

PR #516, opened 2016-02-09
Emily/svg shapes [wip]

emilygu:
- SVG template to create custom shapes using svg path
- If multiple commands that are the same (e.g. M, L, M, L), append numbering (e.g. M0, L0, M1, L1)
- {{...}} represent values that must be evaluated
- Hack to allow for variable usage
- Can look at shape_test/spec for more examples

Todo:
- Using a 'variable' field within shapes for users to directly type in javascript variables they want.  Want to try looking into using vega formula instead
- Create better test cases potentially
- Look into fill/no fill options


emilygu:
Moved to emily/custom-shapes

PR #521, opened 2016-02-12
Remove half a pixel offset from axes

domoritz:
Fixes vega/vega-scenegraph#22

before
screen shot 2016-02-11 at 21 49 43

after
screen shot 2016-02-11 at 21 49 23

I also looked through the vega examples and the axis positioning looks generally better.


arvind:
I suspect the 0.5 pixel offset was introduced for anti-aliasing so I'm assigning to @jheer to double check.

PR #527, opened 2016-02-15
Support setting custom cursors

### LINKED ISSUE #314: Add support to set mouse cursor.
It would be nice to be able to change the mouse cursor upon hover of interactive elements or during interactive operations (drag).

Possibilities:
- Add a `cursor` encoding property to marks. This could propagate down for the scenegraph to handle.
- Add a default `cursor` property to top-level spec? (Really, for the root group mark.)
- Let these properties be driven by signals to support interactive update.

Questions:
- When in the midst of performing an interactive operation (drag, brush resize), how might one ensure that the cursor state is sticky (e.g., not overridden by the cursor setting for a mark that the mouse is passing over)?

arvind:
Resolves #314 with a new named signal `cursor`. By default, if no `cursor` signal is explicitly specified in the specification, the cursor signal is set to the `cursor` property of a the scenegraph item underneath the cursor. A signal handler then takes this value and sets the `cursor` style of the `body`.

If the user specifies a `cursor` signal in their specification, they can interactively specify the cursor style. They are then entirely responsible for managing this signal and the default behaviour described above is ignored. To revert back to the default behaviour, users must set the `cursor` signal to `default`. 

See this spec for an example. Hovering over the rects should switch the cursor to `pointer`. If you `mousedown`, the cursor switches to `ne-resize` and stays this way until you `mouseup`.

```json
{
  "width": 400,
  "height": 200,
  "signals": [{
    "name": "cursor",
    "streams": [
      {"type": "mousedown", "expr": "'ne-resize'"},
      {"type": "mouseup", "expr": "'default'"}
    ]
  }],
  "data": [
    {
      "name": "table",
      "values": [
      ]
    }
  ],
  "scales": [
    {
      "name": "x",
      "type": "ordinal",
      "range": "width",
    },
    {
      "name": "y",
      "type": "linear",
      "range": "height",
      "domain": {"data": "table", "field": "y"},
      "nice": true
    }
  ],
  "axes": [
    {"type": "x", "scale": "x"},
    {"type": "y", "scale": "y"}
  ],
  "marks": [
    {
      "type": "rect",
      "from": {"data": "table"},
      "properties": {
        "enter": {
          "width": {"scale": "x", "band": true, "offset": -1},
          "y2": {"scale": "y", "value": 0}
        },
        "update": {
        },
        "hover": {
          "fill": {"value": "red"},
          "cursor": {"value": "pointer"}
        }
      }
    }
  ]
}
```

### DIFF
--- a/src/parse/spec.js
+++ b/src/parse/spec.js
@@ -54,10 +54,11 @@ var dl = require('datalib'),
           height  = spec.height || 500,
           padding = parsers.padding(spec.padding);
 
-      // create signals for width, height and padding
+      // create signals for width, height, padding, and cursor
       model.signal('width', width);
       model.signal('height', height);
       model.signal('padding', padding);
+      cursor(spec);
 
       // initialize model
       model.defs({
@@ -84,6 +85,20 @@ var dl = require('datalib'),
     });
   } else {
     onError('INVALID SPECIFICATION: Must be a valid JSON object or URL.');
+  }
+
+  function cursor(spec) {
+    var signals = spec.signals || (spec.signals=[]),  def;
+    signals.some(function(sg) {
+      return (sg.name === 'cursor') ? (def=sg, true) : false;
+    });
+
+    if (!def) signals.push(def={name: 'cursor', streams: []});
+
+    // Add a stream def at the head, so that custom defs can override it.
+    def.streams.unshift({
+      type: 'mousemove', expr: '{default: eventItem().cursor}'
+    });
   }
 }
 
--- a/src/parse/spec.js
+++ b/src/parse/spec.js
@@ -1,7 +1,7 @@
-var dl = require('datalib'),
+var dl  = require('datalib'),
     log = require('vega-logging'),
     Model = require('../core/Model'),
-    View = require('../core/View');
+    View  = require('../core/View');
 
 /**
  * Parse graph specification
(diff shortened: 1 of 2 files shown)


src/parse/spec.js
```diff
@@ -69,21 +64,38 @@ var dl = require('datalib'),
...
+      return (sg.name === 'cursor') ? (def=sg, true) : false;
     });
-  } else {
-    onError('INVALID SPECIFICATION: Must be a valid JSON object or URL.');
+
+    if (!def) signals.push(def={name: 'cursor', streams: []});
+
+    // Add a stream def at the head, so that custom defs can override it.
+    def.streams.unshift({
+      type: 'mousemove', expr: '{default: eventItem().cursor}'
```
jheer: What are the performance implications associated with updating the DOM style on _every_ mousemove event? (mousemove -> signal change -> listener -> DOM body style update)

jheer:
Merging now, but do see my note on mousemove updates. Perhaps there is another alternative (e.g., mouseover / mouseout)?

arvind:
I used mousemove because mouseover/out was giving me inconsistent results (e.g., with the above spec, `pointer` would only get set after brushing over the bars a few times). I haven't had time to debug further, but you point is well taken. e7f51b6 updates the cursor default expression to reuse the object if possible. This should ensure that although we will still reevaluate the expression on every mousemove, the rest of the chain (signal change -> listener -> DOM body style update) will only occur on actually different cursor styles.

PR #546, opened 2016-03-03
Add background to overview to enable interaction.

domoritz:
Without this fix you can only brush on the area mark.


arvind:
Thanks, good catch!

Issue #570, opened 2016-05-18
Legend should support opacity


YuhanLu:
Legend should support with opacity
This issue is linked to vega-lite 1349. https://github.com/vega/vega-lite/issues/1349#issuecomment-219920718

jheer:
Addressed in #572, pending merge.

PR #572, opened 2016-05-18
Legend Updates

jheer:
+ Fix legend scale collection bug in GroupBuilder.
+ Add support for opacity channel in legends.
+ Add new positioning options for legends (top-left, bottom-right, etc).

Opacity was requested by the Vega-Lite team (@kanitw, @YuhanLu). See #570.
Position options requested by @jpocom for example generation.

### DIFF
--- a/src/parse/legends.js
+++ b/src/parse/legends.js
@@ -9,10 +9,11 @@ function parseLegends(model, spec, legends, group) {
 
 function parseLegend(def, index, legend, group) {
   // legend scales
-  legend.size  (def.size   ? group.scale(def.size)   : null);
-  legend.shape (def.shape  ? group.scale(def.shape)  : null);
-  legend.fill  (def.fill   ? group.scale(def.fill)   : null);
-  legend.stroke(def.stroke ? group.scale(def.stroke) : null);
+  legend.size   (def.size    ? group.scale(def.size)    : null);
+  legend.shape  (def.shape   ? group.scale(def.shape)   : null);
+  legend.fill   (def.fill    ? group.scale(def.fill)    : null);
+  legend.stroke (def.stroke  ? group.scale(def.stroke)  : null);
+  legend.opacity(def.opacity ? group.scale(def.opacity) : null);
 
   // legend orientation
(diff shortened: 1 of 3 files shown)


arvind:
What is the expected behaviour for `top-left`, `bottom-right`, etc.? I'm seeing the following, which positions the legend within the group, rather than outside of it. I'm unsure if this is intended or not? E.g., I would assume that `top-left` and `top-right` are synonymous for `left` and `right`. If this is intended behaviour, are there other position names we can think of that are clearer?

`top-left`:
screenshot 2016-05-19 11 12 13

`top-right`:
screenshot 2016-05-19 11 12 21

`bottom-left`:
screenshot 2016-05-19 11 12 31

`bottom-right`:
screenshot 2016-05-19 11 12 37

`left`:
screenshot 2016-05-19 11 13 30

`right`:
screenshot 2016-05-19 11 13 36

jheer:
These are all intended to be *inside* the group. They are not synonymous
with any existing orientation setting. Here the offset controls both
horizontal and vertical offset from the corner of the data frame.

This is a common pattern in many plots (including scientific figures). We
need to generate these kinds of figures as part of Jorge's research. As in
other tools (Excel, etc) here it is up to the author to ensure that the
legend does not interfere with the data.

FWIW, I think we should keep this an "unofficial" feature for now, not part
of the official Vega schema.

kanitw:
> opacity channel in legends

I assume the opacity channel would be a part of the schema?  (Otherwise, we can't create an example that passes the test in VL.)

jheer:
Yes, opacity would be in the schema, but we could leave out the new legend position settings for now.

arvind:
Thanks all, merged!

kanitw:
Skimming through the PR, I'm not sure which line adds the `opacity` to legend's schema?

arvind:
I added `opacity` to the schema after merging the PR in 4f22b59.

YuhanLu:
May be we need to add `opacity` in `anyOf` also?

arvind:
Good catch @YuhanLu. Fixed with b50bdae.

PR #575, opened 2016-05-20
[WIP] Update dependencies.

domoritz:


domoritz:
Jsdom really wants node 4.0 onwards.

arvind:
Yeah, I've been considering dropping JSDOM from the tests and relying on vega-scenegraph. The reason I've not is because it's caught some bugs in the past that the scenegraph has missed.

PR #576, opened 2016-05-21
Add mouseup signal to brush_end in overview+detail

kevinmickey:
With larger datasets, things get sluggish (as expected).  With this slow-down, a quick mousedown, mousemove, and mouseup by the user produces a much smaller brush region than desired, missing the true brush_end.  If you include mouseup as a signal in brush_end, it corrects this.


arvind:
Thanks @kevinmickey!

Issue #590, opened 2016-06-19
Incorrect stacked area when having null categories.


kanitw:
spec: http://pastebin.com/ThRcjRFG
data: https://gist.githubusercontent.com/kanitw/59986cffa2f9bf36871ca07ab3e28c3e/raw/50b22fa5dfd3b5794de62856a2b958c1103b8dc6/data2.json

The `cat` field mixes null, numbers and strings.  This is originally created to produce `NaN` if we parse `cat` as `number` for debugging  [this PR](https://github.com/vega/datalib/pull/75).  

Now the original problem is solved.  However, I tried changing `parse` from `number` to `string` and it produces incorrect stacking as shown below:

![vega_editor](https://cloud.githubusercontent.com/assets/111269/16174871/dabfe080-358c-11e6-8901-1f94abdebac4.png)

Note that I inspect the parsed data and found that `1` in `cat` remains `1` (a number).

![developer_tools_-_http___localhost_1234__mode_vega](https://cloud.githubusercontent.com/assets/111269/16180046/8a03cbb0-362d-11e6-887d-4011ff803eaf.png)

When I replace `1` with `"1"`, it no longer creates the problem.  

data: https://gist.githubusercontent.com/kanitw/59986cffa2f9bf36871ca07ab3e28c3e/raw/f2d9754f6cf3e01df7a3d811bb4dec6956df7982/data2b.json

spec: the same, just replace data url with the new one

output: 

![vega_editor](https://cloud.githubusercontent.com/assets/111269/16180058/f289e476-362d-11e6-8412-6c90bb48a690.png)

jheer:
This is not a Vega bug, and not due to null values. The problem is the comparison of data with different types (1 and 'b', for example). For example, `1 < ' b' == false` AND `1 > 'b' == false`, so a type-preserving comparator will treat them as equivalent. (Normal JS sort does not exhibit this problem because it string-coerces everything.) I think a reasonable solution may be to avoid this situation altogether. For starters, it looks like we should adjust datalib parsing to force string coercion when the 'string' type is specified.

jheer:
Addressed in vega/datalib commit [7a9439a](https://github.com/vega/datalib/commit/7a9439a80d9542ab20ca0563b5bc8e2591afc47f), slated for datalib 1.7.1 release.

Issue #584, opened 2016-06-07
scale.domain.sort=true can work improperly


kanitw:
<img width="263" alt="vega_editor" src="https://cloud.githubusercontent.com/assets/111269/15845150/f6a6bd9e-2c24-11e6-9c94-621466095ad6.png">

```
{
  "width": 1,
  "height": 1,
  "padding": "auto",
  "data": [
    {
      "name": "source",
      "url": "data/movies.json",
      "format": {"type": "json","parse": {}},
      "transform": []
    },
    {
      "name": "summary",
      "source": "source",
      "transform": [
        {
          "type": "aggregate",
          "groupby": ["Major_Genre"],
          "summarize": {"*": ["count"]}
        }
      ]
    },
    {
      "name": "layout",
      "source": "summary",
      "transform": [
        {
          "type": "aggregate",
          "summarize": [{"field": "Major_Genre","ops": ["distinct"]}]
        },
        {"type": "formula","field": "width","expr": "21"},
        {
          "type": "formula",
          "field": "height",
          "expr": "(datum.distinct_Major_Genre + 1) * 21"
        }
      ]
    }
  ],
  "marks": [
    {
      "name": "root",
      "type": "group",
      "from": {"data": "layout"},
      "properties": {
        "update": {
          "width": {"field": "width"},
          "height": {"field": "height"}
        }
      },
      "marks": [
        {
          "name": "marks",
          "type": "symbol",
          "from": {"data": "summary"},
          "properties": {
            "update": {
              "x": {"value": 10.5},
              "y": {"scale": "y","field": "Major_Genre"},
              "size": {"scale": "size","field": "count"},
              "strokeWidth": {"value": 2},
              "stroke": {"value": "#4682b4"}
            }
          }
        }
      ],
      "scales": [
        {
          "name": "y",
          "type": "ordinal",
          "domain": {
            "data": "summary",
            "field": "Major_Genre",
            "sort": true
          },
          "bandSize": 21,
          "round": true,
          "padding": 1,
          "points": true
        },
        {
          "name": "size",
          "type": "linear",
          "domain": {"data": "summary","field": "count"},
          "range": [9,361],
          "round": true,
          "nice": false,
          "zero": false
        }
      ],
      "axes": [
        {
          "type": "y",
          "scale": "y",
          "grid": false,
          "title": "Major_Genre",
          "properties": {
            "labels": {
              "text": {"template": "{{ datum.data | truncate:25}}"}
            }
          }
        }
      ],
[...]

PR #591, opened 2016-06-19
Updating datalib to v1.7.1 for next release

kanitw:
This will ensure that the next release includes fixes in datalib for #584 and #590


arvind:
Thanks, will merge once Travis completes. Though, as I understand it, the caret (`^`) ensures the most recent datalib v1 is pulled in during deployment, so such PRs are not strictly necessary.

Issue #609, opened 2016-07-29
TypeError: this._ is undefined in utc axes with zero=false and no data


Jerrythafast:
Dear Vega people,

I just ran into this issue which only seems to occur with the `utc` scale.

I have a graph spec that contains a `data` section with an empty `values: []` as I intend to use the Streaming Data API to insert the values only after initialising. Fortunately, the MWE is really quite simple:

```json
{
  "data": [{"name": "table", "values": []}],
  "scales": [
    {
      "name": "x",
      "type": "utc",
      "zero": false,
      "range": "width",
      "domain": {"data": "table", "field": "timestamp"}
    }
  ],
  "axes": [{"type": "x", "scale": "x"}]
}
```

Copy-paste this into the Editor, hit Parse, and you get "TypeError: this._ is undefined" somewhere from the depths of D3, downstream from `getTicks()` in Vega. If you change `utc` to any other scale type, such as `time` or `linear` or `log`, the crash does not happen and instead you get a nice blank axis. I would expect `utc` to yield the same result instead of crashing.

The issue does not present itself with `zero` set to `true` (though I don't want zero in my domain). But more interestingly, it seems to be fixed by setting `domainMin` explicitly, even though the DataRef still points to the empty data set.

```json
{
  "data": [{"name": "table", "values": []}],
  "scales": [
    {
      "name": "x",
      "type": "utc",
      "zero": false,
      "range": "width",
      "domainMin": {"data": "table", "field": "timestamp"},
      "domain": {"data": "table", "field": "timestamp"}
    }
  ],
  "axes": [{"type": "x", "scale": "x"}]
}
```

I suppose this is a pretty tiny mistake in the `utc` scale somewhere, but I don't really know where to go looking for it. So I hope you could have a look! :)

Kind regards,
Jerry

PR #635, opened 2016-10-19
Option "zero" should be ignored for utc scales

floribon:
This is already the case for "time" scales and there is no reason for UTC times to behave differently.

### DIFF
--- a/src/scene/Scale.js
+++ b/src/scene/Scale.js
@@ -452,7 +452,7 @@ function domainMinMax(scale, group) {
       domain[z] = def.domainMax;
     }
   }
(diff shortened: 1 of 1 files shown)


floribon:
Note that to reproduce you can open the `lifelines` example and use `utc` instead of `time` for the x scale: http://vega.github.io/vega-editor/?mode=vega&spec=lifelines

Jerrythafast:
This probably also solves the crash I reported in #609

jheer:
Thanks for the PR! Merged into `v2.x` branch.

PR #646, opened 2016-11-06
Update PORTING_GUIDE.md about domain/range_min/max

kanitw:
I got an error that `range_min` and `range_max` is not supported.

I guess the same would be applied for `domain_*` but not sure where to check / if this is intended. 

Please merge if this is the case / feel free to reject this PR if you plan to add these properties back.


jheer:
domainMin/Max *are* supported. rangeMin/Max currently are not.

kanitw:
Sounds like you plan to add them, so I'll close this :) 

Sorry I didn't checked domain_* last night ...

PR #707, opened 2017-01-12
Add convention change for `"interpolate"` to porting guide.

kanitw:


kanitw:
Actually I wonder if we should revert this to dash-separated-strings to be consistent with other properties such as shape names (e.g., `"triangle-up"`).    That said, I don't know if some other properties use camelCase.  (We should be consistent for all of them though.)

kanitw:
Actually I wonder if we should revert this to dash-separated-strings to be consistent with other properties such as shape names (e.g., `"triangle-up"`).    That said, I don't know if some other properties use camelCase though.  (In any case, I think we should be consistent for all of them.)

jheer:
Good catch. The camel case was introduced while moving to the new d3 v4 curve interpolators. However, there's no reason we can't go back to dash-separated names. I'll need to modify [vega-scenegraph's curve table](https://github.com/vega/vega-scenegraph/blob/master/src/path/curves.js#L15).

jheer:
OK, I changed the naming scheme in vega/vega-scenegraph@5ec53e5c48c5b63a58eb7584de984c551fefeafd. A new version of vega-scenegraph has been published, and the change will be included in Vega v3.0.0-beta.15.

PR #766, opened 2017-03-28
Vega-embed example also needs vega-lite to run

larskotthoff:


jheer:
Thanks!

domoritz:
https://vega.github.io/vega-lite/vega-lite.min.js currently links to Vega-Lite 1. Would it be better to use exclusively recommend CDNJS. The advantage is that we will never break users. the disadvantage is that users are not automatically updated to the latest versions.

PR #823, opened 2017-05-23
Delete chart.json

domoritz:
Why is this file in data?


jheer:
Weird. Must be a holdover from earlier development and testing. Thanks for catching that!

PR #842, opened 2017-06-05
Add links to forkable Vega Block.

domoritz:


domoritz:
So it's "block" not "Block"? If so, I should change that in the Vega-Lite docs.

jheer:
I don't know if there is a canonical capitalization here. TBH I just went based on how it looked to me...

domoritz:
Looks like Mike uses "Block" (https://bl.ocks.org/-/about).

screen shot 2017-06-05 at 16 31 25

jheer:
I'm convinced! I'll switch it back in my beta.34 branch.

PR #859, opened 2017-06-15
Fix build for apps that compile vega

domoritz:
`build` may not exist (e.g. when getting vega via github tags)


domoritz:
Unfortunately, I couldn't update the yarn.lock file because of https://github.com/yarnpkg/yarn/issues/3643

PR #897, opened 2017-07-06
Fix gradients on the `Color Schemes` docs

rgrochowicz:
Hello!

Currently, every gradient on the `Color Schemes` page looks the same:
image

Because every svg is reusing the same gradient id, and the id leaks to the window scope, every gradient uses the first `linearGradient` defined in the page (`blues`).

This PR switches each gradient to use a unique id based on the scheme name.


jheer:
Thanks! Testing on Mac OS, both Chrome and Safari show the correct gradients, but in Firefox I'm able to replicate the error. In any case, the unique ids are a welcome improvement!

PR #956, opened 2017-08-03
Update county-unemployment.vg.json

nyurik:
use specific value lookup - makes tooltip more useful, and allows for scale to use the value directly

### DIFF
--- a/docs/examples/county-unemployment.vg.json
+++ b/docs/examples/county-unemployment.vg.json
@@ -32,8 +32,7 @@
     {
       "name": "color",
       "type": "quantize",
-      "domain": {"data":"counties", "field": "rate"},
-      "domainMax": 0.22,
+      "domain": [0, 0.15],
       "range": {"scheme": "blues-9"}
     }
   ],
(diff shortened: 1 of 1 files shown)


nyurik:
@jheer is it better to hardcode domain vs calculating it?

jheer:
Depends on context. In the generic case, no. But I want a specific look in this example so it makes sense to hardcode (plus it saves the need to regenerate thumbnails for the gallery, but that's not the main reason why). Your example hard-coded the domain max anyway, so not a huge difference in terms of its function as an example.

nyurik:
One unfortunate side-effect is that the legend shows incorrect maximum, making it appear as if the maximum unemployment is only 15% in US. While it may look better, it may cause major political problems if published :). I do agree that hardcoding makes for a better visual... I wonder if its possible to calibrate it dynamically, and still achieve similar effect - perhaps with some non-linear scaling?

jheer:
Yes, it would be better for the last legend label to be something like `13.3% or higher`.

PR #991, opened 2017-08-19
Update arc.md

nyurik:


jheer:
Thanks! However, this is not the _intended_ behavior of arc marks and actually arose due to how the browser deals with `NaN` values in the generated path. The latest Vega version includes fixes to vega-scenegraph. If _innerRadius_ is undefined it now defaults to `0`.

PR #1011, opened 2017-09-13
Align view API example with Embed example and improve DX

clemens-tolboom:
I had a hard time using the view API example. Let's just make it copy / paste proof.

I failed in using

```js
d3.json('https://vega.github.io/vega/examples/bar-chart.vg.json', render);
```

> ReferenceError: d3 is not defined

Not sure how that happened as I'm not loading core only.


clemens-tolboom:
Thanks for the updated version. Awesome!

PR #1050, opened 2017-10-24
Common syntax style between VL and VG

domoritz:
Before:

screen shot 2017-10-23 at 18 09 10

After:

screen shot 2017-10-23 at 18 08 34


domoritz:
Looks like this was a bad change. The difference is because rogue just updated their JSON parser in a minor version and now doesn't differentiate keys and values anymore. https://github.com/jneen/rouge/issues/675. I will send a PR to revert this.

PR #1057, opened 2017-10-27
Twitter cards

domoritz:
I only added image URLs to two pages. Is there a script to generate the examples or do I need to manually (or programmatically) edit all pages?

### DIFF
--- a/docs/Gemfile
+++ b/docs/Gemfile
@@ -1,7 +1,2 @@
 source 'https://rubygems.org'
-
-require 'json'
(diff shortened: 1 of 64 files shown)


jheer:
There is no example generation script. Edits to all example pages are needed.

domoritz:
Okay, updated all of them.

domoritz:
The build errors seem unrelated to my changes. `An error occurred while generating the build script.`

PR #1062, opened 2017-10-31
Add Diverging Stacked Bar Chart to Bar Chart Examples

haldenl:
To show usage of a zero-centerline diverging stacked bar chart.


domoritz:
Since this example is generated from Vega-Lite, I'm not sure how much value this example adds. I don't think we want to have much redundancy between Vega and Vega-Lite examples unless the Vega example uses some additional features like the crossfilter example.

haldenl:
@kanitw suggested I run this by you, but that makes sense! I'll close this for now.

kanitw:
What I suggest is actually do the transformation in Vega (which cannot be done in Vega-Lite yet) so it makes sense to add the example to Vega.

domoritz:
I think this is a very subtle difference that is not going to be obvious to anyone unless it's a different example. Would it be sufficient to add an example to the docs?

kanitw:
Are there really full examples in Vega docs?

haldenl:
Ah yes, I should have remembered to bring this up. 

In particular, the example uses a window (not supported in vega-lite at the moment) and stack transform (with calculations littered between) to interpolate bar locations from raw question percentages. The vega-lite example uses x, x2 (percentage_start, percentage_end) to manually specify the bar locations.

I've removed the percentage_start, percentage_end, and value fields in the in-line data to make this more clear.

domoritz:
I still think the example should also look different to avoid confusion.

haldenl:
If we still think this is valuable, I can find a different dataset to map this to.

arvind:
The other point I'd note is that we want Vega examples to be written in a canonical fashion, rather than including conventions Vega-Lite uses for its compilation process (e.g., the `toNumber` formula, layout calculations, etc.).

kanitw:
@haldenl  Since it will be a lot of work to make this PR work, perhaps let's wait until VL supports this properly and add a proper example to VL gallery.

haldenl:
@kanitw sounds good.

PR #1121, opened 2018-02-01
Document runAsync. Fixes #1119

### LINKED ISSUE #1119: Document view.runAsync()
I'm not sure whether this is part of the public API. It would be useful to have access to this method to wait for the dataflow to run before adding an element to a page. See https://talk.observablehq.com/t/bug-report-layout-gets-pushed-around/111/3

cc @mbostock

domoritz:


jheer:
Thank you!

PR #1123, opened 2018-02-02
Add note that signal definitions are not shown in the example.

ericsoco:
Took me a bit to understand that the `shift` and `clicked` values are actually signals defined elsewhere in the spec, not shown in the example. Hopefully this clarification helps others too!


jheer:
Thank you! Very helpful.

PR #1138, opened 2018-02-11
Updated types doc

nyurik:
See https://github.com/vega/vega/issues/1139 before merging.


jheer:
Thanks, but these are not correct, nor complete. So closing this PR in lieu of future updates.

Issue #990, opened 2017-08-18
[enh] Support for group attribute access in expressions


nyurik:
Please allow expression access to the group's attributes. The most useful would be to get the width/height of the immediate and parent groups.

```js
// This gets the width of the current (parent) container
// either the enclosing group, or the plot area (not the same as the graph's "width")
getAttr('width')
// get the width of grand-parent container,
// or if current group is first level, gets the width of the whole graph's plot area
getAttr('width', -1)
// get the width of the plot area
getAttr('width', null)
// get the width of the named group
getAttr('width', 'groupname')
```
Ideally, all other attributes of a group, including x,y positioning and even fill color should be available via this method.

jheer:
Within an encoding signal expression, parent groups are already accessible via `item.mark.group`.

Alternatively, group fields can be accessed within an encoder using the `field` directive. For more, see https://vega.github.io/vega/docs/marks/#fieldref

nyurik:
@jheer I wasn't talking about the parent's **data**, but rather parent's **attributes** like width/height, and possibly more complex (reflection-style) things strikeColor.  In a mark, I often want to get the width/height of the current group, or the parent group.  I know that x,y are relative to group's x,y.

jheer:
So was I. The `parent` identifier accesses the parent group's datum. What I described above accesses the group scenegraph item, which houses the visual properties.

nyurik:
Thanks @jheer, I wasn't aware of the "item" keyword. Is it documented anywhere? I only found "datum" and "parent", but "item" is not mentioned. What properties are expected on it?  (I see them in the debugger, but I doubt most users will find them). BTW, my only way to figure things out is to create a "valid but crashing expression" - e.g.  `"parent.blah.blah"` -- this gets processed by initial vega parser, but throws an exception during execution. Catching that expression in debugger allows me to examine the context of that expression. Sadly, this is the only way I found to debug anything. I hope the debugger will get migrated to Vega 3, and we can easily integrate it into end user env. Thanks!

PR #1144, opened 2018-02-12
Update signals.md

nyurik:


jheer:
Thanks! I removed the subsection describing the `item` variable for a couple of reasons:
- I do not want to encourage direct use of item (a variable intended primarily for internal use) in expressions, particularly for chained references to group properties. There is an existing JSON syntax for encoders to reference such properties.
- The example provided was not useful nor indicative of best practices.
- Detailed information about scenegraph elements and structure probably belongs in a more appropriate, dedicated place.

nyurik:
@jheer wrt `item` use:
* I was following your advice from https://github.com/vega/vega/issues/990#issuecomment-323404482 : to access parent group's width/height, e.g. `height: {signal: "item.mark.group.height"}` and `yc: {signal: "item.mark.group.height/2 + 2"}`. See [example](https://gist.github.com/nyurik/2729e86307fbc600d15c736c37392514#file-sankey-vega-graph-json-L228).  I would love to be able to access these fields via some more standardized way from an expression, but sadly this is the only path I found for a reliable display of a button (group of a group of a text mark).

* I agree that `item` use should have a warning to discourage its use (e.g. access to internals, structure likely to change), but not documenting it would raise a red flag for any security review of Vega - viewed from security perspective, expressions is the very first thing people ask about. They are concerned with XSS and privacy issues, especially due to `eval()` use. Having a backdoor into Vega internals, especially when it's not fully documented is a big concern, as it leads to questions of "what else is hidden in there that's not documented".

PR #1168, opened 2018-02-22
Reexport typings from vega-typings

pelotom:


domoritz:
Should we add this to vega-parser as well? We could close https://github.com/vega/vega-parser/issues/63.

pelotom:
Yep, and I would imagine `vega-lite` should depend on it as well.

domoritz:
> Yep

Specifically, we only want the typings in the spec directory there. 

> I would imagine vega-lite should depend on it as well.

On it.

jheer:
Included in Vega 3.2 onwards.

PR #1188, opened 2018-03-13
Make canvas a peer dependency

matt-d-rat:
__Description:__

This PR changes the npm dependency of `canvas` from an optional dependency to a peer dependency.

The motivation behind this change is to mitigate installations from failing (particularly in build environments / CI environments) when the underlying dependencies that `canvas` post-install script requires to compile are not available.

While optional dependencies are not supposed to fail the installation, we have seen several instances within our own builds, as well as these [other](https://github.com/npm/npm/issues/14185#issuecomment-264879789) documented cases, where this is not the case for this particular dependency.

__What's Changed?__

- Updated `package.json` to include `canvas` as an **opt-in** peer-dependency
- Updated the `README.md` with a section on Peer Dependencies and Optional Dependencies in order to communicate the change and what uses need to do if they wish to not use `canvas-prebuilt`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,14 @@ folder. The last command launches a local webserver. Now, you can open
 [`http://127.0.0.1:4000/vega/`](http://127.0.0.1:4000/vega/) to see the
 website.
 
+### Peer Dependencies and Optional Dependencies
+
+If you require HTML5 Canvas support and are targetting an environment which does not natively support HTML5 Canvas (eg: [Node.js](https://nodejs.org/en/)), Vega provides support for this feature via a third-party node dependency called [`canvas`](https://www.npmjs.com/package/canvas).
+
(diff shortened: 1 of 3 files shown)


domoritz:
You can use vega-lib, which does not depend on Canvas: https://github.com/vega/vega-lib.

matt-d-rat:
@domoritz Ok. I will raise two PRs against third-party dependencies which currently depend on Vega.

P.S Closing immediately without taking the time to thank for the contribution (even if it is not the right path) discourages contributions and doesn't act in accordance with your [Code of Conduct](https://github.com/vega/vega/blob/master/CODE_OF_CONDUCT.md).

domoritz:
> @domoritz Ok. I will raise two PRs against third-party dependencies which currently depend on Vega.

@matt-d-rat Excellent, I hope everything works well. 

> P.S Closing immediately without taking the time to thank people for their contributions (even if it is not the right path) discourages future contributions and doesn't act in accordance with your Code of Conduct.

Thank you for immediately bringing this up. I'm sorry if I sounded harsh and let me say it now: thank you for your pull request and your contribution to the project. Please do continue to send pull requests! A thriving community is very important to me and I've been giving shout-outs to first-time contributors on Twitter (e.g. https://twitter.com/vega_vis/status/966023876915355648).

matt-d-rat:
@domoritz Not a problem, thank you for addressing my feedback. Love this project and want to see it thrive also

PR #1211, opened 2018-03-29
Document "name" property of input bindings.

arvind:


kenklin:
@arvind Thank you for the doc update.  

As I am updating my spec's input bindings with "name", I also see that elsewhere in my spec that "title" is used for a similar purpose for axes (https://vega.github.io/vega/docs/axes/).  So I wonder if "title" would be a better property name vs "name" for input bindings?  I haven't checked if there are other defacto Vega naming conventions.

jheer:
Thanks!

PR #1224, opened 2018-04-05
Update zoomable-scatter-plot.vg.json

### LINKED ISSUE #1223: Background fill for axes
I have line chart that allows pan-zoom functionality, bit like https://vega.github.io/editor/#/examples/vega/zoomable-scatter-plot. I note that as you pan and zoom, the chart overlays the axis and apparently there is no way to set the background to the axis so that as chart is panned-zoomed the axis is still in clear view...
chrome_2018-04-05_12-18-36

domoritz:
Fixes #1223


jheer:
Thanks!

PR #1233, opened 2018-04-08
[docs] add "name" description to the "bind"

nyurik:


jheer:
Duplicates #1211, which will be merged once new schema validation support for bind name has been released.

PR #1234, opened 2018-04-08
Enable resolution scale factor in vg2png

### LINKED ISSUE #528: vg2png with higher resolution
Is there a way to generate images with a higher resolution? It looks like vega-scenegraph has support for retina. We would like to use high res images in the VL gallery and on the home page.

RobinL:
This should fix #528.

I have tested on mac and using the following Docker image:

```
FROM node:latest

RUN npm config set unsafe-perm true

RUN npm install -g vega
RUN npm install -g vega-lite

COPY vg2png /tmp/
RUN cp /tmp/vg2png /usr/local/bin/vg2png && rm -f /tmp/vg2png

ENTRYPOINT ["vl2png"]
```

```
docker build . -t vg2_1
docker run --rm -v `pwd`:/tmp vg2_1 tmp/chart.json -s4 > chart.png
```

For future reference:  I had some issues with misalignment of the text in the output with the axes.  This was due to using `libpango1.0-dev` rather than `libpango1.0-0`.  

To use the changes made in this PR in `vl2png` you would use:

`vl2png chart.json -s 2 chart.png`

I considered updating vl2png so that it could be called like `vl2png -s2 chart.json chart.png` but I don't have much experience with bash, and parsing arguments seemed to [add a lot of complexity](https://stackoverflow.com/questions/192249/how-do-i-parse-command-line-arguments-in-bash).

If this solution makes sense, we could add a similar thing [here](https://github.com/altair-viz/altair/blob/92e9de8ddaae05a57137ba2491a45a110622eb01/altair/utils/headless.py#L73) in altair.

Do let me know if this PR is naive/way off base - it's my first contribution so entirely possible I've missed something obvious.

### DIFF
--- a/bin/vg2png
+++ b/bin/vg2png
@@ -19,8 +19,13 @@ var path = require('path'),
 var args = require('yargs')
   .usage(helpText)
   .demand(0)
-  .string('b').alias('b', 'base')
-  .describe('b', 'Base directory for data loading.')
+  .string('b')
+    .alias('b', 'base')
+    .describe('b', 'Base directory for data loading.')
+  .number('s')
+    .alias('s', 'scale')
+    .default('s', 1)
+    .describe('s', 'Output resolution scale factor.')
   .help()
   .version()
   .argv;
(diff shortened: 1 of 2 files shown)


jheer:
Thank you @RobinL! This looks great. Will merge when preparing the next release -- that way the online documentation will be consistent with the published package.

jheer:
Merged! Thanks again!

PR #1237, opened 2018-04-09
Update group.md

nyurik:


nyurik:
@jheer sorry, this is not very clear.  When I was trying to attach a signal to the mousemove over a group mark, it wasn't working until I specified the fill for it.  This behavior is not intuitive, and needs to be documented.  What is the best way to do that?  Also, are you saying that this behavior is the same for other marks with fill?  E.g. if I have a circle or a rect, events would not trigger unless they have a fill property?

jheer:
Yes, this behavior is the same with all marks. This is the standard behavior in web browsers outside of Vega as well (e.g., for event handling in SVG).

PR #1242, opened 2018-04-11
[WIP] nest transform example

nyurik:
This example is broken - possibly due [to this Vega bug](https://github.com/vega/vega/issues/1243) -- the `generate` signal does not update the graph. Setting the initial value to false and clicking the checkbox only draws the lines, not the symbols. 

image


jheer:
Thanks! I will merge this example once the fix for #1243 has also been merged in.

jheer:
Thanks!

PR #1249, opened 2018-04-20
Add test flag to vg2png and vg2svg for deterministic testing purpose

kanitw:
Vega-Lite uses vg2svg for generating SVG output of the examples.  However, without a deterministic random seed, the examples with CIs get changed everytime.   Thus, I want to add this flag. 

(I'm not super married to the `-t` and `--test` flag, so please feel free to replace it if you have better suggestions.)


domoritz:
This is great. Why not let users provide a seed?

kanitw:
That may be useful, but also seems to require more input from users?

domoritz:
`vg2png -t` (vg2png --test) vs `vg2png -s 42` (vg2png --seed=42) isn't much of a difference and it is immediately clear what is happening.

jheer:
I agree with @domoritz and would prefer a more general `seed` argument.

Note also that this PR needs to be updated now that `vg2png` supports a scale parameter (`-s`, `--scale`) for adjusting the resolution of the generated PNG image. Perhaps simply use `--seed` with no alias?

kanitw:
Rebased and fixed :)

jheer:
FYI, added a subsequent commit to fix bugs, add error checking, and to lift lcg to a shared import.

Issue #1278, opened 2018-05-18
Label flush incorrect for reversed scales


domoritz:
```json
{
  "$schema": "https://vega.github.io/schema/vega/v3.json",
  "width": 200,
  "scales": [
    {
      "name": "xscale",
      "domain": [0, 1000000],
      "nice": true,
      "range": "width",
      "reverse": true
    }
  ],
  "axes": [
    {
      "orient": "bottom",
      "scale": "xscale",
      "labelFlush": true,
      "labelOverlap": true
    }
  ]
}
```

![image](https://user-images.githubusercontent.com/589034/40260928-6ba14598-5ab2-11e8-8681-fceb187f9724.png)

jheer:
Thanks! Will fix.

jheer:
Fixed in vega-parser 3.1.0: https://github.com/vega/vega-parser/releases/tag/v3.1.0

PR #1296, opened 2018-05-25
v4.0.0-rc.2

jheer:
Changes from [v4.0.0-rc.1](https://github.com/vega/vega/pull/1269):
- Add `timer` event type, can be used to drive frame-by-frame animation.
- Add `merge` expression function to merge objects into a new output object.
- Add new example test specifications.
- Add runtime description `$ref` object support for event update sources.
- Add cyclical color schemes `rainbow` and `sinebow`.
- Fix container element check in default tooltip handler, preventing errors for headless displays.
- Fix value equality check in bind input update to suppress unnecessary updates.
- Fix flush axis label alignment. (vega/vega#1278)
- Fix dataflow bug where operator evaluation might skip upon signal update.
- Update dependencies.
- Update documentation.


domoritz:
🎉

PR #1314, opened 2018-06-09
Add year format to x axis

jwoLondon:
The presidents' timeline example does not render year labels on the x axis correctly in Chrome (but does with Safari). Explicitly setting the format in the axis definition appears to solve the problem across browsers.

(My editor auto-formatted the .json whitespace; the only substantial change is in new line 76)

### DIFF
--- a/docs/examples/timelines.vg.json
+++ b/docs/examples/timelines.vg.json
@@ -47,12 +47,12 @@
     },
     {
       "name": "events",
-      "format": {"type":"json", "parse":{"when":"date"}},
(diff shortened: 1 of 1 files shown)


jheer:
Thanks @jwoLondon! I appreciate the PR. However, to isolate the change (separate from the mass reformatting of the file), I'm going to close the PR and add the change manually.

This same issue, which was reported earlier, seems to have been introduced in the latest Chrome version. We'll have to track down if the issue lies in Vega or in d3 (e.g., d3-time-format).

PR #1336, opened 2018-06-21
Use official Slack join link

arvind:
Rather than third party communityinviter.com as we did previously.


domoritz:
Can we create a short link?

kanitw:
Let me try to update the bit.ly link to the new url.

PR #1353, opened 2018-07-09
Implement documentation version tagging

nyurik:
This allows all new features in documentation
to be tagged with the minimum required version.

Includes one example usage for the event timer.
image

To use:   `{% include tag ver="4.0" %}`

Eventually the template should be expanded to support
deprecations, and possibly other warnings.

### DIFF
--- a/docs/_includes/tag
+++ b/docs/_includes/tag
@@ -1,0 +1,1 @@
+<span class="tag version" title="Requires version {{ include.ver }} or later">≥&nbsp;{{ include.ver }}</span>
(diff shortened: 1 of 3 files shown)


domoritz:
Can you send a PR to add this to Vega-Lite as well? It would also be nice if we tracked this in the schema instead of in the documentation.

nyurik:
@domoritz i agree it would be great to structure tracking of this, but i doubt it would be very easy - in some cases schema might not change with the introduction of a new feature, but it should still be documented.

jheer:
Thanks @nyurik, this is a great addition!

domoritz:
In Vega-Lite, we generate the docs from the schema (which in turn is generated from the code). We will need to decide on a special jsdoc comment and then we can automatically generate the docs with the right version tags.

PR #1358, opened 2018-07-12
add missing 4.1 tags

nyurik:


jheer:
Thanks! Before merging, we should probably have a procedure for determining what should be documented in this way. Do we assume 4.0.0 as the base version and annotate newer things? Or version 3.0.0? Or...? (For example, should we also annotate the _many_ axis and legend definition properties made available in 4.0.0?)

One idea might be to (eventually) annotate *everything* added post-3.0.0 with a version but then modify the tag template as needed to show version information from a specific "base" version?

nyurik:
@jheer I have been going through the release notes, annotating everything since 3.0, but there is some work left to do.  We  cannot assume 4.0 is the base, because all pre-4 users still use this site for the documentation. For example, Many Kibana users still don't have access to `flatten` transform from version 3.2, because they haven't upgraded their Elasticsearch, and in some cases such upgrade can be very costly and time consuming, yet they still need to be able to view Vega documentation.

jheer:
@nyurik Sounds good to me! Just wanted to make sure we're all on the same page first.

PR #1361, opened 2018-07-12
Added missing 'property' parameter for topojson

nyurik:
Seems like type=topojson supports the `property` parameter, but doesn't document it.


jheer:
Thanks! The TopoJSON reader first applies the JSON reader (and so supports all format options also supported for JSON), then performs TopoJSON-specific steps: https://github.com/vega/vega-loader/blob/master/src/formats/topojson.js#L7

nyurik:
yep, that's what i discovered in the debugger :) thx for merging

PR #1411, opened 2018-08-28
Docs: correct sum output

danmarshall:


danmarshall:
🤔 perhaps I misunderstoof the `mean`. Closing.

danmarshall:
OIC - there are 2 rows with the key of `2` 👍

PR #1511, opened 2018-12-29
Add Vega Typings

domoritz:
Fixes vega/vega-typings#58


.editorconfig
```diff
@@ -0,0 +1,11 @@
+# top-most EditorConfig file
```
jheer: As this adds a new top-level config file, can you say more about what it does and what (if anything, such as a plug-in) is needed for editors to adhere to it?

package.json
```diff
@@ -24,8 +24,8 @@
     "rollup-plugin-json": "^3.1.0",
     "rollup-plugin-node-resolve": "^4.0.0",
     "tape": "^4.9.1",
-    "tv4": "^1.3.0",
     "terser": "^3.11.0",
+    "tv4": "^1.3.0",
```
jheer: This rearrangement makes sense but also caused a merge conflict. These kinds of orthogonal changes to the main intent of the PR are best included as a separate PR altogether.

packages/vega-typings/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "repository": "vega/vega",
+  "author": "Dominik Moritz <domoritz@gmail.com>",
+  "keywords": [
+    "vega",
+    "typescript",
+    "typings"
+  ],
+  "license": "BSD-3-Clause",
+  "devDependencies": {
+    "dtslint": "^0.4.2",
```
jheer: Although only used by vega-typings at the moment, I'm fine having these devDependencies moved to the top-level package.json file for consistency across packages and easier updates to the development environment.

jheer:
I merged the earlier dependency update PR, but now there is a conflict here. @domoritz, can you update to resolve the conflict first?

.editorconfig
domoritz: This sets consistent whitespace characters and line endings for various editors. We use it in Vega-Lite but I'm happy to remove it from this PR. 

Some editors (e.g. github, vs) don't need any additional plugins. There are plugins for various editors at https://editorconfig.org/#download.

package.json
domoritz: This might have been yarn. I'll rebase.

.editorconfig
jheer: I'm all for having a useful config and this looks totally reasonable. I just wanted to understand it before merging. I'm also open to using prettier in a consistent way across the Vega modules (though obviously that would be the subject of a different PR).

domoritz:
Rebased and tests are passing.

PR #1512, opened 2018-12-29
Add script to build typings test cases

domoritz:


docs/examples/crossfilter-flights.vg.json
```diff
@@ -301,7 +301,7 @@
               "text": {"value": "Arrival Delay (min)"},
               "baseline": {"value": "bottom"},
               "fontSize": {"value": 14},
-              "fontWeight": {"value": "500"},
+              "fontWeight": {"value": 500},
```
domoritz: I'm changing this because in typescript we don't allow numeric font weights as strings.

PR #1532, opened 2019-01-23
Provide a rolled-up es modules as part of vega's build

eterna2:
## Summary
- Updated `vega` build steps to include rolling up es modules (with `rollup-esm.js`).
- Updated `package.json`'s `module` to points to `vega-core.mjs` instead of `index`.

## Background
I having been maintaining [`vega-element`](https://github.com/PolymerVis/vega-element) which is a webcomponent for `vega` lib for a couple of years.

In the past, I was able to get by with `vega` umd import, but this is not ideal as the dependency is not properly captured in the webcomponent. 

## Issues
- current `vega` lib does not fit the strict es module spec (imports a json file):
  - users are still unable to import `vega` natively in browsers which support es modules
  - [`polymer`](https://www.polymer-project.org/) tool chain is entirely unusable (as they have yet to support user-specified rollup plugins)
    - no one can use my webcomponent unless I pre-pack with with a compiled `vega` or they customized their tool-chain

- unable to rollup `vega` properly from `vega` `npm packages`:
  - I have not figure out why, rollup is unable to properly pack `vega` if I am doing it from my `node_modules`
  - based on my investigation on the rolled-up sources, roll-up seems to be unable to reconcile duplicates imports in multiple modules, resulting in some of the "singleton" objects existing as multiple objects 
  - (e.g. `transform.proxy` used in `signalTest` was wrongly renamed as `transform$6.proxy` while the rest of code like `extend(transform, ....)` still refers to `transform` object.


jheer:
Thanks! This looks reasonable to me, though ideally I'd also like to have a clearer picture of why rollup fails as is. Do you have a simple example project you can share that replicates the issue? I'd be happy to take a closer look, and then merge this PR once I verify that this is indeed the best solution.

eterna2:
@jheer 
Sorry for the late response. I was trying to reproduce it.

I couldn't reproduce the issue in a separate clean repo (the rollup output works just fine), but I think I have narrowed down the issue  - something might have gone wrong when I upgraded vega, vega-lib, vega-lite versions in my mono-repo.

---
But regardless, it work be nice if this PR can be approved so that my webcomponent can be used as-is without me packing a version-locked vega lib inside my package or getting downstream users to customize their build chain.

```html
<script type="module" src="./node_modules/vega-element/vega-element.mjs"></script>

<vega-element hover tooltip spec="spec.json"></vega-element>
```

jheer:
Rather than duplicate most of the code, I've updated the `rollup` script to support an option to generate an ES module build. I also updated `package.json` to reference that build as the `module` entry point. The changes were added to master in d3b7a5a and will be released in v5.0.0.

PR #1541, opened 2019-01-29
Split Data type into separate dataset definitions in vega-typings

jonathanzong:
Hello! In https://github.com/vega/lyra/pull/416#issuecomment-443005618 we found that when we use the `Data` type definition, Typescript complains when we reference properties specific to a subtype, like `source` or `values` or `url`. This change splits `Data` into a union of separate type definitions so that we can use type guards to narrow the type later on.

cc @arvind


jheer:
Thanks! I've requested that @domoritz and/or @arvind take a look at this PR, given they have more TypeScript experience!

domoritz:
Actually, looks like a test fluke. I restarted https://travis-ci.org/vega/vega/jobs/486166271.

PR #1544, opened 2019-02-05
Update scale handling.

### LINKED ISSUE #995: Ordinal scale can interpolate scheme, but not range
Vega's ordinal scales accept `scheme` and seems to be able to interpolate sequential color scheme

For example, using the following scale works (full spec)
```
{ 
  "name": "color",
  "type": "ordinal",
  "domain": {"data": "source_0","field": "Cylinders","sort": true},
  "range": {"scheme": "blues"}
}
```

image

Full spec: https://pastebin.com/KzifPxmJ.

However, when we specify an array `"range": ["pink", "red"]`, there is no way to force it to interpolate like it does with `scheme` (to the best of my knowledge).  
Of course, it makes sense not to interpolate the color by default, since this is D3 ordinal scale's default behavior.  However, since Vega's ordinal scales, unlike D3's, allow interpolation with `range.scheme`, shouldn't there be a way to enable interpolation when `range` is an array of color?  For example, maybe we could support `interpolate` property for ordinal scales?

This is not really a pressing issue, but it seems like we're a bit inconsistent here (unless I misunderstand something) so it's worth reporting. :)

### LINKED ISSUE #1479: Sequential scale does not show any color when the domain has only one value.
Example: [link](https://vega.github.io/editor/#/url/vega-lite/N4KABGBEAkDODGALApgWwIaQFxUQFzwAdYsB6UgN2QHN0A6agSz0QFcAjOxge1IRQyUa6ALQAbZskoBmOgCtY3AHaQANOCgATZAgBOjQnh4qckAIJhYjVITHIw7dLrBIneMAHdmiMGnbJNbU0wTXQ8ejUNSFDw7DBQCAhICnQxVh04gG0NRPjITFMzNUh2OIAmAA4AX1U8gqgAIWLSnABWVpq6uMgAYWa4gBZpGpzE4HzugBF+nABOAEZO8frIAFEZsArF2uXugDEN1uH1XIhd0wBxDfnZpYnTAAkNioB2O5WASUOyqtGAXQ0IySeF06CUsAAZtxdKgsglEpB4Kl4KwxGFkN15pFcvlYN14JBAX8TlAMLoANbdRy6bFQZBKeDcTSMJTUOLwpIAD3ZkAhjGQYk03UwtUgeAAnoQMaZocylKlIECEeKeXyBULTKVRRKpd0AI6sMFGcJGKiKklJRliaGq-mC-FqKA66VQA1G5hhRhm34QX6-IA)

Changing `1` in the calculate transform to `random()` will show the color. 

image

This looks like a broken plot. 

I think we should at least use the max color in the color scheme.

(I'm filing it here as we're moving toward a mono-repo.)

### LINKED ISSUE #1536: Support D3 Scales 2.2 scale types
https://github.com/d3/d3/releases/tag/v5.8.0/https://github.com/d3/d3-scale/pull/156 has new scale types that would be great for Vega-Lite.

jheer:
Changes:

**vega**

- Update test specifications and rendered scenegraphs.

**vega-encode**

- _**Breaking**_: Remove `bin-linear` scale support. (Use new `bins` parameter instead.)
- _**Breaking**_: Remove `utcInterval` method. (Use updated `timeInterval` method instead.)
- Add automatic scale selection for linear, log, pow, sqrt, symlog types. If a color range is specified, sequential or diverging scale implementations will be used depending on the domain length.
- Add interpolated color array input to `scheme` property. (Fix #995)
- Add color interpolate support for all scheme-compatible scales.
- Add support for `bins` property to guide tick/legend generation.

**vega-parser**

- Add `panSymlog` and `zoomSymlog` expression functions.
- Update legend and scale parsers to use scale type predicates.
- Fix `scaleGradient` function to handle zero span domains. (Fix #1479)

**vega-scale**

- _**Breaking**_: Remove `bin-linear` scale type.
- Add `symlog` scale type.
- Add transformed sequential and diversing scale implementations.
- Add scale type constants and predicate methods.
- Upgrade to d3-scale 2.2+. (Fix #1536)

**vega-schema**

- Update schema to reflect additions and removals.
- Refactor to use internal JSON schema API.

**vega-util**

- Add `panSymlog` and `zoomSymlog` methods.

**vega-view-transforms**

- Update `Overlap` transform to pre-filter all labels with null bounds.


packages/vega-schema/src/on-events.js
```diff
@@ -1,72 +1,49 @@
...
-        },
-        {"$ref": "#/defs/stream"}
-      ]
-    },
+import {
+  array, allOf, def, oneOf, object, ref,
+  anyType, booleanType, stringType, signalRef
+} from './util';
+
+// ???
```
domoritz: ???

jheer: '???' is an extremely detailed and informative comment to myself to revisit later to see if I want to consolidate reference values as a set of exports rather than recreate them across various schema generation files.

packages/vega-schema/src/axis.js
```diff
@@ -3,128 +3,126 @@ import {
...
+const guideEncodeRef = def('guideEncode');
 
-        // DOMAIN CONFIG
-        "domain": {"type": "boolean"},
-        "domainColor": colorValue,
-        "domainOpacity": numberValue,
-        "domainWidth": numberValue,
+const axis = object({
+  _orient_: enums(axisOrientEnum),
+  _scale_: stringType,
```
mattijn: From what I understand here is that vega 5 will not support signals for the `scale` parameter within the axes properties? And based on the discussion in https://github.com/vega/vega/pull/1544 I understand there was an idea for a `transform` parameter that accept a signal within a scale, but I cannot find this in the code. Still hoping to be able to swap between log and linear scales through signals one day.

jheer: Scales are still string-typed (no signals allowed), and after further discussion, scale types still determine the transform. We do have a potential path forward for supporting dynamic scale types (with some natural limitations), but that has not yet been implemented. The related feature requests remain open in the issue tracker.

jheer: (Part of the reason for this is that there was some urgency to get Vega 5.0.0 released sooner to support a Vega-Lite 3 release.)

mattijn: Thanks for clarifying. It's unfortunate but understandble

PR #1551, opened 2019-02-07
Update typing tests

domoritz:
I've created a [script to update the tests](https://github.com/vega/vega/blob/master/packages/vega-typings/build-tests.sh) for the typings with new examples. You can run it to make it easier to see what changes may require typings changes. I don't know what would be the best way to automatically call this script.


domoritz:
cc @arvind

domoritz:
I updated the typings so that the tests pass. This PR also includes #1565.

domoritz:
Rebased and updated typings for dynamic data and label separation property.

PR #1553, opened 2019-02-08
Add async evaluation support.

jheer:
Changes:

**vega**
- Add test specifications and scenes for dynamic data loading.

**vega-encode**
- Update unit tests to work with async evaluation.

**vega-dataflow**
- _**Breaking**_: Decouple dataflow `request` from pending load tracker.
- Add dataflow `preload` method for initial data loading.
- Add async evaluation support. Transforms can now return a Promise that resolves to a Pulse.
- Update data loading and dataflow methods to use async / await.
- Add async evaluation tests.

**vega-parser**
- _**Breaking**_: Update data parser to include `load` transform for dynamic data directly within generated transform pipelines.

**vega-runtime**
- _**Breaking**_: Remove context support for dynamic loading. The dataflow now supports this directly.
- Use dataflow `preload` for initial dataset loading.

**vega-schema**
- Update schema to support signal values in dataset definitions.

**vega-transforms**
- Update `load` transform to use async evaluation.
- Update `load` transform to support `values` parameter in addition to `url`.
- Update unit tests to work with async evaluation.

**vega-view**
- Update View API to override `Dataflow.runAsync` instead of `run`.
- Update headless rendering methods to use async / await.


domoritz:
I'm unreasonably excited about this. :tada:

PR #1565, opened 2019-02-14
Build tests before testing typings.

domoritz:
I think this could be a good way to ensure that the typings stay up to date.

### DIFF
--- a/packages/vega-typings/package.json
+++ b/packages/vega-typings/package.json
@@ -14,6 +14,7 @@
   "scripts": {
     "format": "prettier $([ \"$CI\" = true ] && echo --list-different || echo --write) './**/*.{ts,tsx,js,json,css}'",
     "build-tests": "./build-tests.sh",
+    "pretest": "npm run build-tests",
     "test": "dtslint ."
   },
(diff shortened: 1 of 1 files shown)


domoritz:
Related: https://github.com/vega/vega/pull/1551

jheer:
Why are builds failing for this and #1551? I don't really understand what is going on here as I have not been involved with the vega-typings package, and I don't want to merge anything until I have a better sense of what is failing here. Explanations welcomed! :)

domoritz:
It's failing because the typings are not up to date with the examples.

This PR only enables additional tests to show that the typings are not up to date and provides a way to test typings against the examples automatically. What the tests do is generate a JS version of the examples and test whether they are valid according to the typings we have for Vega specs. The tests will pass again once the typings have been updated.

jheer:
I see. So to resolve we need to update the typings to reflect the changes that have been made to the vega schema?

domoritz:
Yes. At the very least, we need to update the typings to reflect the changes in the examples.

domoritz:
Part of #1551

PR #1566, opened 2019-02-15
Convert Vega Util to Typescript

domoritz:


jheer:
Why?

domoritz:
> Why?

I wanted to see what it takes to move over a Vega package to Typescript. Since we already had typings for the utils, this seemed the most straight-forward package. Making the switch amounted to renaming the files and applying the types to the code directly. I left the tests and build pipeline alone as much as possible. Along the way, I found a few bugs in the typings that were easy to spot when I checked the typings against the implementation.

From now on the typings will make it easier to check correctness, support autocomplete, and provide documentation for newcomers.

jheer:
Thanks! We'll discuss offline and make a decision about our strategy here. As a result, if we incorporate this it likely won't be until after the 5.0 release.

PR #1570, opened 2019-02-16
Improve utils typings

domoritz:
I think these are the most important typing fixes from https://github.com/vega/vega/pull/1566.

Small aside: It's easier to ensure that typings are correct when they are linked to the actual implementation and so I would still prefer to move to typescript.


jheer:
Thanks!

PR #1572, opened 2019-02-16
Standardize tape tests to use t as the test variable.

domoritz:
Tape tests usually use the variable name `t` for the test variable. This PR standardizes on that practice.


jheer:
I don't see the point of doing this. Vega uses a standardized convention and it is arguably more readable.

domoritz:
I was looking at https://github.com/substack/tape and they use just `t`. The main reason for me to send this PR was to enable automatic conversion with https://github.com/skovhus/jest-codemods, which could enable us to run tests in parallel and thus significantly reduce testing times.

domoritz:
By upgrading the testing infrastructure, we could get a few advantages

* Support running tests in parallel and therefore reducing delays.
* Use the same framework as Vega-Lite and thereby making it easier for developers to develop both.
  * We have experience with it in Vega-Lite and know how to make things work for Typescript if we wish to switch. 	
* Jest has some nice features for automatically running tests for changed code.
* Use puppeteer like we do in the Vega-Lite runtime tests to not rely on MacOS for testing to broaden access for developers.

jheer:
Thanks for the explanation, will merge!

PR #1573, opened 2019-02-16
Automatically use data from vega datasets and update datasets

domoritz:


package.json
```diff
@@ -10,7 +10,8 @@
     "license": "lerna exec -- cp ../../LICENSE .",
     "prepublishOnly": "yarn license",
     "serve": "http-server packages/vega/ -c-1 -p8080 -o",
-    "test": "lerna run test"
+    "test": "lerna run test",
+    "postinstall": "npm run data"
```
jheer: Why not `yarn data`? At this point I don't think I call `npm run` anywhere in Vega.

domoritz: I think for the normal packages we should not force people to use yarn but since this package is only a meta package it doesn't matter. I changed it.

PR #1578, opened 2019-02-19
Add Tableau color schemes

domoritz:
this PR adds Tableau's color schemes to Vega. I'm personally very excited to use these color schemes for Vega-Lite scatterplots where our current colors are not that great. 

Density color schemes

density

Tableau symbol color schemes

symbol

I'm happy to update the documentation once we decide which colors schemes we want and how we want to name them.


jheer:
I just added a commit that uses the existing, slight compressed, format for color scales. There are a number of outstanding questions:

- What names should we use? I'm partial to using simple names ("blue", "orange", etc) for the symbol palettes. For the density palettes, maybe "lightorange", "lightgrayred", "darkred", etc.
- Should we replace any existing schemes from d3-scale-chromatic, such as blues, oranges, etc?
- These all represent continuous palettes, even though they are defined in terms of discrete color points. It might work better for Vega if we register interpolator functions (i.e., using `piecewise` from d3-interpolate) based on these palettes rather than register the color arrays directly.

domoritz:
> I just added a commit that uses the existing, slight compressed, format for color scales.

Thanks!

> * What names should we use? I'm partial to using simple names ("blue", "orange", etc) for the symbol palettes. For the density palettes, maybe "lightorange", "lightgrayred", "darkred", etc.

I think simpler names are better but we need to make sure to credit Tableau and Maureen appropriately in the code and the docs. 

> * Should we replace any existing schemes from d3-scale-chromatic, such as blues, oranges, etc?

The tableau colors have a more uniform interpolation than the d3 colors but they also don't get as light. I'm not judging here but just noting the differences. I'd be in favor of simplifying the color schemes we offer to make it easier for people to use them.

PR #1579, opened 2019-02-19
Add typing for transform and operator, fix typings for inherits

domoritz:


jheer:
Will take a closer look when updated so that the typings tests pass.

domoritz:
We have a name conflict between `vega.Transform` and the `Transform` type that is the union of all transform interfaces in specifications. I renamed the latter to `Transforms` to resolve the conflict.

domoritz:
Tests are passing again

docs/docs/api/extensibility.md
```diff
@@ -136,11 +136,11 @@ space for the last (largest) color array in *schemes*.
 ## <a name="transform"></a>Transforms
 
 <a name="transforms" href="#transforms">#</a>
-vega.<b>transforms</b>
+vega.<b>transform</b>
 [<>](https://github.com/vega/vega/blob/master/packages/vega-dataflow/src/register.js "Source")
 
 An object that maps transform names to transform constructors. To add a new
-transform to Vega, assign the transform to this object under the desired name.
+transform to Vega, pass the transform to `vega.transform`.
```
domoritz: @jheer I changed this to follow what you suggested in the meeting today.

domoritz: Reverted because this does not exist.

packages/vega-typings/types/spec/transform.d.ts
```diff
@@ -1,6 +1,6 @@
 import { SignalRef, Compare, Vector2, ExprRef, FontWeight, FontStyle, Vector7 } from '.';
 
-export type Transform =
+export type Transforms =
```
pelotom: FYI it's not idiomatic in TypeScript to pluralize the name of a union type, because then a variable representing a single transform looks like `transform: Transforms`. A plural type name should imply that values of that type are somehow plural in nature, e.g.

```ts
type Transforms = Transform[];
```

Besides that, this seems like a needless backwards-breaking change and one made extra confusing by the fact that there is a _new_ `Transform` type, which represents something completely different than the old thing. Just my $.02!

domoritz: I see your points. The new transform type is for something that's actually a Vega property while this is just a Union type. I hope it didn't cause too much trouble and I'll be more mindful with breaking changes in the typings in the future.

PR #1586, opened 2019-02-20
Add unpkg and jsdelivr for vega-loader.

jheer:
Changes:

**vega-loader**

- Update `package.json` to include `unpkg` and `jsdelivr` properties.

### DIFF
--- a/packages/vega-loader/package.json
+++ b/packages/vega-loader/package.json
@@ -20,6 +20,8 @@
   "author": "Jeffrey Heer (http://idl.cs.washington.edu)",
   "main": "build/vega-loader.node.js",
   "module": "index",
+  "unpkg": "build/vega-loader.min.js",
+  "jsdelivr": "build/vega-loader.min.js",
   "repository": "vega/vega",
   "scripts": {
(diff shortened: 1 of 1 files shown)


packages/vega-loader/package.json
```diff
@@ -20,6 +20,8 @@
   "author": "Jeffrey Heer (http://idl.cs.washington.edu)",
   "main": "build/vega-loader.node.js",
   "module": "index",
+  "unpkg": "build/vega-loader.min.js",
```
domoritz: Alternatively, you can add `"browser": build/vega-loader.min.`

jheer: We already use browser in a more nuanced way, and in addition (according to vega/dataflow-api#3) Observable / d3-require doesn't use the browser field.

domoritz: It does but only after the other two. https://github.com/d3/d3-require/blob/master/src/index.js

mbostock: Yep—to clarify, d3-require *will* use the `browser` entry if it’s a string, but not if it’s an object.

PR #1589, opened 2019-02-20
Fix reference to tickCount

### LINKED ISSUE #1588: Undefined reference to `tickCountRef`
`tickCountRef` is referenced in the schema but not defined anywhere.

domoritz:
Fixes #1588


domoritz:
cough Typescript cough 😛

PR #1599, opened 2019-02-21
Update color schemes.

jheer:
Changes:

**vega**

- Update test specifications and scenes for adjusted color scales.
- Drop d3-scale-chromatic dependency.

**vega-encode**

- Remove logic for discretized scale lookup.
- Move interpolation utilities to vega-scale.

**vega-parser**

- Update config to use adjusted color scales.

**vega-scale**

- _**Breaking**_: Remove discrete schemes for continuous data.
  - The `schemeDiscretized` method has been removed. For interpolating schemes, we now always generate discrete schemes via quantization.
  - Scheme names such as `blues-7` are no longer supported; instead use `{"scheme": "blues", "count": 7}`.
- Adjust existing color schemes to avoid extreme light / dark colors.
- Add new color schemes:
  - Single-hue: `browns`, `tealblues`, `teals`, `warmgreys`
  - Multi-hue: `goldgreen`, `goldorange`, `goldred`
  - Density for dark backgrounds: `darkblue`, `darkgold`, `darkgreen`, `darkmulti`, `darkred`
  - Density for light backgrounds: `lightgreyred`, `lightgreyteal`, `lightmulti`, `lightorange`, `lighttealblue`
- Add interpolation utilities: `interpolateColors`, `quantizeInterpolator`
- Drop d3-scale-chromatic dependency.

**vega-typings**

- Update tests to follow test specification changes.

Closes #1578.


jheer:
This is an alternative to / modification of #1578.

jheer:
The updated schemes:

image

domoritz:
Why do some colors have plurals and others don't? `tealblues` vs `greenblue`?

kanitw:
> Why do some colors have plurals and others don't? `tealblues` vs `greenblue`?

If I have to guess, the scheme based on one color would have `s`.  
(Here "tealblue" is a name, not from teal to blue).

domoritz:
Okay, still a bit confusing. @kanitw actually got in wrong on slack 😛

kanitw:
Yes, I agree that it is confusing.   The explanation above is just my attempt to reverse engineering the reasoning :p

jheer:
Kanit got the logic right, but the _real_ reason they are that way is that the original set of names were taken as-is directly from d3-scale-chromatic. I've just kept the scheme consistent...

Issue #1603, opened 2019-02-21
Unsupported object for bins


domoritz:
I thought this would work according to https://vega.github.io/vega/docs/scales/#quantitative. 

```json
{
  "$schema": "https://vega.github.io/schema/vega/v5.json",
  "scales": [
    {
      "name": "x",
      "type": "linear",
      "domain": [1,4],
      "range": [0, 100],
      "bins": [1, 2, 3, 4]
    }
  ]
}
```

jheer:
Thanks. That is definitely a parser bug.

PR #1610, opened 2019-02-22
Expand scale bins options.

### LINKED ISSUE #1601: Convenience function to generate bins
I often see myself writing `sequence(bin_maxbins_10_Rotten_Tomatoes_Rating_bins.start, bin_maxbins_10_Rotten_Tomatoes_Rating_bins.stop + bin_maxbins_10_Rotten_Tomatoes_Rating_bins.step, bin_maxbins_10_Rotten_Tomatoes_Rating_bins.step)`. It would be great to have a function like `bins(bin_maxbins_10_Rotten_Tomatoes_Rating_bins)` instead.

### LINKED ISSUE #1607: Allow passing bin signal to `scale.bins`
This would simplify a lot of specs that use binned data and we would not need #1601 at all.

jheer:
Changes:

**vega**
- Update `nulls-histogram` and `scales-bin` test specifications.

**vega-encode**
- Add bin boundary generation from `bins` parameter object.
- Add domain specification using `bins` property for `BinOrdinal` scales.
- Drop `BinOrdinal` special case for legend label generation, use `bins` instead.

**vega-schema**
- Add scale `bins` object to schema.

**vega-typings**
- Add scale `bins` object to typings.

Closes #1607. Closes #1601.

### DIFF
--- a/packages/vega-encode/src/Scale.js
+++ b/packages/vega-encode/src/Scale.js
@@ -20,6 +20,8 @@ import {
   scaleImplicit
 } from 'vega-scale';
 
+import {range as sequence} from 'd3-array';
+
 import {
   piecewise,
   interpolate,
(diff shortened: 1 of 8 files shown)


jheer:
I don't understand that last comment about new inputs to the bin transform, but feel free to raise that separately as a new issue.

domoritz:
I think the parser needs to be updated (like in https://github.com/vega/vega/issues/1603).

```json
{
  "scales": [
    {
      "name": "x",
      "type": "linear",
      "domain": [1,4],
      "range": [0, 100],
      "bins": {
        "step": 2
      }
    }
  ]
}
```

breaks

PR #1611, opened 2019-02-22
Add font style properties for axis, legend, and title.

jheer:
Changes:

**vega**
- Update Playfair test specification to use top-level axis properties.

**vega-parser**
- Add `labelFontStyle` property for axes and legends.
- Add `titleFontStyle` property for axes and legends.
- Add `fontStyle` property for chart titles.

**vega-schema**
- Add new font style properties to schema.

**vega-typings**
- Add new font style properties to typings.


packages/vega-schema/src/axis.js
```diff
@@ -61,6 +61,7 @@ const axis = object({
   titleColor: colorValue,
   titleFont: stringValue,
   titleFontSize: numberValue,
+  titleFontStyle: stringValue,
```
domoritz: Since there are only three different possible values (https://developer.mozilla.org/en-US/docs/Web/CSS/font-style#Values), would it make sense to use an enum here?

domoritz: Since there are only three different possible values (https://developer.mozilla.org/en-US/docs/Web/CSS/font-style#Values), would it make sense to use an enum here?

jheer: I thought about that, too. Scroll up to see the examples on the linked page.. there are a large number of possible values once you factor in oblique angles. So I decided to stick with a string type. Though I suppose we might do a union type that includes an enum for better schema-based auto-completion? If folks feel strongly about that we could raise it in a future issue.

domoritz: Don't feel super strongly about it but would be nice.

domoritz: Turns out, the typings already have a type for it: https://github.com/vega/vega/blob/master/packages/vega-typings/types/spec/encode.d.ts#L214. 

Using string raises a few warnings in Vega-Lite so I sent https://github.com/vega/vega/pull/1612.

PR #1616, opened 2019-02-22
Change default schemes

domoritz:
Viridis has a number of issues as the default color so we propose to switch to `yellowgreenblue`. It has good discriminability while still looking good on white background. 0 is also still visible. 

Our choice
screen shot 2019-02-22 at 15 33 15

Viridis
screen shot 2019-02-22 at 15 35 42

blueteal
screen shot 2019-02-22 at 15 34 09

tealblues
screen shot 2019-02-22 at 15 34 01

blues
screen shot 2019-02-22 at 15 40 37

### DIFF
--- a/packages/vega-parser/src/config.js
+++ b/packages/vega-parser/src/config.js
@@ -206,7 +206,7 @@ function defaults() {
         scheme: 'blues'
       },
       heatmap: {
-        scheme: 'greenblue'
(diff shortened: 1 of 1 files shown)


domoritz:
We will need to fix the examples if we think this is what we want.

jheer:
Superceded by 9cb38438ee893d38bf30d414251c568c39fee48d.

PR #1618, opened 2019-02-23
Correct typings for `array()`

kanitw:
This simpler typing should match the use case for array better


jheer:
Will merge if @domoritz approves!

domoritz:
I'm honestly a bit surprised this works but it does. Thanks!

PR #1652, opened 2019-02-27
[WIP] Run only scenegraph tests on osx

domoritz:
Still working on this. The goal is to make tests run faster overall.


jheer:
I would *love* to speed up the tests. However, I've found that the Mac OS provisioning and installation times dominate the tests, so I don't think reducing the OSX test footprint to vega-scenegraph will buy us very much -- we still have to go through the same bottlenecking steps, and will have also (unnecessarily?) complicated our testing setup.

Ultimately I think a better solution would be to find a way such that we can run the tests both locally on OSX and in Travis CI on Linux and get consistent results. Any thoughts (short of having two variants of the "correct" test results) on how to achieve this?

domoritz:
Yeah, I was thinking we could try to run the scenegraph tests with puppeteer so that they always run in Chromium, which is consistent regardless of the OS. I want to play with a few more ideas that might make the tests faster and will use this PR to test things out.

jheer:
Scenegraph text rendering for tests was adjusted to ensure identical output on MacOS and Linux. As this PR is no longer needed, I'm closing it out!

PR #1683, opened 2019-03-09
Fix stroke dash support. Add title dx/dy.

jheer:
Changes:

**vega**
- Add `font-size-steps` test specification.

**vega-parser**
- Add `dx`, `dy` properties for title guides.
- Fix scale parser to allow array-valued range entries.

**vega-schema**
- Add `dx`, `dy` properties for title guides.
- Fix schema to allow array-valued range entries.

**vega-typings**
- Add `dx`, `dy` properties for title guides.
- Fix typings for scaled encode entries.

**vega-view**
- Prevent event extension if no active renderer.

### DIFF
--- a/packages/vega-typings/types/spec/scale.d.ts
+++ b/packages/vega-typings/types/spec/scale.d.ts
@@ -10,7 +10,8 @@ export type RangeEnum =
   | 'ramp'
   | 'diverging'
   | 'heatmap';
-export type RangeRaw = (null | boolean | string | number | SignalRef)[];
+export type RangeRawArray = (number | SignalRef)[];
+export type RangeRaw = (null | boolean | string | number | SignalRef | RangeRawArray)[];
 export type RangeScheme =
   | RangeEnum
   | RangeRaw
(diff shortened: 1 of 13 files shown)


jheer:
@domoritz Can you double check the changes I made to encode.d.ts in vega-typings? The issue is that we need to support arbitrary literal values for the `value` property when using `scale`: in those cases the value should match the scale domain type, but it doesn't need to match the encoding channel return type.

jheer:
@domoritz I'm going to go ahead and merge to push things forward now, providing more time for you all to update to Vega 5.2 prior to the VL3 release. But do let me know if you see any issues with the encode typings fixes I added here.

packages/vega-typings/types/spec/scale.d.ts
```diff
@@ -10,7 +10,8 @@ export type RangeEnum =
   | 'ramp'
   | 'diverging'
   | 'heatmap';
-export type RangeRaw = (null | boolean | string | number | SignalRef)[];
+export type RangeRawArray = (number | SignalRef)[];
+export type RangeRaw = (null | boolean | string | number | SignalRef | RangeRawArray)[];
```
domoritz: `RangeRaw` can be an array of `RangeRawArray`s?

jheer:
Yes, because strokeDash.

PR #1691, opened 2019-03-11
Update signal event handling.

### LINKED ISSUE #1689: Resolve "conditional" cycles in dataflow graph?
Vega 5.2 introduced changes in the underlying dataflow graph construction to enforce reranking (topological sorting) of operator chains induced by signal event listeners. This change resolves some concerning issues about execution order (#1678).

However, these changes also eliminated some "escape hatches" that Vega-Lite selections are using to support what I will call "conditional cycles": cycles in the dataflow graph due to dependency chains that are never evaluated simultaneously. That is, the chains are conditioned on particular, non-overlapping event triggers, and so can evaluate without issue in practice.

Here is a basic example when a brush updates in response to both direct input events for "drawing" the extents and scale updates due to panning/zooming (where `(op)` indicates a proxy operator that calculates the value of an update expression and "injects" it into a named signal):

```
                                 scale
                                  v      
input -> (op) -> brush_pixels -> (op) -> brush_data -> (op) -> brush_tuple
                    ^            scale        |
                    |             v           |
                    \----------- (op) <-------/    
```

There are a couple paths for resolving this issue. One approach is to keep Vega as-is, and require clients to never create cycles in the dataflow. In the example above, one could break the chain `brush_pixels -> (op) -> brush_data`, by having `brush_data` calculate an interval in the data domain directly in response to event listeners, along the way duplicating some of the logic behind `brush_pixels`. This approach is clean from a dataflow perspective, but at the cost of some redundant code and computation. Another approach would be to consolidate both pixel- and data-domain information in a singular brush object representation, but this would result in much longer and more complicated expressions.

Instead of (or in addition to) these client-focused approaches, we could add an explicit escape hatch to Vega: add a `"rerank"` property to the event update specification. The `rerank` property would default to `true`, but if set to `false` would skip operator reranking, ensuring that an update has no effect on the topological sort ranking, and thus also bypassing the cycle detection check. The idea is that this would be an opt-in, expert-level feature for people who "know what they are doing" and who would (in theory) only add "conditional cycles" as defined above.

I've implemented initial support for `rerank` and confirmed that it solves the observed issues in Vega-Lite. Adding `"rerank": false` to the "scale_trigger" updates results in correctly-working specifications. It is possible we might use `rerank` to further simplify the design, for example by depending on a scale explicitly (`{"events": {"scale": "scale-name"}, "rerank": false, "update": "..."}`).

Thoughts? If folks are onboard with this plan, I can release `rerank` support in Vega 5.3 in short order.

jheer:
Changes:

**vega**
- Add `scatter-brush-filter` and `scatter-brush-panzoom` test specifications.

**vega-dataflow**
- Streamline signal event handler code.

**vega-view**
- Fix View event extension logic.

Close #1689.


arvind:
As I was able to address the cycle in the dataflow graph without the `rerank` property, my vote would be to not expose this type of low-level control in Vega unless/until we absolutely need it.

jheer:
Thanks @arvind, I've dropped the `rerank` option, and updated this PR to include only minor updates and some new selection-based test examples to help prevent future issues.

PR #1695, opened 2019-03-13
Fix bar chart in the tutorial

akngs:
Production rule for tooltip text in "Let's Make A Bar Chart Tutorial"
and related description contains some mistakes.

Since the tooltip text is not bound to any data, it has no `datum`.
Because of this, `datum === tooltip` always evaluates to false and it
makes `fillOpacity` 1 regardless of mouse hover. The reason why the
example seems to work is because of `translate` transform sends the text
outside of the screen, `(0, -2)`, when mouse is out.


jheer:
Great catch, thank you!

PR #1698, opened 2019-03-15
Manually list the keys for EventStream type

EvsChen:
As mentioned in vega/vega-lite#4620, we may need to manually list the keys to avoid ts json schema building error.

### DIFF
--- a/packages/vega-typings/types/spec/stream.d.ts
+++ b/packages/vega-typings/types/spec/stream.d.ts
@@ -1,4 +1,5 @@
 import { Expr, MarkType } from '.';
+import { type } from 'os';
 
 export type EventSource = EventStream['source'] & {};
 export type EventType =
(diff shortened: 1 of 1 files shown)


packages/vega-typings/types/spec/stream.d.ts
```diff
@@ -21,6 +22,96 @@ export type EventType =
   | 'touchmove'
   | 'touchstart'
   | 'wheel';
+
+export type HTMLBodyEventType =
+  | EventType
+  | 'abort'
```
domoritz: Can we use `keyof GlobalEventHandlersEventMap` instead of the list here?

EvsChen: Referencing to `GlobalEventHandlersEventMap` is the reason why the schema generator fails. It would result in
```bash
UnknownNodeError: Unknown node "GlobalEventHandlers" (ts.SyntaxKind = 165)
```

domoritz: What is syntax kind 165?

jheer: From the linked issue vega/vega-lite#4620:

> Currently it does not have support for SyntaxKind = 165, which is FunctionType.

jheer:
I'm wondering if fixing the ts json schema generator might be a better route than modifying the typings here? I'll leave that for @domoritz to decide.

domoritz:
The schema generator is correct. The issue is that the types include functions and we should not include functions in types that are used to generate a schema.

EvsChen:
> The issue is that the types include functions and we should not include functions in types

Actually I'm a little confused about this. What's the reason for not supporting FunctionType for the schema?  This type seems quite common in cases like, event handler. Or is there any way to skip this type when generating the schema?

domoritz:
There are no functions in json so json schema does not support it either.

jheer:
Hey @domoritz, can you let me know if there is something we should be doing here, and if we should be looking to merge or close out this PR?

domoritz:
Not needed anymore since we now have https://github.com/vega/ts-json-schema-generator/pull/130

PR #1699, opened 2019-03-15
vega-typings: add datum to SceneGroup

danmarshall:


jheer:
I think `datum` should be part of `SceneItem`, not just `SceneGroup`. Also flagging @domoritz, as I believe he originally wrote these typings.

danmarshall:
@jheer thanks. I'm also sure that the `any` is not perfect, depending on the `SceneGroup.mark.role`.

jheer:
Ah yes... `datum` should be of type `object`.

PR #1700, opened 2019-03-15
Update bin method to adjust zero-span extents.

kanitw:
Fix https://github.com/vega/vega-lite/issues/4717


jheer:
Superceded by #1704.

PR #1706, opened 2019-03-16
Vega-Utils in Typescript

domoritz:
New version of https://github.com/vega/vega/pull/1566

Please squash when merging.


domoritz:
Feel free to just squash this.

domoritz:
Do you want me to make any changes @jheer?

jheer:
Not at the moment. The main issue for now is that this has the lowest priority on my much-too-long TODO list. Plus, I'm still ultimately unsure if I want to switch to TypeScript and need to do more careful consideration. 🤷‍♂️

jheer:
After working more with TypeScript over in vega/vega-lite, I've decided not to initiate a port from JS to TS for the core Vega projects. (At least, not in any near-term future.) As a result, I'm closing out this PR as I want to minimize inconsistent languages and styles within this monorepo -- vega-typings being an obvious and necessary exception.

Thanks again for your efforts here. We could have a separate conversation about whether or not it makes sense for vega-util to leave the monorepo as a stand-alone project (I'm skeptical, but it's a possibility), in which case a TypeScript conversion might be more fitting.

PR #1707, opened 2019-03-17
Incorrect types for Legend typings

marcprux:
The `Legend` typings `size`, `shape`, `strokeDash`, `strokeWidth`, and `opacity` are all `string`, which seems incorrect. I believe this is what leads to the vega-lite schema for `Legend` not showing `strokeWidth`, even though it is present in `LegendConfig`: https://github.com/vega/vega-lite/blob/f72d8be7fba154437ceaf51ec332d066c1e89313/build/vega-lite-schema.json#L7311

### DIFF
--- a/packages/vega-typings/types/spec/legend.d.ts
+++ b/packages/vega-typings/types/spec/legend.d.ts
@@ -53,13 +53,13 @@ export type LegendOrient =
   | 'bottom-right';
 
 export interface Legend extends BaseLegend {
-  size?: string;
-  shape?: string;
+  size?: number;
+  shape?: SymbolShape;
   fill?: string;
   stroke?: string;
-  strokeDash?: string;
-  strokeWidth?: string;
-  opacity?: string;
+  strokeDash?: number[];
+  strokeWidth?: number;
+  opacity?: number;
 
   /**
    * The type of legend to include. One of `"symbol"` for discrete symbol legends, or `"gradient"` for a continuous color gradient. If gradient is used only the fill or stroke scale parameters are considered. If unspecified, the type will be inferred based on the scale parameters used and their backing scale types.


domoritz:
I don't think so. The expected type here is the name of the scale. See https://github.com/vega/vega/blob/fd508ee9ef92a54d029dfc8a87be931ddec01584/packages/vega-schema/src/legend.js#L46. However, we should not use the legend directly in Vega-Lite.

domoritz:
We then add `strokeWidth` in https://github.com/vega/vega/blob/fd508ee9ef92a54d029dfc8a87be931ddec01584/packages/vega-typings/types/spec/config.d.ts#L405. I think it's correct that the config supports signals.

marcprux:
I didn't realize those properties were the names of the scales for those legend properties. I still wonder, though, if it has something to do with the absence of `strokeWidth` in vega-lite's `Legend`. The properties that exist in `LegendConfig` but are absent from `Legend` seem to correlate with the properties that have string types for the scales in the vega typings.

For example, I notice that VL's `Legend` uses `strokeColor` instead of `stroke` for the color; might this name have been chosen to avoid conflicting with the legend's `stroke` scale name?

domoritz:
Vega legends use `gradientStrokeWidth` to set the stroke width of gradient legends (similar for symbol legends).

domoritz:
Actually, I'm wrong. The `gradientStrokeWidth` is for the actual gradient, not around the whole legend.

domoritz:
I somehow don't get the stroke around the legend to work, though.

```json
{
  "autosize": "pad",
  "data": [
    {"name": "source_0", "url": "data/cars.json", "format": {"type": "json"}}
  ],
  "scales": [
    {
      "name": "color",
      "type": "ordinal",
      "domain": {"data": "source_0", "field": "Origin", "sort": true},
      "range": "category",
      "interpolate": "hcl"
    }
  ],
  "legends": [
    {
      "stroke": "color",
      "title": "Origin"
    }
  ],
  "config": {
    "legend": {
      "strokeWidth": 10,
      "stroke": "red",
      "titleColor": "red"
    }
  }
}
```

image

domoritz:
Either way, I don't think the typings are wrong here.

marcprux:
You're probably right, but I'm still wondering why `strokeWidth` is available for `LegendConfig` but not `Legend`? (ditto for `strokeDash`)

domoritz:
Because Vega uses top level properties of legends for both scale references and to set specific properties there is no way around this. Maybe legends should not do this to avoid this confusion (that we have hit in Vega before as well).

marcprux:
Ouch! I wasn't aware of that limitation. It certainly seems odd that a user can set both `strokeColor` and `strokeWidth` in `LegendConfig` but they can only set `strokeColor` in `Legend` (because it doesn't conflict with the `stroke` or `strokeWidth` properties). But I guess that should be a topic for a separate issue – I'll close this one and make a new one.

domoritz:
You can set the `strokeWidth` in the `encode` block so it's not a limitation (only inconvenient).

marcprux:
Except you can't specify an `encode` block from vega-lite, which is how I initially encountered the issue. Granted, I could generate the vega spec and then inject an `encode` block, but I feel that crosses the line from "inconvenient" to "limitation".

domoritz:
I think you can even though we don't allow it in the typings. Having said that, we could redirect the `strokeWidth` in Vega-Lite to an encode block in Vega. However, a solution in Vega would be nice so we don't have to add this hack in Vega-Lite.

marcprux:
I tried some variants of embedded `encode` blocks from vega-lite, and none of them work. That'd be a great feature, though!

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v3.json",
  "description": "A bar chart that directly encodes color names in the data.",
  "data": {
    "values": [
      {"color": "red", "b": 28},
      {"color": "green", "b": 55},
      {"color": "blue", "b": 43}
    ]
  },
  "mark": "bar",
  "encoding": {
    "x": {"field": "color", "type": "nominal"},
    "y": {"field": "b", "type": "quantitative"},
    "color": {
      "field": "color",
      "type": "nominal",
      "legend": {
        "strokeColor": "purple",
        "encode": {"legend": {"update": {"strokeWidth": {"value": 10}}}}
      }
    }
  }
}
```

domoritz:
Hmm, I don't have time to dig into this right now but we do use https://github.com/vega/vega-lite/blob/master/src/compile/guide.ts#L8. It's not exposed in the schema but it should work.

domoritz:
Either way, you can file an issue with Vega about the scales vs properties issue.

PR #1729, opened 2019-03-25
Upgrade dependencies

domoritz:


jheer:
It looks like the dtslint bug persists, even with the new patch version? Travis is still failing on vega-typings and I'm getting the same errors locally.

domoritz:
I sent a fix in https://github.com/Microsoft/dtslint/pull/220. I guess we are blocked by that.

domoritz:
Should be good now.

Issue #305, opened 2015-07-22
Gradient and image fill


rbishop-bah:
I would very much like to be able to use gradients and/or images instead of colors for the fill attributes of marks.

Any chance of adding a Paint type that's one of (color, gradient, image) that can be used in many places where only color can be used now used?

jheer:
Duly noted. While gradients and images are **not** currently part of the official API, there does exist internal support for linear gradients. (For example, to create a legend with a color ramp.)

You can specify a linear gradient in a spec like so:
```
{
  "id": "rgb",
  "x1": 0.0,
  "y1": 0.0,
  "x2": 0.0,
  "y2": 1.0,
  "stops": [
    {"color": "red", "offset": 0.0},
    {"color": "green", "offset": 0.5},
    {"color": "blue", "offset": 1.0}
  ]
}
```
The x and y coordinates give the start and end points of the gradient (in a fractional [0,1] coordinate space). The items being drawn with a gradient must have width and height values. The color stops give the color values and fractional offsets in [0,1]. Note that the `id` property in the gradient is required.

To test this out, open the [bar chart example in the vega-editor](http://vega.github.io/vega-editor/?spec=bar) and replace `"steelblue"` in the enter fill color with the snippet above. 

Image fills and radial gradients are not currently on the roadmap, so I'm leaving this issue open as a feature request for now. We'd happily consider PRs along those lines. Note also that until officially supported, gradient definitions in a spec may not pass validation against the Vega JSON schema.

rbishop-bah:
Very interesting, thanks for the example.

darrengriffith:
I'm trying to do a radial gradient or something close for a pie chart.  I can manually change the x1, y1, x2, y2 attributes to "rotate" the gradient for each pie slice, but I can't figure out if there's any way to do this programmatically.  Any ideas about how I might achieve this, or are there any further plans to implement these gradients?

mattijn:
I tried the provided snippet in an area chart, but this removes the whole chart except the axes. A pity, also would be interested in this feature

anagrius:
Gradients would be very nice for shape marks.

kanitw:
I'm curious if there is any reason that prevents us from making this an official feature? (Besides more work to document the feature.)

marcopesani:
Having the possibility to use patterns is quite important for accessibility. Relying on colors is a huge problem for people suffering from color blindness, background patterns are the way to go.

kanitw:
[...]

PR #1734, opened 2019-03-26
Correct GradientValue in docs

csgibson:
Correct GradientValue in types.md; add example


csgibson:
Noticed this while looking at @jsalankey's comment in https://github.com/vega/vega/issues/305

jheer:
Thanks!

PR #1741, opened 2019-03-30
Add number coercion to stack transform.

jheer:
Changes:

**monorepo**:
- Update all dev dependencies.

**vega-encode**:
- Add number coercion to stack transform.
- Add test cases for stack transform.

Close #1740


domoritz:
Thank you!

PR #1750, opened 2019-04-03
isNumber: add documented values and missing one to tests

clemens-tolboom:
On https://vega.github.io/vega/docs/expressions/#isNumber it states

> NaN and Infinity are considered numbers.

So let's add those.

~~I myself found empty string allowed by isNumber so hope this PR will fail.~~
PR should pass (as empty strings are false under isNumber)


domoritz:
Thank you for the new tests! Can we wait with this after https://github.com/vega/vega/pull/1706 is merged (if it is) to avoid conflicts?

clemens-tolboom:
> Thank you for the new tests! Can we wait with this after #1706 is merged (if it is) to avoid conflicts?

@domoritz sure.

jheer:
Thanks!

Issue #1502, opened 2018-12-20
Generate schema from typings


domoritz:
## Advantages
* Consistent schema (good for Altair)
* Only one source of truth. Schema and types will not become inconsistent again. 
* Consistency with Vega-Lite

## Disadvantages
* Need to develop infrastructure to support custom schema builds with a subset of the transforms.

hydrosquall:
This seems useful!

Does the infrastructure cost become lower or stay the same if interfaces in vega-typings were directly replaced with typescript code? Currently there is some drift between the vega-typings files and the js implementation.

PR #1762, opened 2019-04-10
[WIP] Add description to specification properties

algomaster99:
- This adds documentation to the properties of vega specifications so that the documentation can also be displayed for vega specification in the editor.
- Formatting of the JSON file is automatically done which is the reason for so many deletions.

Screenshot from 2019-04-10 13-21-03


algomaster99:
@domoritz I have added the docs by referring to the original vega docs. Please review and suggest if there are any changes.

domoritz:
Wow, nice. How did you do it? Can we add this process to the Vega repo directly so that the Schema has docs everywhere?

jheer:
Thanks!! This PR updates the schema file published alongside the Vega docs. However, this file is automatically overwritten upon new Vega releases, and the version in this PR differs from the official schema published elsewhere. The "single source of truth" is the schema generation code in `packages/vega-schema`. Can we add the changes there instead? Thanks again!

algomaster99:
@jheer For some properties like `height`, `width` there is no file such as `height.js` or `width.js` respectively. So for where the description should be written?
For properties such as `autosize`,
```
export default {
  defs: {
    autosize,
    description: "Sets how the visualization size should be determined. If a string, should be one of `\"pad\"`, `\"fit\"` or `\"none\"`.\nObject values can additionally specify parameters for content sizing and automatic resizing.\n`\"fit\"` is only supported for single and layered views that don't use `rangeStep`.\n\n__Default value__: `pad`"
  }
};
```
This can be done right?

jheer:
The description needs to be _inside_ the autosize object definition, _not_ in the enclosing `def` or `ref` block. Something like this, perhaps?

```js
const autosize = oneOf(
  autosizeType,
  object({
    _type_: autosizeType,
    resize: booleanType,
    contains: enums(containsEnum)
  })
);

autosize.description = 'description text';
```

Top level width and height are defined in `src/schema.js`, but will need new objects. Instead of `numberType`, we'd need objects like `{"type": "number", "description": "desc text"}`. You could use `type` method from `src/util.js` to do this like so: `type('number', {description: 'desc text'})`.

That said, I wonder if it might make sense to maintain the descriptions in completely separate files, perhaps in a parallel schema structure that can be merged with the generated schema? That might make adding internationalization (non-English descriptions) easier in the future.

algomaster99:
@jheer Okay, I'll make the modifications and commit them.

algomaster99:
@domoritz @jheer I have tried and separated description into different files. Although the properties under scope fail to recognise the description attribute. But for most of them it works.

#### List of all the attributes with their description
- [x] autosize
- [ ] axes
- [x] background
- [ ] config &rarr; `description` being overridden by `usermeta`
- [ ] data
- [ ] description &rarr; `description` being overridden by `background`
- [x] encode
- [x] height
- [x] layout
- [ ] legends
- [ ] marks
- [x] padding
- [ ] projections
- [ ] scale
- [ ] signals
- [ ] style
- [x] title
- [x] usermeta
- [x] width

schema

domoritz:
Thank you for the PR. I think it's probably best not to focus on translation for now so we can have fewer changes. Can you add the description directly to the existing files rather than splitting existing files?

algomaster99:
@domoritz Okay, I will do that.

algomaster99:
@domoritz I have made the changes.
And if you're merging it, please don't squash because we may implement i18n standards in the future so commit 
9376df6  can be referred at that time.

packages/vega-schema/src/autosize/index.js
```diff
@@ -0,0 +1,35 @@
+import {enums, object, oneOf, booleanType} from '../util';
```
domoritz: this file should remain `autosize.js`

algomaster99:
@domoritz I have changed the directory structure as well.

packages/vega-schema/src/autosize.js
```diff
@@ -24,6 +24,10 @@ const autosize = oneOf(
   })
 );
 
+const DESCRIPTION = "Sets how the visualization size should be determined. If a string, should be one of `\"pad\"`, `\"fit\"` or `\"none\"`.\nObject values can additionally specify parameters for content sizing and automatic resizing.\n`\"fit\"` is only supported for single and layered views that don't use `rangeStep`.\n\n__Default value__: `pad`";
+
+Object.assign(autosize, { description: DESCRIPTION });
```
domoritz: I think it would be better to put the text into the object itself rather than assigning it later.

algomaster99: @domoritz The text can be too long. That is why I created a new variable.

domoritz: I still think we can avoid the assign.

algomaster99: @domoritz okay I will remove the `DESCRIPTION` variable.

algomaster99: @domoritz Done

domoritz:
I think this needs some more work and maybe the better approach is to do https://github.com/vega/vega/issues/1502 instead.

PR #1766, opened 2019-04-17
add scale method type to runtime

coffeedoughnuts:
Hi,

I'm new to Vega; really just experimenting with it at the moment.

I noticed that the typescript definitions is missing the scale method on a Vega view, as documented here: https://vega.github.io/vega/docs/api/view/#view_scale

I'm not sure I have the return type correct; but the `Scale` type from `vega-typings/types/spec/scale.d.ts` doesn't seems to be what is being returned when I check at run time (for example, I am able to run the following: `view.scale('x')(new Date("2019-04-17"))` and get back a number, but the Scale type looks to be something else..)

Happy to make or accept changes on my fork; for now my workaround is to cast the view to any which is not ideal (`(view as any).scale('x')`)

Cheers,
Daniel

### DIFF
--- a/packages/vega-typings/types/runtime/index.d.ts
+++ b/packages/vega-typings/types/runtime/index.d.ts
@@ -1,4 +1,4 @@
-import { DataType, EncodeEntryName, Format, SignalValue, Spec, Scale } from '../spec';
+import { DataType, EncodeEntryName, Format, SignalValue, Spec } from '../spec';
 import { Renderers } from './renderer';
 
 // TODO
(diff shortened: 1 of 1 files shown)


jheer:
Thanks! The return type can be quite varied. Particularly for ordinal scales, where the range can consist of arbitrary values (even arrays, for example defining stroke dash arrays). So I think `any` is an appropriate right return type.

jheer:
Also, you'll probably want to rebase on master when making updates, which should resolve some CI test failure issues!

PR #1772, opened 2019-04-19
Add DSV header to schema. Update dev dependencies.

jheer:
Changes:

**vega-schema**
- Add DSV/CSV/TSV header property to schema.

**vega-typings**
- Add DSV/CSV/TSV header property to typings. Refactor DSV typings.

**monorepo**
- Update dev dependencies.


jheer:
@domoritz Any ideas why dtslint would fail on Travis but succeed locally? I see the same errors popped up during CI for your PR #1771. All dev dependencies should be up to date here.

domoritz:
I saw this earlier today and was equally confused why it works locally but not on travis. I'm looking into it.

jheer:
I'm solving the issues on Travis by patching up the typings, so don't worry too much about that. If you can figure out the discrepancy between local and Travis, that would be very helpful.

PR #1774, opened 2019-04-19
Update tslint.json

domoritz:


domoritz:
We don't really need this but I don't see why we wouldn't make the test more strict.

PR #1776, opened 2019-04-20
Move usermeta from schema to scope

marcprux:
The `usermeta` property is very useful for tools that need to generate interoperable vega spec while still being able to retain tool-specific metadata. It would be even more useful if it was not constrained to top-level specs so that, for example, each sub-spec could store its own tool-specific identifier. This is implemented by simply moving `usermeta` from the top-level schema object into the `scope` def.

If this is accepted it should also be enabled in `vega-lite`.


jheer:
Thanks! This looks totally reasonable to me. I just restarted the CI tests. Once everything passes we should be able to merge.

cc: @domoritz @kanitw to check for parity with Vega-Lite

jheer:
Merged. Thanks!

domoritz:
Vega-Lite just passes it through if it's defined at the top level.

PR #1783, opened 2019-04-25
Add KDE transform.

jheer:
Changes:

**vega**
- Add `kde` test specification.

**vega-transforms**
- Add KDE transform and tests.

**vega-typings**
- Add KDE transform to typings.


packages/vega-typings/types/spec/transform.d.ts
```diff
@@ -252,6 +253,17 @@ export interface JoinAggregateTransform {
   as?: (string | SignalRef | null)[] | SignalRef;
 }
 
+export interface KDETransform {
+  type: 'kde';
+  field: string | TransformField | SignalRef;
+  groupby?: (string | TransformField)[] | SignalRef;
+  method?: 'pdf' | 'cdf' | SignalRef;
+  bandwidth?: number | SignalRef;
+  extent?: number[] | SignalRef;
```
domoritz:
```suggestion
  extent?: [number, number] | SignalRef;
```

jheer: Thanks! I've pushed some minor refinements to the transform typings in response.

PR #1784, opened 2019-04-26
Add regression transform.

jheer:
Changes:

**vega**
- Register regression transform.
- Add `regression` test specification.

**vega-schema**
- Add regression transform to schema.

**vega-regression**
- Create regression package, add regression transform.

**vega-transforms**
- Fix KDE transform groupby access.

**vega-typings**
- Add regression transform to typings.


domoritz:
🎉

PR #1785, opened 2019-04-28
Update types.md

kanitw:
Note that the following  statement still confuses me 

> [signal references](#Signal) that are evaluated once per parameter

Basically, mark encoding set can include signals, but signals are evaluated independently per data object in such cases. 

(So it seems like signals are  evaluated once per parameter for transform parameter, but not for mark encoding set value.)


jheer:
Thanks!

PR #1786, opened 2019-05-02
Add value update checks to encoder functions.

jheer:
Changes:

**vega-encode**
- Add logic to `Encode` transform to avoid suppression of unchanged `mod` tuples if non-datum fields have been flagged as modified (e.g., as the result of a `modify` trigger).

**vega-force**
- Update `Force` transform so that it is always flagged as modified. This ensures that downstream operators (namely `linkpath`) can always re-evaluate in response.
- Update `Force` to re-initialize when the `restart` parameter turns on, regardless of tuple change status.

**vega-parser**
- Add value update checks to generated encoder functions, add only updated items to the `mod` set. Results in more efficient re-rendering when only a subset of items have encoded values modified.

### DIFF
--- a/packages/vega-encode/src/Encode.js
+++ b/packages/vega-encode/src/Encode.js
@@ -39,17 +39,19 @@ prototype.transform = function(_, pulse) {
       update = encoders.update || falsy,
       enter = encoders.enter || falsy,
       exit = encoders.exit || falsy,
-      set = (encode && !reenter ? encoders[encode] : update) || falsy;
+      set = (encode && !reenter ? encoders[encode] : update) || falsy,
+      // if item (not datum) fields are modified, don't suppress mod output
+      fieldmod = pulse.fields
+              && Object.keys(pulse.fields).some(f => f !== 'datum');
 
   if (pulse.changed(pulse.ADD)) {
-    pulse.visit(pulse.ADD, function(t) {
-      enter(t, _);
-      update(t, _);
-      if (set !== falsy && set !== update) set(t, _);
-    });
+    pulse.visit(pulse.ADD, function(t) { enter(t, _); update(t, _); });
     out.modifies(enter.output);
     out.modifies(update.output);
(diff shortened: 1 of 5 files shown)


jheer:
@arvind: Unfortunately adding update check logic to the encoder functions results in downstream complications. Unlike Vega 2, v3+ allows transforms to be added to the dataflow directly _after_ the encoders. If the encoder suppress all `mod` tuples, the downstream transforms may be "starved" of items they would otherwise process and potentially modify.

That said, I believe the only existing example that currently breaks is `force-network`. I added some changes to this PR that updates the `Force` and `Encode` transforms to fix that example. The changes are reasonable if not exactly elegant (though there are already other, more inelegant, aspects elsewhere in the code base!), but in any case these do not fix the more general problem.

So I'm hesitant about this change. On the plus side, it does improve performance in multiple situations we care about: both for existing conditional encodings, and when highlighting points using the `intersect` functionality. In the latter case the performance improvements can be quite significant. On the down side, I could easily construct example specs that would now break, even if they are unlikely to arise in practice. On the whole, this dissuades me from merging this PR in its current state.

Unfortunately, I have yet to come up with an alternative scheme for checking value updates after the encoder and any additional encoding transforms have run, short of copying all values up front and then comparing them all at then end of the pipeline. While that makes sense for tasks like animated transitions, I'd prefer not to have that overhead for standard interactive dataflows.

If you have any thoughts on the matter, let me know!

jheer:
I came up with a working alternative scheme that does not involve extra overhead: when parsing a mark specification, analyze any included post-encoding transforms to see if the encoder can safely cull `mod` tuples, or if they must be preserved. This determination then sets a new parameter flag for the `Encode` transform to control how `mod` tuples are processed. Assuming this works out, PR #1788 will replace (and close out) this one.

arvind:
Yes, this sounds very similar to what we did with Vega 2 (spec analysis -> setting flags on data flow assembly -> optimize data flow processing). Glad it translates to the 3+ world.

PR #1793, opened 2019-05-03
[docs] clarify the meaning of impute's "method"

nyurik:
I had to read "group by" to understand the meaning of "method", so would be good to clarify. May need a better language.


jheer:
Thanks!

PR #1796, opened 2019-05-03
Add isDefined and isValid expression functions.

### LINKED ISSUE #1792: Support `undefined` in expressions
It seems that Vega does not allow `undefined` as a known keyword inside the expressions, similar to `null`.  Undefined is very useful to check if a value is present in the datum, plus many built-in Vega functions return undefined as part of their normal operations, so it should be possible to use it.

Currently using `undefined` in an expression shows `Error: Unrecognized signal name: "undefined"`

I'm not sure if this is a bug or a feature request.

jheer:
Changes:

**vega-functions**
- Add `isDefined` and `isValid` expression functions.

Fix #1792.


jheer:
Lambda expressions are anonymous functions, and can be written either with the `function` keyword or as arrow (`=>`) functions. I assume you mean arrow functions?

For these new functions (`isDefined`, `isValid`) either should work. In general, however, some Vega expression functions rely on the binding of the `this` context, and so absolutely must be defined with the `function` keyword. Thus for consistency I prefer to keep these expression functions defined using `function`.

nyurik:
@jheer, thanks for the explanation, I thought of `this` binding, but wasn't sure if that was the reason here.  I doubt there is much performance diff between the arrow and regular functions, so consistency definitely wins :)

PR #1804, opened 2019-05-16
Add event configuration security controls.

jheer:
Changes:

**vega-view**
- Add security event config for allowable event sources and types.
- Add security event config for allowable input binding targets.

**vega-typings**
- Update typings with new event config options.

### DIFF
--- a/packages/vega-typings/types/spec/config.d.ts
+++ b/packages/vega-typings/types/spec/config.d.ts
@@ -37,7 +37,12 @@ export interface Config
   background?: string;
   group?: any; // TODO
   events?: {
+    bind?: 'any' | 'container' | 'none';
     defaults?: DefaultsConfig;
+    selector?: boolean | string[];
+    timer?: boolean;
+    view?: boolean | string[];
+    window?: boolean | string[];
   };
   style?: any; // TODO
   legend?: LegendConfig;
--- a/packages/vega-view/src/View.js
+++ b/packages/vega-view/src/View.js
@@ -58,6 +58,9 @@ export default function View(spec, options) {
   view._eventListeners = [];
   view._resizeListeners = [];
 
+  // initialize event configuration
+  view._eventConfig = initializeEventConfig(spec.eventConfig);
+
   // initialize dataflow graph
   var ctx = runtime(view, spec, options.functions);
   view._runtime = ctx;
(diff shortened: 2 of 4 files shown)


jheer:
Draft documentation update for https://vega.github.io/vega/docs/config/#events

| Property      | Type                                  | Description    |
| :------------ | :-----------------------------------: | :------------- |
| bind          | {% include type t="String" %}         | Configuration control for binding input DOM elements to signals. The available options are `"any"` (default, all bindings are allowed), `"container"` (use only the view container DOM element for all bindings, suppressing per-binding selectors), and `"none"` (suppresses all input bindings). |
| defaults      | {% include type t="Object" %}         | An object describing which events that originate within the Vega view should have their default behavior suppressed by invoking the [event.preventDefault](https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault) method. The _defaults_ object should have a single property: either `"prevent"` (to indicate which events should have default behavior suppressed) or `"allow"` (to indicate only those events whose default behavior should be allowed). This property accepts either a boolean value (to prevent/allow all events) or an array of event type strings.|
| selector      | {% include type t="Boolean or String[]" %} | Configuration control for event listeners for external sources specified using a CSS selector. If a boolean value, `true` (default) permits selector event listeners, `false` to disallow all selector events. If a string array, the entries specify a whitelist of which event types (such as `"mousemove"` or `"wheel"`) are allowed. |
| timer         | {% include type t="Boolean" %}        | Configuration control for event listeners for a `"timer"` source. One of `true` (default) to permit timer event listeners, or `false` to disallow timer events. |
| view          | {% include type t="Boolean or String[]" %} | Configuration control for event listeners for the Vega `"view"` source. If a boolean value, `true` (default) permits view event listeners, `false` to disallow all view events. If a string array, the entries specify a whitelist of which event types (such as `"mousemove"` or `"wheel"`) are allowed. |
| window        | {% include type t="Boolean or String[]" %} | Configuration control for event listeners for the browser `"window"` source. If a boolean value, `true` (default) permits window event listeners, `false` to disallow all window events. If a string array, the entries specify a whitelist of which event types (such as `"mousemove"` or `"wheel"`) are allowed. |

nyurik:
This relates to the [Wikipedia feedback](https://phabricator.wikimedia.org/T172938#3693347).
CC: @Bawolff

PR #1819, opened 2019-06-01
Separate `NonArgAggregateOp` from AggregateOp

kanitw:
In Vega-Lite's fielddef, we only support `NonArgAggregateOp` (and Argmin/maxDef)


kanitw:
@domoritz Actually should we instead use `Exclude` in VL and avoid changing the main `vega-typings`?

domoritz:
You can do that but make sure that it works with the schema generator. I have not had the time to add support for Exclude.

kanitw:
Instead of polluting vega-typings, I think it's better to wait for the exclude support and do it cleanly in VL.

PR #1833, opened 2019-06-11
correct order of size and nodeSize parameters

111andre111:
clarify the meaning of the definition of nodeSize


jheer:
Hmm. The original documentation should be correct, as we simply pass the parameter to the D3 layout methods. For example, see https://github.com/d3/d3-hierarchy#tree_nodeSize

If you are seeing different behavior, please file a bug report with a replicating specification. I'm wondering if perhaps the orientation of the overall diagram (e.g., left-to-right vs. top-to-bottom) might be affecting the results you're seeing.

PR #1839, opened 2019-06-14
Add selection function for comparison test

djbarnwal:
Adds Selection Comparison function `vlComparisonTest` 

vega/vega-lite#5068 uses this function.

### DIFF
--- a/packages/vega-functions/src/codegen.js
+++ b/packages/vega-functions/src/codegen.js
@@ -11,7 +11,8 @@ import {
 import {
   selectionTest,
   selectionResolve,
-  selectionVisitor
+  selectionVisitor,
+  comparisonTest
 } from 'vega-selections';
 
 import {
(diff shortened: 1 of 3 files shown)


jheer:
Thanks for the PR! Adding @arvind to review prior to merge.

jheer:
I pushed some changes to clean up the code and also removed the use of the Function constructor. Compiling a new function for every invocation is problematic from both a performance and a security perspective.

packages/vega-functions/src/codegen.js
```diff
@@ -274,6 +275,7 @@ expressionFunction('treeAncestors', treeAncestors, dataVisitor);
 
 // register Vega-Lite selection functions
 expressionFunction('vlSelectionTest', selectionTest, selectionVisitor);
+expressionFunction('vlComparisonTest', comparisonTest, selectionVisitor);
```
jheer: The name `comparisonTest` is rather generic, and breaks the convention of naming selection-specific methods with the `selection` prefix. Can we name it something like `selectionCompare` instead? The registered expression function name would then be `vlSelectionCompare`.

arvind:
Stale and superseded by #3675

PR #1841, opened 2019-06-17
Upgrade dependencies

domoritz:


domoritz:
This waited too long and got out of date. I'll make a new one later.

Issue #608, opened 2016-07-28
Custom tick formatting function


dsschwarz:
I saw that this issue was closed https://github.com/vega/vega/issues/194
However, I'm finding that the d3 number format and the template syntax don't seem to be able to cover my requirements.

I'm looking to customize the SI suffixes (i.e. use B instead of G for 1e9). That's possible to do by overriding some d3 methods, but it's not ideal. Having a custom formatting function would be very helpful here.

domoritz:
Here is an example of how you can use expressions to customize axis labels with an expression: https://github.com/vega/vega-lite-api/issues/12#issuecomment-506739847.

PR #1853, opened 2019-06-27
WIP Custom tick formatting function

curran:
First step proposal towards #608

This is horribly incomplete as a PR I do realize (no tests, have not even verified if it works), but I wanted to put this forth as a proposed direction to move in.

Any thoughts? What would this break? It would be outside the JSON Schema for Vega specs definitely, but maybe that's OK.

Related discussion in https://github.com/vega/vega-lite-api/issues/12

### DIFF
--- a/packages/vega-functions/src/format.js
+++ b/packages/vega-functions/src/format.js
@@ -21,7 +21,9 @@ function formatter(type, method, specifier) {
 }
 
 export function format(_, specifier) {
-  return formatter('format', d3_format, specifier)(_);
+  return typeof specifier === 'function'
+    ? specifier
+    : formatter('format', d3_format, specifier)(_);
 }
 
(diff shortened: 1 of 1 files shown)


domoritz:
You can add custom functions with https://vega.github.io/vega/docs/api/extensibility/#expressions. I don't think we necessary need this addition, although it is certainly less cumbersome.

curran:
Expression functions could work!

At the end of the day, what I'd really like to see is a working implementation that formats this original example https://observablehq.com/@mbostock/working-with-wikipedia-data using "$90B" rather than "90G".

domoritz:
I think our discussion in https://github.com/vega/vega-lite/issues/5122 shows that we should go a different route first and see whether it's too limited.

curran:
Sounds good. Thank you for moving this forward!

PR #1855, opened 2019-06-28
Add naturalEarth1 projection type

chucklam:
Vega's [projection types](https://vega.github.io/vega/docs/projections/#projection-types) support `naturalEarth1` but it was missing in the typings.

I discovered this bug while using the Vega-Lite schema, so this may be an additional data point to discussion #1502 .


domoritz:
Thank you!

PR #1881, opened 2019-07-21
docs: gradient stops is required

domoritz:


kanitw:
We won't need to do this manually if we read from TS-based json schema :p

PR #1883, opened 2019-07-21
Add comments to gradient types

domoritz:
Merge after #1882


packages/vega-typings/types/spec/encode.d.ts
```diff
@@ -81,27 +81,75 @@ export interface ColorHCL {
   l: NumericValueRef;
 }
 export interface GradientStop {
+  /**
+   * The offset fraction for the color stop, indicating its position within the gradient.
+   */
   offset: number;
-  color: string;
+  /**
+   * The color value at this point in the gradient.
```
kanitw: point -> stop?  (given the class name is stop)

packages/vega-typings/types/spec/encode.d.ts
```diff
@@ -81,27 +81,75 @@ export interface ColorHCL {
...
+   * The type of gradient. Use `"linear"` for a linear gradient.
+   */
   gradient: 'linear';
+  /**
+   * An array of gradient stops defining the gradient color sequence.
+   */
   stops: GradientStop[];
   id?: string;
+  /**
+   * The starting x-coordinate, in normalized [0, 1] coordinates, of the linear gradient (default 0).
```
kanitw: Better consistently use `__Default value:__`?  Same for other ones with defaults.

packages/vega-typings/types/spec/encode.d.ts
```diff
@@ -81,27 +81,75 @@ export interface ColorHCL {
...
+   */
   offset: number;
-  color: string;
+  /**
+   * The color value at this point in the gradient.
+   */
+  color: Color;
 }
 export interface GradientLinear {
+  /**
```
kanitw: Better have GradientLinear/ a generic gradient type so that docs for type can describe both types. 

```
gradient: 'linear' | 'radial'  // could be a generic T that extends 'linear' | 'radial' too 
```

packages/vega-typings/types/spec/encode.d.ts
domoritz: The stop defines a color at a point. I think it's right.

packages/vega-typings/types/spec/encode.d.ts
domoritz: Done

packages/vega-typings/types/spec/encode.d.ts
domoritz: Done

Issue #1807, opened 2019-05-17
Rename GradientLinear/Radial to Linear/RadialGradient in vega-typings?


kanitw:
We will likely import this to Vega-Lite.  Thus, it will show up in the schema and Altair.

So Linear/RadialGradient might be a more natural name?

domoritz:
Note that we also call it in that order in the schema: https://github.com/vega/vega/blob/235572cece28d80efae3866d40821151279e48e5/packages/vega-scenegraph/schema.js#L122. I think the typings and schema should be consistent.

jheer:
PR?

jheer:
Fixed in #1884.

PR #1884, opened 2019-07-24
Rename gradient type

domoritz:


kanitw:
Fix https://github.com/vega/vega/issues/1807

domoritz:
Should we rename the Vega schema as well then?

kanitw:
Definitely!

In #1807, @jheer seems to be ok with this change (or at least not against it).

domoritz:
I renamed Vega as well.

PR #1887, opened 2019-07-30
ENH: expose aggregate sum value for hierarchy transforms, fixes #1380

### LINKED ISSUE #1380: Treemap - Not able to retrieve aggregate values and parent object
Hi,
I am using treemap with nest transform and in the resulting data i can see aggregate values for each key generated if i check root object. However i am not able to access the same in vega datum object for each node.
Docs state: The sum of values for a node and all its descendants is available on the node object as the value property.
It would be helpful to have a way to get these values in vega if not already there.

I asked the same question in google groups thread with more details and example. Here's the link address:
https://groups.google.com/forum/#!topic/vega-js/jvdqLbYLItQ

tejesh95:
The changes are only from `packages/vega-hierarchy` folders.

I am not sure how `yarn build` generates vega-core and other files, with extra code.


domoritz:
Could you remove the updates to compiled files from your PR?

domoritz:
Thank you!

tejesh95:
@domoritz I added typings. Schema (`vega-schema.json` file) is auto-updated correctly on running `yarn build`.

packages/vega-typings/types/spec/transform.d.ts
```diff
@@ -288,6 +288,55 @@ export interface LookupTransform {
...
+  field?: string | TransformField;
+  sort?: Compare;
+  size: Vector2<number | SignalRef> | SignalRef;
+  radius?: string | TransformField;
+  padding?: number | SignalRef;
+  as?: (string | SignalRef)[] | SignalRef;
+}
+
+export interface PartitionTransform {
+  type: 'pack';
```
domoritz: this should be partition

tejesh95:
Also, Are these failing tests to be looked into? https://travis-ci.org/vega/vega/jobs/565949712#L8835

tejesh95:
5 tests are failing as this is backward-incompatible change. Let me work on an alternate backward compatible spec change and revert in a couple of days.

tejesh95:
should the `tests/spec/valid/*.ts` be git ignored? Because these files are auto built based on `vega/test/specs-valid/*.json`

tejesh95:
will place a new pull request with neat commit history

domoritz:
We will likely squash all commits in a PR anyway, btw.

PR #1899, opened 2019-08-06
add equalEarth to the projections

mattijn:
This PR adds the [equalEarth](https://github.com/d3/d3-geo#equal-earth) projection. The `equalEarth` projection was added in d3-geo [v1.11.0](https://github.com/d3/d3-geo/releases/tag/v1.11.0).

As far as I can see is Vega on `"d3-geo": "^1.11.5"`. The only thing I found is [this](https://github.com/vega/vega/blob/master/yarn.lock#L1600) line in yarn.lock:
```
d3-geo@^1.10.0, d3-geo@^1.11.5:
```

Note: one could argue to also make this the default projection and replace `mercator`.


domoritz:
Could you also update the typings?

mattijn:
Ah forgot that one. Nothing else missing?

jheer:
This is great, thank you!

PR #1900, opened 2019-08-06
ENH: expose aggregate sum value for hierarchy transforms

### LINKED ISSUE #1380: Treemap - Not able to retrieve aggregate values and parent object

tejesh95:
fixes #1380

### DIFF
--- a/docs/docs/transforms/pack.md
+++ b/docs/docs/transforms/pack.md
@@ -21,7 +21,7 @@ Internally, this transform processes a collection of special tree node objects g
 | size                | {% include type t="Number[]" %}| The size of the layout, provided as a [width, height] array.|
 | radius              | {% include type t="Field" %}   | An explicit node radius to use. If `null` (the default), the radius of each leaf circle is derived from the _field_ value.|
 | padding             | {% include type t="Number" %}  | The approximate padding to include between packed circles (default `0`).|
-| as                  | {% include type t="String[]" %}| The output fields at which to write the layout results. The default is `["x", "y", "r", "depth", "children"]`, where `x` and `y` are the layout coordinates, `r` is the node radius, `depth` is the tree depth, and `children` is the count of a node's children in the tree.|
+| as                  | {% include type t="String[]" %}| The output fields at which to write the layout results. The default is `["x", "y", "r", "depth", "value", "children"]`, where `x` and `y` are the layout coordinates, `r` is the node radius, `depth` is the tree depth, `value` is the sum of values for a node, and `value` is the sum of values for a node and all its descendants.|
 
 ## Usage
 
--- a/docs/docs/transforms/partition.md
+++ b/docs/docs/transforms/partition.md
@@ -21,7 +21,7 @@ Internally, this transform processes a collection of special tree node objects g
 | padding             | {% include type t="Number" %}  | The padding to include between adjacent nodes (default `0`).|
 | round               | {% include type t="Boolean" %} | Indicates if node layout values should be rounded (default `false`).|
 | size                | {% include type t="Number[]" %}| The size of the layout, provided as a [width, height] array.|
-| as                  | {% include type t="String[]" %}| The output fields at which to write the layout results. The default is `["x0", "y0", "x1", "y1", "depth", "children"]`, where `x0`, `y0`, `x1` and `y1` are the starting and ending layout coordinates for each segment, `depth` is the tree depth, and `children` is the count of a node's children in the tree.|
+| as                  | {% include type t="String[]" %}| The output fields at which to write the layout results. The default is `["x0", "y0", "x1", "y1", "depth", "value", "children"]`, where `x0`, `y0`, `x1` and `y1` are the starting and ending layout coordinates for each segment, `depth` is the tree depth, `children` is the count of a node's children in the tree, and `value` is the sum of values for a node and all its descendants.|
 
 ## Usage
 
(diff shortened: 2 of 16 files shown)


tejesh95:
could you please review, @domoritz

docs/docs/transforms/pack.md
```diff
@@ -21,7 +21,7 @@ Internally, this transform processes a collection of special tree node objects g
 | size                | {% include type t="Number[]" %}| The size of the layout, provided as a [width, height] array.|
 | radius              | {% include type t="Field" %}   | An explicit node radius to use. If `null` (the default), the radius of each leaf circle is derived from the _field_ value.|
 | padding             | {% include type t="Number" %}  | The approximate padding to include between packed circles (default `0`).|
-| as                  | {% include type t="String[]" %}| The output fields at which to write the layout results. The default is `["x", "y", "r", "depth", "children"]`, where `x` and `y` are the layout coordinates, `r` is the node radius, `depth` is the tree depth, and `children` is the count of a node's children in the tree.|
+| as                  | {% include type t="String[]" %}| The output fields at which to write the layout results. The default is `["x", "y", "r", "depth", "value", "children"]`, where `x` and `y` are the layout coordinates, `r` is the node radius, `depth` is the tree depth, `value` is the sum of values for a node, and `value` is the sum of values for a node and all its descendants.|
```
jheer: Typo. The `value` field is listed twice here and the `children` field is missing.

docs/docs/transforms/partition.md
```diff
@@ -21,7 +21,7 @@ Internally, this transform processes a collection of special tree node objects g
 | padding             | {% include type t="Number" %}  | The padding to include between adjacent nodes (default `0`).|
 | round               | {% include type t="Boolean" %} | Indicates if node layout values should be rounded (default `false`).|
 | size                | {% include type t="Number[]" %}| The size of the layout, provided as a [width, height] array.|
-| as                  | {% include type t="String[]" %}| The output fields at which to write the layout results. The default is `["x0", "y0", "x1", "y1", "depth", "children"]`, where `x0`, `y0`, `x1` and `y1` are the starting and ending layout coordinates for each segment, `depth` is the tree depth, and `children` is the count of a node's children in the tree.|
+| as                  | {% include type t="String[]" %}| The output fields at which to write the layout results. The default is `["x0", "y0", "x1", "y1", "depth", "value", "children"]`, where `x0`, `y0`, `x1` and `y1` are the starting and ending layout coordinates for each segment, `depth` is the tree depth, `children` is the count of a node's children in the tree, and `value` is the sum of values for a node and all its descendants.|
```
jheer: Typo. The order of field descriptions is incorrect. Based on the default list, `value` should come before `children`. However, I think the actual order should be `children`, then the new `value` field, in order to avoid introducing a breaking change.

docs/docs/transforms/tree.md
```diff
@@ -22,7 +22,7 @@ Internally, this transform processes a collection of special tree node objects g
...
 | size                | {% include type t="Number[]" %}| The size of the layout, provided as a [width, height] array.|
 | nodeSize            | {% include type t="Number[]" %}| The size of each node, provided as a [width, height] array.|
-| as                  | {% include type t="String[]" %}| The output fields at which to write the layout results. The default is `["x", "y", "depth", "children"]`, where `x` and `y` are the layout coordinates, `depth` is the tree depth, and `children` is the count of a node's children in the tree.|
+| as                  | {% include type t="String[]" %}| The output fields at which to write the layout results. The default is `["x", "y", "depth", "value", "children"]`, where `x` and `y` are the layout coordinates, `depth` is the tree depth, `children` is the count of a node's children in the tree, and `value` is the sum of values for a node and all its descendants.|
```
jheer: See other comment about avoiding breaking changes. Shouldn't `value` come *after* `children`?

packages/vega-hierarchy/src/HierarchyLayout.js
```diff
@@ -50,10 +50,11 @@ function setParams(layout, params, _) {
 
 function setFields(node, fields, as) {
   var t = node.data;
-  for (var i=0, n=fields.length-1; i<n; ++i) {
+  for (var i=0, n=fields.length-2; i<n; ++i) {
     t[as[i]] = node[fields[i]];
   }
   t[as[n]] = node.children ? node.children.length : 0;
+  t[as[n+1]] = node.value;
```
jheer: OK, it looks like `value` is indeed the last field in the `as` array. That means all the documentation updates have ordering typos.

jheer:
Thanks for working on this PR! Please see my notes for needed fixes.

Also, it would be helpful to document why this change is needed and valuable. In particular, I'd love to see a test specification that demonstrates the feature and its use.

tejesh95:
Currently, sum value is _implicitly_ encoded to relevant hierarchy transform's core mark properties. With `value` being exposed, other mark properties like `fill`, `opacity`, `stroke` etc., can be encoded too.

This change enables the creation of these examples in Vega:

https://gramener.com/fmcg/ 
https://gramener.com/treemap/treemap?app=Nutrition

PR #1909, opened 2019-08-14
Modify's remove predicate should also test for object equality.

arvind:
The `modify` expression function did not previously test for object equality. Our selection logic was unaffected because although tuples frequently had nested objects, these objects were reused. And thus a strict equality test (`===`) was sufficient. 

After making the change to initialize selection values in the dataset rather than via signals (vega/vega-lite#4889, vega/vega-lite#4943), nested objects are now directly embedded in output JSON. Thus, strict equality testing is no longer sufficient.


jheer:
Thanks. This looks fine and we certainly want to be correct! I'm a little nervous about potential performance implications, though, as we end up with super-linear performance (num items * num attributes). I'll think about this a bit more once I have time to start processing the backlog of Vega issues (next week or two).

arvind:
Agreed -- I think this is why we'd left deep object equality out initially. One other idea I had is whether object equality testing is only invoked if the tuple being passed into `toggle` requested it (e.g., had a special `Symbol` or property added?).

jheer:
@arvind: Any updates on your thinking here? It would be nice to get a fix in soon, without churning on it later. Thanks!

arvind:
The best I've come up with is what I have above -- checking for a property on the object being sent in as the remove/toggle predicate. However, I realized that a special `Symbol` would not work as I do not think we have access to the expression language when enumerating dataset values. 

If this approach sounds good to you, I can update this PR. Do we have a property name we like? Perhaps `_vgObjectEqual: true`?

jheer:
Thanks. After closer review, I think merging as-is is preferable to more complicated designs. We should check for possible performance issues downstream and respond if necessary. I did make some minor changes to `equal` to take advantage of an early exit condition.

There is also one thing we should double-check. Previously, our remove predicate directly checked for properties defined on the input object only, which is asymmetric but correct for the remove/toggle semantics. Now we have a recursive setup and for nested objects our "equality" check is also asymmetric. So, data tuples need only match on the sub-properties defined on nested objects provided as input. I think this is OK, but I'm not as close to the issue. @arvind can you confirm that this is the correct behavior? If so, we should be good to squash + merge.

PR #1947, opened 2019-08-22
Fix + expand types for vega-dataflow exports.

arvind:


packages/vega-typings/types/runtime/dataflow.d.ts
```diff
@@ -0,0 +1,28 @@
+export interface Changeset {
+  insert(tuples: any): this;
```
domoritz: Should this be an array?

packages/vega-typings/types/runtime/dataflow.d.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export class Operator {
+  constructor(init?: any, update?: (obj: any, pulse: any) => any, params?: any, react?: boolean);
+  targets(): any;
+  set(value: any): 1 | 0;
+  skip(): (state: any) => any;
+  modified(): (state: any) => any;
+  parameters(params: object, react?: boolean, initonly?: boolean): Operator[];
+  marshall(stamp: number): any;
+  evaluate(pulse: any): any;
```
domoritz: Do all of these need to return any?

packages/vega-typings/types/runtime/dataflow.d.ts
jheer: The input can be an array of tuples, or a single tuple.

packages/vega-typings/types/runtime/dataflow.d.ts
jheer: Some of these methods have multiple return types:
- skip: boolean (if called with no arguments) | Operator (if called with a state argument)
- modified: boolean (if called with no arguments) | Operator (if called with a state argument)
- evaluate: undefined | StopPropagation dummy object

Others have a singular return type:
- targets: UniqueList object
- marshall: Parameters object

jheer:
This looks OK to me, but I'm not the TypeScript czar. I've added some comments regarding return types. Let me know if you plan to update further, or if you think this is ready to merge.

packages/vega-typings/types/runtime/dataflow.d.ts
arvind: Due to time constraints, I'm being opportunistic about updating the typings. Once we flesh out typings for `Tuple` and `Pulse`, I think we can come back and type these to completion.

arvind: Due to time constraints, I'm being opportunistic about updating the typings. Once we flesh out typings for Tuple and Pulse, I think we can come back and type these to completion. I'd advocate for merging these now, so that downstream projects can more immediately reap the benefits.

domoritz: Sounds good to me. 

If you know that there will be some type later, you can always create an alias. 

For example:

```ts
type Tuple = any;
```

PR #1951, opened 2019-08-23
Add View API data replace method.

### LINKED ISSUE #1799: Request for View API method to replace all values in dataset
Often it is necessary to replace all the values in a Vega dataset. To do so in current Vega (v5.3.5) [View API](https://vega.github.io/vega/docs/api/view/) requires the following:
`view.change(datasetName, vega.changeset().remove(function(){return true;}).insert(dataValues)					)`

It would be easier for the user and perhaps faster to execute if there is a method such as view.replace():
`view.replace(datasetName, dataValues)`

or if existing view.data() can be overloaded to set data values:
`view.data(datasetName, dataValues)`

jheer:
**vega-view**
- Add `replace` View API method to replace the contents of a data set.

Fix #1799.

### DIFF
--- a/packages/vega-view/src/View.js
+++ b/packages/vega-view/src/View.js
@@ -1,5 +1,5 @@
 import cursor from './cursor';
-import {data, dataref, change, insert, remove} from './data';
+import {data, dataref, change, insert, remove, replace} from './data';
 import {initializeEventConfig, events} from './events';
 import hover from './hover';
 import finalize from './finalize';
@@ -345,6 +345,7 @@ prototype.data = data;
 prototype.change = change;
 prototype.insert = insert;
 prototype.remove = remove;
+prototype.replace = replace;
 
 // -- SCALES --
 prototype.scale = scale;
--- a/packages/vega-view/src/data.js
+++ b/packages/vega-view/src/data.js
@@ -1,5 +1,5 @@
 import {changeset, isChangeSet} from 'vega-dataflow';
-import {error, hasOwnProperty} from 'vega-util';
+import {error, hasOwnProperty, truthy} from 'vega-util';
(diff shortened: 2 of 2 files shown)


domoritz:
I have mixed feelings about this. First, I think the name `replace` suggests that something is being replaced but that's not necessarily true. How about we name the operation `set` instead? Second, I think a `clear` operation would be both more generic and combined with insert make the semantics clearer than a replace shorthand. 

Also noting that we need to update the typings. 

On a related note, could clearing be made more efficient than propagating a changeset that removes each tuple individually?

jheer:
With respect to naming, `set` also feels too generic for me. An alternative suggestion was to override the `data` method to allow it to function as a setter. That works better for me from a naming perspective, but is misleading in that, as a getter, `data` returns the output data at the end of a pipeline, where as insert/set inject tuples into the beginning of the pipeline.

Thanks also for the reminder about the typings. Duly noted.

As for performance, theoretically I think we could, but not without some significant (though probably straightforward) engineering. We currently need to propagate removed tuples through the dataflow to update the state. To avoid this, we'd need every (stateful) operator to support some kind of fast "reset" capability. That said, we still need to process all the added tuples and the rendering needs wouldn't change, so I'm not sure what performance delta we'd expect to see.

domoritz:
Another possible name would be `assign`.

I like the idea of using `data('foo', tuples)`. I don't think that the difference to getting the dataset at the end of the pipeline is a problem. It wouldn't be possible to set the data as part of a changeset, would it? I suppose one could always use clear and insert in that case.

jheer:
Removed `replace` method, extended the `data` method, and updated typings in #1958.

PR #1957, opened 2019-08-23
Support min/max op for multi-domain sort.

### LINKED ISSUE #1854: Support sorting unioned domains that are aggregated by min and max
Vega does not allow sorting unioned domains by anything other than count (See https://vega.github.io/vega/docs/scales/#domain). A discussion of this issue for Vega-Lite is at https://github.com/vega/vega-lite/issues/5048 and for Vega at https://github.com/vega/vega/issues/1500. However, I think that `min` and `max` aggregates shouldn't be affected by aggregating with wrong multiplicities. There is still the issue that fields may not exist in one dataset but I think it's okay to just ignore the corresponding tuples (which I believe is the default already).

jheer:
**vega-parser**
- Add min/max sort op support for discrete scales with unioned domains.

**vega-schema**
- Update schema for sorting unioned scale domains.

**vega-typings**
- Update typings for sorting unioned scale domains.

Close #1854.

### DIFF
--- a/packages/vega-typings/types/spec/scale.d.ts
+++ b/packages/vega-typings/types/spec/scale.d.ts
@@ -37,12 +37,17 @@ export type SortField =
     };
 
 /**
- * Unioned domains can only be sorted by count aggregate.
+ * Unioned domains can only be sorted by count, min, or max aggregates.
  */
 export type UnionSortField =
   | boolean
   | {
       op: 'count';
+      order?: SortOrder;
+    }
+  | {
+      field: ScaleField;
+      op: 'count' | 'min' | 'max';
       order?: SortOrder;
     };
 export type ScaleField = string | SignalRef;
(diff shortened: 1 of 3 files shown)


packages/vega-typings/types/spec/scale.d.ts
```diff
@@ -37,13 +37,18 @@ export type SortField =
...
  */
 export type UnionSortField =
   | boolean
   | {
       op: 'count';
       order?: SortOrder;
+    }
+  | {
+      field: ScaleField;
+      op: 'count' | 'min' | 'max';
```
domoritz: Count ignores the field, right? Should we remove count then since we have the case above?

jheer: The calculation ignores the field, but specifying a field for count is still legal. If we want to ban that option for multi-domain (but not other cases...) I suppose we could but I don’t see a strong reason to...

PR #1984, opened 2019-09-02
Update flare.json

aradip:


jheer:
Hmm, this PR seems to have been made in error?

In any case, the datasets in the docs folder are drawn from the vega-datasets repo, so any dataset changes directly to this repo will get overwritten anyway!

PR #1988, opened 2019-09-08
update d3-format to 1.4.1

benib:
This PR updates d3-format to v1.4.1

The main reason for this update is to allow a custom minus character used as the prefix for negativ numbers, see https://github.com/d3/d3-format#locales

This allows charts with negative numbers prefixed by the typographically much nicer minus-sign (U+2212) instead of the hyphen-minus which makes people caring about typesetting happy.

Also `nan` is available as a property in the d3 locale to allow the configuration of the string used when d3-format is used to format a value that is `NaN`.

### DIFF
--- a/packages/vega-encode/package.json
+++ b/packages/vega-encode/package.json
@@ -25,7 +25,7 @@
   },
   "dependencies": {
     "d3-array": "^2.3.1",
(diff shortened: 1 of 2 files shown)


jheer:
Thanks!

PR #2002, opened 2019-09-09
Selection fields are flattened, so should be directly accessed.

arvind:
As of vega/vega-lite#3676 and then vega/vega-lite#5351, data fields in Vega-Lite are flattened (i.e., `datum["nested"]["a"]` becomes `datum["nested.a"]`). As a result, using vega-util's `field` accessor will yield the wrong behavior and we, instead, want to directly access fields on the datum.


arvind:
Closing this in favor of emitting escaped flattened field names in Vega-Lite, to keep things consistent in the VL compiler.

Issue #1860, opened 2019-07-03
Feature Suggestion: color contrast check


benib:
# goal
It is possible to check the contrast value of colors to make sure labels are rendered with a good constrast ratio. This would improve accessibility of produced charts.

# reason to add this to vega
color/fill properties can be computed at runtime. Signals, other expressions, data, can influence the color. To ensure a minimum contrast ratio it is helpful to have the ability of applying different text colors depending on expression values so this can be defined in production rules of the marks.

This would allow to render labels in a chart like this either black or white, depending on the contrast with the color of the bars.
<img width="325" alt="Screenshot 2019-07-03 at 13 28 01" src="https://user-images.githubusercontent.com/821875/60588133-6bc15b80-9d96-11e9-85c0-394bd4637fd9.png">

# possible solutions
1. add a dependency to https://github.com/gka/chroma.js to vega-functions and use it's contrast function (with the possibility to expose more of chroma.js API if there is a need)
2. implement lumincance/contrast functions directly in vega-functions based on https://en.wikipedia.org/wiki/Rec._709 by converting the colors to rgb using d3-color and then computing the luminance using the coefficients described in this standard.

# example of usage
```
fill: [
  {
    "test": "contrast('black', scale('colorScale', datum.value)) > 4.5",
    "value": "black"
  },
  {
    "value": "white"
  }
},
```

I am happy to work on either solution and provide a PR if you think the addition of luminance/contrast functions are useful and something you would merge into vega.

domoritz:
You can actually prototype this by [adding expression functions at runtime](https://vega.github.io/vega/docs/api/extensibility/#expressions). I do think that a simple contrast formula would be nice. https://danielleszafir.com/ has done some amazing work on this issue recently and I think we should build on it if we want to add a formula.

benib:
I am doing this by adding an expression function according to solution 2 just now. It works great so far. I definitely have to check out Danielles work on this and would love to have a solution to this in core vega instead of depending on runtime added expression functions.

domoritz:
@connorgr has written some JS libraries for color comparison based on her work that we could leverage.

jheer:
Thanks! If there is a useful -- and ideally relatively simple! -- function to add, I'd be happy to review it. I'm wary of adding new third-party dependencies that might bloat the (already large) Vega footprint.

As @domoritz mentions, one can always add functionality to Vega for custom deployments using the extensibility API. This would be my preference as a default starting point, as ultimately we can't add every extension one might request.

[...]

PR #2004, opened 2019-09-15
Add luminance and contrast functions

benib:
# Implements luminance and contrast expression functions

This PR adds 2 new functions. It was discussed in #1860. The implementation is pretty small (20 LOC). Adding this to vega allows for more accessible charts and would help to build awareness about accessibility in my opinion.

- `luminance` returns the relative luminance according to https://www.w3.org/TR/2008/REC-WCAG20-20081211/#relativeluminancedef for a given color
- `contrast` returns the contrast ratio between two colors according to https://www.w3.org/TR/2008/REC-WCAG20-20081211/#contrast-ratiodef
- tests are added for both functions in vega-functions

This PR extends the grouped-bar example and test to use the contrast function to determine if the text mark should use black or white fill, depending on the higher contrast ratio against the rect color. The scenegraph of the grouped-bar test is changed to reflect the new text fill values.

If you don't want to add this to vega in ordner to not add more bloat to the package I would totally understand that.

Let me know if there is anything that I missed, did wrong or should change in this PR.

### DIFF
--- a/packages/vega-functions/src/codegen.js
+++ b/packages/vega-functions/src/codegen.js
@@ -64,6 +64,11 @@ import {
 } from 'd3-color';
 
 import {
+  luminance,
+  contrast
+} from './luminance';
+
+import {
   data,
   indata,
   setdata
@@ -185,6 +190,8 @@ export const functionContext = {
   lab,
   hcl,
   hsl,
+  luminance,
+  contrast,
   sequence,
   format,
   utcFormat,
--- a/packages/vega-functions/src/luminance.js
+++ b/packages/vega-functions/src/luminance.js
@@ -1,16 +1,19 @@
+import { rgb } from "d3-color";
+
 // https://www.w3.org/TR/2008/REC-WCAG20-20081211/#relativeluminancedef
 function channel_luminance_value(channelValue) {
   const val = channelValue / 255;
   if (val <= 0.03928) {
-    return val / 12.92
+    return val / 12.92;
   }
   return Math.pow((val + 0.055) / 1.055, 2.4);
 }
 
 export function luminance(color) {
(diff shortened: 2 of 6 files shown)


domoritz:
I love this addition and the example.

benib:
@jheer: Thanks for the suggestions. `luminance` takes a string now and calls `rgb` internally.

`/docs` changes are removed. Here they are for pasting later:

```
<a name="luminance" href="#luminance">#</a>
<b>luminance</b>(color)<br/>
Returns the luminance for the given color, an object created using the rgb function (see above). The luminance is calculated according to https://www.w3.org/TR/2008/REC-WCAG20-20081211/#relativeluminancedef.

<a name="contrast" href="#contrast">#</a>
<b>contrast</b>(color1, color2)<br/>
Returns the contrast ratio as a float between 1 and 21. color1 and color2 are objects created using the rbg function (see above). The contrast is calculated according to https://www.w3.org/TR/2008/REC-WCAG20-20081211/#contrast-ratiodef.
```

`/docs/examples/grouped-bar-chart.vg.json`: (the text marks)
```
{
  "type": "text",
  "from": {"data": "bars"},
  "encode": {
    "enter": {
      "x": {"field": "x2", "offset": -5},
      "y": {"field": "y", "offset": {"field": "height", "mult": 0.5}},
      "fill": [
        {"test": "contrast(rgb('white'), rgb(datum.fill)) > contrast(rgb('black'), rgb(datum.fill))", "value": "white"},
        {"value": "black"}
      ],
      "align": {"value": "right"},
      "baseline": {"value": "middle"},
      "text": {"field": "datum.value"}
    }
  }
}
```

jheer:
Thanks!

PR #2006, opened 2019-09-16
Docker setup

domoritz:
With this setup
* everyone can develop Vega regardless of their operating system
* we can run tests on Linux -> Travis is faster


domoritz:
#2007 is cleaner

PR #2007, opened 2019-09-16
Docker compose setup

domoritz:
Better alternative to https://github.com/vega/vega/pull/2006

With this setup

* everyone can develop Vega regardless of their operating system
* you don't need to install any specific dependencies to develop Vega (only Docker)
* reproducible dev environment: we use the same environment on Travis and locally
* we can run tests on Linux -> Travis is way faster (10mins -> 3mins)
* we can easily switch to another testing environment like GitHub actions or CircleCI if we ever want to
* we can make tests even faster by caching the whole docker image (not done yet since it increases the complexity a bit)


jheer:
Thank you for working on this! This provides a nice solution to our testing issues, but I'm wary of bringing in all the extra complexity/hassle of installing and working with Docker.

As an alternative, I took a closer look at what is causing our Linux and MacOS tests to diverge. The difference lies entirely in text rendering. I tried playing with different configurations for node-canvas text rendering and antialiasing, but cross-platform differences are still observed. However, if I instead stub out the text mark canvas renderer to draw text bounding boxes instead of text glyphs, we can achieve identical outputs across platforms. As the total resulting churn is three output images, I'm comfortable making this change and moving back to Linux based CI runs, without need for a container infrastructure. See #2029.

jheer:
Closing in favor of the approach in #2029 / #2022.

PR #2027, opened 2019-09-26
Update geoshape.md

afdta:
I was confused to see `projection` as a required argument. It seems as though it is acceptable to omit it in which case the null transform (identity function) would be used to map coordinates. The same is true for the geopath transform, where the contour plot example omits it: https://vega.github.io/vega/examples/contour-plot/


jheer:
You are correct. Thanks! I will update GeoPath to match.

PR #2047, opened 2019-10-03
add mergeDeep and mergeConfig util functions

### LINKED ISSUE #2045: Inconsistent merge logic for configs
Configs are often used as themes. In a specific Vega spec, one might want to further customize the config and override specific properties. We use this functionality in Vega-Lite, the editor and Vega-Embed. However, with the new [signals in the config](https://vega.github.io/vega/docs/config/#signals) it becomes unclear how to correctly merge configs.

Specifically, when merging two arrays there are two options. 1. overwrite the array and 2. union the arrays. We think that 1 is what a user wants for all properties except the signals. Signals should behave more like a map than an array.

I'm going to send a PR to add a function to merge configs to Vega-Util. 

cc @kanitw @arvind @jheer

domoritz:
Fixes #2045

### DIFF
--- a/packages/vega-util/src/mergeDeep.js
+++ b/packages/vega-util/src/mergeDeep.js
@@ -1,0 +1,33 @@
+import isArray from "./isArray";
+import isObject from "./isObject";
+import hasOwnProperty from "./hasOwnProperty";
+
+export default function mergeDeep(dest, ...src) {
+  for (const s of src) {
+    dest = mergeDeep_(dest, s);
+  }
+  return dest;
+}
+
+function mergeDeep_(dest, src) {
+  if (!isObject(src) || src === null) {
+    return dest;
+  }
+
+  for (const p in src) {
+    if (!hasOwnProperty(src, p)) {
+      continue;
+    }
+    if (src[p] === undefined) {
+      continue;
+    }
+    if (!isObject(src[p]) || isArray(src[p]) || src[p] === null) {
+      dest[p] = src[p];
+    } else if (!isObject(dest[p]) || dest[p] === null) {
+      dest[p] = mergeDeep(isArray(src[p].constructor) ? [] : {}, src[p]);
+    } else {
+      mergeDeep(dest[p], src[p]);
+    }
+  }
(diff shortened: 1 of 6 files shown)


packages/vega-util/src/mergeDeep.js
```diff
@@ -0,0 +1,33 @@
...
+
+export default function mergeDeep(dest, ...src) {
+  for (const s of src) {
+    dest = mergeDeep_(dest, s);
+  }
+  return dest;
+}
+
+function mergeDeep_(dest, src) {
+  if (!isObject(src) || src === null) {
```
jheer: Why are there strict equality checks for null? The vega-util `isObject(null)` returns `false`, so I think we can remove the extra check. Also on lines 24 and 26.

packages/vega-util/src/mergeDeep.js
```diff
@@ -0,0 +1,33 @@
...
+    if (!hasOwnProperty(src, p)) {
+      continue;
+    }
+    if (src[p] === undefined) {
+      continue;
+    }
+    if (!isObject(src[p]) || isArray(src[p]) || src[p] === null) {
+      dest[p] = src[p];
+    } else if (!isObject(dest[p]) || dest[p] === null) {
+      dest[p] = mergeDeep(isArray(src[p].constructor) ? [] : {}, src[p]);
```
jheer: 1. Can't you call `mergeDeep_` (with underscore) here instead?
2. Why check the constructor instead of just `src[p]`?
3. Wouldn't `null` (rather than `[]`) work, and remove one unnecessary object instantiation.

packages/vega-util/src/mergeDeep.js
```diff
@@ -0,0 +1,33 @@
...
+    } else {
+      mergeDeep(dest[p], src[p]);
```
jheer: Only two arguments, so why not call `mergeDeep_` (with underscore) instead?

jheer:
See #2051 for an alternative PR that moves the parser's merge logic to vega-util.

PR #2051, opened 2019-10-04
Add mergeConfig utility method.

### LINKED ISSUE #2045: Inconsistent merge logic for configs

jheer:
**vega-parser**
- Use `mergeConfig` utility to combine input config objects to parser.

**vega-util**
- Add `mergeConfig` utility function.

Close #2045.
Candidate to supercede and close #2047.

### DIFF
--- a/packages/vega-parser/package.json
+++ b/packages/vega-parser/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vega-parser",
-  "version": "5.9.0",
+  "version": "5.10.0",
   "description": "Parse Vega specifications to runtime dataflows.",
   "keywords": [
     "vega",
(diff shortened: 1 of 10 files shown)


domoritz:
Looks great. Thank you for the update!

I wonder whether you could expose the copy method as well since it could replace our mergeDeep function in Vega-Lite.

PR #2075, opened 2019-10-09
Simplify guide types by removing generics

domoritz:
**vega-typings**

* Remove generics from guide types.


domoritz:
Merge after https://github.com/vega/vega-lite/pull/5000 is approved.

Issue #1904, opened 2019-08-08
Support custom color name in the config 


kanitw:
We should be able to define:

```
config: {
  color: {
    "uw-purple": "#XXXXXX"
  }
}
```

and then use `"fill": {"value": "uw-purple"}` in the code. 

This will be especially useful to create a visualization that works across themes.  For example, one can name a color by role (such as "highlight") and then set the "highlight" color to be different color for different themes. 

cc: @Saba9

domoritz:
Can't we use styles for this?

kanitw:
> Can't we use styles for this?

Not for conditional encoding.

domoritz:
Good point.

In general, I like to solve a single problem in a way that it solves other problems at the same time. It would be great not to add an ad-hoc solution for one problem.

In this case, I think named colors are a necessary extension given that CSS also introduced them despite the availability of styles.

jheer:
Should color variables references be simple strings, or should they have a special prefix? That is, do you use `{"value": "uw-purple"}` or something like `{"value": "@uw-purple"}`. I ask because otherwise you can cause some major mischief by overriding existing colors. Consider:

```json
"config": {
  "colors": {
    "#000": "#f00",
    "#0000ff": "#00ff00",
    "black": "white",
    "white": "black"
  }
}
```

Also, if you allow lookup to be recursive (variables can be defined by pointing to another variable), you can more easily end up with accidental infinite loops if the syntax doesn't distinguish between CSS color strings and Vega color variable names.

So, we need to further flesh out any designs here. Thoughts?

domoritz:
We could use the format that css variables use https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties.  Or the $ prefix as in scss https://sass-lang.com/documentation/variables.

jheer:
So I thought about it some more, and I'm becoming increasingly uncomfortable with this request, as formulated. If we are going to add variables, what are all the different places in Vega in which they are usable? And why only for colors? Maybe someone wants to do this for shapes, or stroke dash patterns, or, ...

Quickly this expands o a global variable system. Fortunately, we already have a global, reactive variable system: signals. So, my counter-proposal would be to add a new `signals` entry to the config that allows one to define a set of signals in a reusable fashion. These would then be processed at parse time and otherwise act as normal signals.

[...]

Issue #666, opened 2016-11-30
png creation problem in v3.0.0-beta*


alkan:
While running

node_modules/vega/bin/vg2png node_modules/vega/spec/area.vg.json

command  we get the following error:

TypeError: Cannot read property 'getContext' of null
    at resize (/Users/ayhan/projects/matriks/dev/social-net/botanic-charts/node_modules/vega/build/vega.js:10800:23)
    at CanvasRenderer.prototype$38.resize (/Users/ayhan/projects/matriks/dev/social-net/botanic-charts/node_modules/vega/build/vega.js:10841:3)
    at CanvasRenderer.prototype$36.initialize (/Users/ayhan/projects/matriks/dev/social-net/botanic-charts/node_modules/vega/build/vega.js:10439:15)
    at CanvasRenderer.prototype$38.initialize (/Users/ayhan/projects/matriks/dev/social-net/botanic-charts/node_modules/vega/build/vega.js:10836:26)
    at initializeRenderer (/Users/ayhan/projects/matriks/dev/social-net/botanic-charts/node_modules/vega/build/vega.js:23846:6)
    at /Users/ayhan/projects/matriks/dev/social-net/botanic-charts/node_modules/vega/build/vega.js:23913:12

jheer:
Just to sanity check, was node-canvas (an optional dependency for Vega overall, but required for vg2png) successfully installed?

jheer:
Using a clean install of Vega (via `npm install vega` on Mac OS and Node v6.9), I am able to run `node_modules/vega/bin/vg2png node_modules/vega/spec/area.vg.json` without issues.

I'm closing the issue for now. I suspect any issues here are most likely due to the node-canvas module. If so, please see the [node-canvas repository](https://github.com/Automattic/node-canvas/tree/v1.x) for more information and troubleshooting tips. Feel free to respond / re-open with more information if needed!

nwhitmont:
I am experiencing this error with the following code adapted from the example [here](https://vega.github.io/vega/usage/#node).

Note: using the vg2png command-line utility works fine. But this Node.js code to write to PNG file fails with error.

```js
var vega = require('vega');
var fs = require('fs');

// create a new view instance for a given Vega JSON spec
var view = new vega.View(vega.parse(spec))
  .renderer('none')
  .initialize();

  // generate a static PNG image
view.toCanvas()
  .then(function(canvas) {
    // process node-canvas instance
    fs.writeFile('chart.png', canvas.toBuffer())
  })
[...]

Issue #444, opened 2015-11-21
Sandbox problem


RussellSprouts:
Here's a way to execute arbitrary code -- when using a group scope for a scale lookup, we don't check that the group is actually a group before calling its scale method and using the scale it returns. It should be pretty easy to fix -- create a whitelist of group objects that are valid arguments for the scope.

I looked through the rest of the code that gets the values of a signal, and I couldn't find any other similar issues.

```json
{
  "signals": [
    {
      "name": "signal1",
      "streams": [
        {
          "type": "window:mousemove",
          "expr": "{scale:('').constructor.constructor}"
        }
      ]
    },
    {
      "name": "xss",
      "streams": [
        {
          "type": "signal1",
          "expr": "'xss!'",
          "scale": {"name": "alert(arguments[0])", "scope": "signal1"}
        }
      ]
    }
  ]
}
```

arvind:
Fixed with bd08614. Nice catch!

RussellSprouts:
Unfortunately, the fix doesn't quite work.

```json
{
  "signals": [
    {
      "name": "signal1",
      "streams": [
        {
          "type":"@rectangle:mousemove",
          "expr": "{__proto__:eventGroup(), scale: ('').constructor.constructor}",
          "scale": {"name": "alert('xss')", "scope": {"signal": "signal1"}}
        }
      ]
    }
  ],
  "marks": [
    {
      "type": "group",
      "marks": [{
        "name": "rectangle",
        "type": "rect",
        "properties": {
          "enter": {
            "fill": {"value": "red"},
            "width": {"value": 100},
            "height": {"value": 100}
          }
        }
      }]
    }  
  ]
}
```

Hover the red square to trigger it.

PR #2088, opened 2019-10-16
System colors as signals

domoritz:
This PR changes all default colors to be signals from a predefined set of colors. This change makes it easier to style Vega specifications. We encourage users to use the predefined colors in their specifications rather than hard-coding colors. 

Addresses https://github.com/vega/vega/issues/1904

Depends on https://github.com/vega/vega/pull/2121


arvind:
I think I’m missing some context here. What is the background/motivation for this change? In particular, it’d be helpful to know what is currently difficult about styling/theming that this change would address. Thanks!

domoritz:
Currently, it is difficult to create for example a dark theme because a lot of specs hard code colors. Take for example https://vega.github.io/editor/#/examples/vega-lite/geo_layer_line_londo, https://vega.github.io/editor/#/examples/vega-lite/geo_text, or https://vega.github.io/editor/#/examples/vega-lite/bar_layered_weather which don't work well on dark backgrounds. The issue is that there isn't a standard set of colors that one could use when writing a spec and that can be modified in a theme. In a dark theme, we could, for example, invert the white, light gray, gray, dark gray, black colors and lighten the other colors.

With this change, you could do something similar to theming in terminals. Here is for example what my colors look like in iTerm2.

Screen Shot 2019-10-15 at 20 03 37

domoritz:
Thank you for the comments. To clarify, this PR is here to address the issues raised in https://github.com/vega/vega/issues/1904. Many specs use custom, hard-coded colors that make it impossible to apply a theme. I don't think we need variables for the default mark color or text color (these can be configured) but we do need a set of system colors that people can use and that a theme can customize. Having variables such as the default font could still be useful to make it easier to write configs but it's something one can work around; hardcoded colors are the problem I'm trying to address with this PR. 

The inspiration for this are the colors used in terminal emulators as shown below and system colors as defined by Apple (https://developer.apple.com/documentation/uikit/uicolor/standard_colors). https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/color/ shows how these colors may change in a dark mode. 

I initially chose to just use color names as signals as custom signals outside the spec would override the config signals and any signal that is a color name would unlikely be used for anything other than a color. However, I see that it makes sense to namespace the colors, which I did in a follow-up commit.

I agree that specific color names for black, white, and gray that may be inverted are confusing so I changed to 10 gray color stops. By default, these colors are uniformly sampled from LAB interpolation between white and black.

Screen Shot 2019-10-16 at 23 18 58

packages/vega-parser/src/config.js
```diff
@@ -1,10 +1,20 @@
...
+const tab10 = scheme('tableau10');
+const defaultFont = 'sans-serif',
     defaultSymbolSize = 30,
     defaultStrokeWidth = 2,
-    defaultColor = '#4c78a8',
-    black = '#000',
-    gray = '#888',
-    lightGray = '#ddd';
+    blue = {signal: 'colors.blue'},
+    lightgray = {signal: 'colors.gray8'},
```
kanitw: It's worth noting that lightgray was `#ddd`, which is equivalent to `rgb(221,221,221)`.
Changing to `gray8`, which is rgb(226, 226, 226), would be a breaking change as all Vega plots would look different.

To avoid such breaking changes, one potential compromise is to do 8 steps (#000, #222, #444, #666, ..., #fff), which will include both #888 and #ddd that we previous used.  

We could consider 16 steps as well, but I think 10 is already too many, so I'd prefer 8. 

Since we're not using these grays to encode quantitative values, it might be less important to make them uniformly sampled from LAB.

packages/vega-parser/src/config.js
```diff
@@ -216,6 +226,32 @@ export default function() {
         'triangle-down',
         'triangle-left'
       ]
-    }
+    },
+
+    signals: [{
+      name: 'colors',
+      value: {
+        blue: tab10[0],
```
kanitw: uses these colors in the default `categorical` range as well?

packages/vega-parser/src/config.js
domoritz: I’m okay with this. The reason for 11 stops was to have them represent percentages. I can change to your suggestions if @jheer agrees with it.

domoritz: Which stops do you want @kanitw?

`#000, #222, #444, #666, #888, #aaa, #ccc, #ddd, #fff` (#ccc and #ddd are closer than the others).

kanitw: Right, I'm a bit dumb here.  There are actually only 15 steps from 0, 1, 2, ... , E, F.  

Hmm

packages/vega-parser/src/config.js
domoritz: Done

packages/vega-parser/src/config.js
domoritz: That's what I did in the last few commits. It's a lot of colors but at least we have all the stops we have used so far.

domoritz:
@arvind do the explanations make sense? It would be really great to get better theming support in Vega.

packages/vega/test/scenegraphs/weather.json
```diff
@@ -25,23 +25,23 @@
                       "x": 0,
                       "y": 182,
                       "opacity": 1,
-                      "stroke": "white",
+                      "stroke": "#ddd",
```
domoritz: This should not change.

domoritz:
While the tests are passing now, the behavior seems incorrect. I created an issue to demonstrate the bug in https://github.com/vega/vega/pull/2101.

kanitw:
There is no immediate need for this feature in Vega-Lite as we added our own config.

PR #2099, opened 2019-10-24
Add TimeUnit support.

### LINKED ISSUE #1764: Support imputation of missing time units
Right now line and area marks over a temporal domain will interpolate over missing time units (such as year-month values). Treating the domain as ordinal results in empty bins being dropped. We need an imputation method that "knows" about time units and so can fill in missing values with dummy (e.g., `defined: false`) records to support proper interpolation and ordination.

### LINKED ISSUE #2084: Better handling of time ticks when domain falls between "nice" points
If you have an axis that uses the default "time" format, and have an domain that falls between two "nice" points in time, like e.g. a domain between 10:55:01 and 10:55:59 the time ticks will not include the "nice" points and therefor labels don't contain enough information for the reader to determine what range she is looking at (there is no point of reference):

Here is an example:

image

As you can see the labels include the SECOND component in all ticks, but no information about the hour or minute of the day. Which makes it impossible to reference the time ticks.

Here is an example where it includes on of the nice ticks that can be used for reference:

image

The point is, that since we have so much screen space in this example, there is seaming no reason that Vega could have much more verbose/specific labels in these situations, or ensure that there is at least one tick that contains enough context.  

Alternatively, allow defining which time components should be included at different interval scales without having to defined a complicated custom label encoding.

### LINKED ISSUE #2089: Why dependencies version are not fixed ?
Hi, today I ran into the following problem : 

I'm working on a project that's using Vega 5.4.0 and TypeScript 3.3.
Everything used to be fine until today when the build of this project failed.
But nothing had changed, at least on my side.

After investigating I found out that dependencies of the Vega package are not fixed.
Vega-util (a sub-dependency to my perspective) was updated from 1.10.0 to 1.11.2 and between those two minor version the minimum Typescript support went from 2.7 (fine for me) to 3.4 (my project can't build) https://github.com/vega/vega/commit/da777eb906c370438f2f476a9cadd18b8e0c2667.

I don't think this behavior is intended :upside_down_face: 
Good practice suggest not using dependencies ranges (with `*~^<>`) but fixing dependencies version to avoid this type of issues.

May relate to #1259.

What are your though on that ?

Best regards.

Augustin :fr:

jheer:
**vega**
- Add vega-time utilities.
- Add `bar-time` test specification.
- Fix dependencies to specific versions. (#2089)

**vega-encode**
- Add support for UTC formatType.
- Update time/utc scale formatting to use vega-time methods.

**vega-functions**
- Add `timeUnitSpecifier` expression function.
- Add `timeOffset` and `utcOffset` expression functions.
- Add `timeSequence` and `utcSequence` expression functions.
- Update `timeFormat` and `utcFormat` functions to use vega-time methods.
- Update dependencies.

**vega-parser**
- Update dependencies.

**vega-regression**
- Update dependencies.

**vega-scale**
- Update to use vega-time, drop internal time interval method.

**vega-schema**
- Add UTC timeFormat to schema.
- Update axis, legend format schema.

**vega-statistics**
- Update dependencies.

**vega-time**
- Add new vega-time utility package.

**vega-transforms**
- Add TimeUnit transform.

**vega-typings**
- Add TimeUnit transform typings.
- Add UTC timeFormat typing.
- Update axis, legend format typings.

Close #1764.
Close #2084.
Close #2089.


domoritz:
:heart:

gordonwoodhull:
Awesome. The floor methods here should improve the legibility of my vega code for snapping to time intervals.

I guess this will also pave the way to ceil and round for time intervals?

anagrius:
Really cool

PR #2101, opened 2019-10-28
bug demo: try using colors from signals.

domoritz:
All I changes was using a signal instead of a specific value in the config. According to https://vega.github.io/vega/docs/config/#signals, I should be able to use signals in the config. However, the scenes change in unexpected ways when I do this.

### DIFF
--- a/packages/vega/test/scenegraphs/budget-forecasts.json
+++ b/packages/vega/test/scenegraphs/budget-forecasts.json
@@ -25,7 +25,7 @@
                       "x": 34,
                       "y": 0,
                       "opacity": 1,
-                      "stroke": "white",
+                      "stroke": "#ddd",
                       "strokeOpacity": 0.75,
                       "strokeWidth": 1,
                       "y2": -400
@@ -34,7 +34,7 @@
                       "x": 102,
                       "y": 0,
                       "opacity": 1,
-                      "stroke": "white",
+                      "stroke": "#ddd",
                       "strokeOpacity": 0.75,
                       "strokeWidth": 1,
                       "y2": -400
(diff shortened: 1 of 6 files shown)


packages/vega/test/scenegraphs/budget-forecasts.json
```diff
@@ -25,7 +25,7 @@
                       "x": 34,
                       "y": 0,
                       "opacity": 1,
-                      "stroke": "white",
+                      "stroke": "#ddd",
```
domoritz: The color should not change.

jheer:
Here is what is happening... The examples that are changing each involve a custom axis or legend color setting, that gets mapped to the `enter` set of an encoding block. The signal variables, on the other hand, are being written to the `update` set, as they might dynamically change, hence the overriding. This stems from the behavior of the [`addEncode`](https://github.com/vega/vega/blob/master/packages/vega-parser/src/parsers/encode/encode-util.js#L9) method in vega-parser's encode-util.js.

If I update the `budget-forecasts`, `jobs`, and `weather` examples to use top-level configuration parameters instead of explicit encode blocks targeting the `enter` set, the errors go away. I think this is a reasonable fix... If you target the `enter` set explicitly, Vega should of course respect that. If you use signals for configuration variables, it also seems reasonable that these should support reactive updates by default, requiring the use of the `update` set. The downside is that legacy code that uses custom encoding blocks that target the `enter` set would no longer work as intended if signal-based defaults are introduced.

The case of the `font-size-steps` example is more involved, where an explicit setting of `symbolStrokeColor` in the spec is getting overridden by the `symbolBaseStrokeColor` config value. These two entries are currently processed separately, such that the base config value is first written to the `update` block and the the explicit setting is then written to an `enter` block. This is a bug, regardless of the other aspects of this issue. I can easily fix the legend generation code in vega-parser to consider all these values in tandem.

Let me know if you have any thoughts / comments on these proposed changes.

domoritz:
Ahh, that makes sense. Thank you for the explanation and fix!

PR #2117, opened 2019-11-01
Support different radius for different rect corners

### LINKED ISSUE #1768: Corner radius should be less than mark span
When the `cornerRadius` of a bar exceeds half the span from the scaled data, effects like this result:

Exported

This is odd-looking. The `cornerRadius` should be pegged at runtime to be one half the span of the mark so the mark's corner curve doesn't invert.

The image above was generated from the following vega-lite spec:

```json
{
  "$schema" : "https://vega.github.io/schema/vega-lite/v3.json",
  "autosize" : {
    "contains" : "padding",
    "resize" : true,
    "type" : "fit"
  },
  "data" : {
    "url" : "data/seattle-weather.csv"
  },
  "encoding" : {
    "color" : {
      "field" : "weather",
      "legend" : {
        "title" : "Weather type"
      },
      "scale" : {
        "domain" : [
          "sun",
          "fog",
          "drizzle",
          "rain",
          "snow"
        ],
        "range" : [
          "#e7ba52",
          "#c7c7c7",
          "#aec7e8",
          "#1f77b4",
          "#9467bd"
        ]
      },
      "type" : "nominal"
    },
    "x" : {
      "axis" : {
        "title" : "Month of the year"
      },
      "field" : "date",
      "timeUnit" : "month",
      "type" : "ordinal"
    },
    "y" : {
      "aggregate" : "count",
      "type" : "quantitative"
    }
  },
  "height" : 449.5,
  "mark" : {
    "cornerRadius" : 24,
    "size" : 58,
    "type" : "bar"
  },
  "width" : 937
}
```

### LINKED ISSUE #1968: Support different radius for different rect corners
Currently, we have the `cornerRadius` property, which controls the radius for all corners of the rect marks. 

image

However, designers often prefer that bars have rounded corners only for the top of the bars:

image

image

For syntax, we could name the new encoding channels in one of the following ways:

1) `(topLeft/topRight/bottomLeft/bottomRight)CornerRadius`
2) `(top/bottom/right/left)CornerRadius`

I think 1) would be more flexible, but 2 would better support the described use case.  (We would only need to use `topCornerRadius` to produce the plots above.)

donghaoren:
Implement @kanitw's syntax design for corner radius on different corners. Now in addition to

```ts
cornerRadius: number;
```

we have

```ts
cornerRadius: {
    all?: number;
    left?: number;
    right?: number;
    top?: number;
    bottom?: number;
    topLeft?: number;
    topRight?: number;
    bottomLeft?: number;
    bottomRight?: number;
}
```

Priority is all < left/right < top/bottom < topLeft/topRight/bottomLeft/bottomRight.

Things I'm not sure:

- Should we commit files in the `docs` folder? I don't understand some changes in the compiled js files.
- Can we assume `w >= 0 && h >= 0` in `rectangle.js`, that would simplify the code?
- We can map data to the numerical corner radius, which applies to all corners. Is there a way to map data to individual corners like `cornerRadius.topLeft`?

Suggestion:
- Cubic curves are not good approximations to arcs, for large cornerRadius values, it doesn't look correct. We may consider using Bezier curves. Here is a good approximation: <http://spencermortensen.com/articles/bezier-circle/>.

#1768 is fixed along the way.

Fix #1768
Fix #1968


jheer:
Thanks for the PR! A few thoughts:

1. Please keep web documentation updates separate. Vega's docs folder is "live", such that any changes are directly published to the web. Documentation updates can be provided in a separate PR or as comments on this PR. Similarly, PRs should not include new build output copied to the docs folder.

2. I'm uncomfortable using an union-typed value for cornerRadius, encompassing both a number and a (somewhat complicated) object type. I would prefer a simpler scheme that keeps the current cornerRadius property unchanged, but adds four new flat properties (`cornerRadiusTopLeft`, `cornerRadiusTopRight`, `cornerRadiusBottomLeft`, `cornerRadiusBottomRight`). The existing `cornerRadius` would set the base radius value, which the more specific entries could then override.

3. Following the scheme outlined in item 2 above, I think the rectangle path implementation might be simplified to accept either 1 or 4 corner accessors as input. We could then instead modify [shapes.js](https://github.com/vega/vega/blob/master/packages/vega-scenegraph/src/path/shapes.js#L36) to do something like this:

```js
function value(a, b) { return a != null ? a : b; }

function crTL(item) { return value(item.cornerRadiusTopLeft, item.cornerRadius) || 0; }
function crTR(item) { return value(item.cornerRadiusTopRight, item.cornerRadius) || 0; }
function crBL(item) { return value(item.cornerRadiusBottomLeft, item.cornerRadius) || 0; }
function crBR(item) { return value(item.cornerRadiusBottomRight, item.cornerRadius) || 0; }

// ...

var rectShape = vg_rect().x(x).y(y).width(w).height(h)
                         .cornerRadius(crTL, crTR, crBL, crBR),
```

Alternatively, to avoid ambiguity with the argument order, the rectangle shape cornerRadius setter could accept either a single function or an object with accessor values (`{topLeft: crTL, ...}`).

jheer:
Update: after making the implementation changes described above, you can keep the documentation edits in this PR if you edit it to target the v5.8.0 branch instead. That said, please do not include the new build files (docs/vega.js, etc) in the updated PR. Thanks again!

kanitw:
> I'm uncomfortable using an union-typed value for cornerRadius, encompassing both a number and a (somewhat complicated) object type. I would prefer a simpler scheme that keeps the current cornerRadius property unchanged, but adds four new flat properties (cornerRadiusTopLeft, cornerRadiusTopRight, cornerRadiusBottomLeft, cornerRadiusBottomRight). The existing cornerRadius would set the base radius value, which the more specific entries could then override.

We'd be curious to hear why you prefer flat properties.

We think nesting is probably better for a few reasons:

1) AFAIK, I don't remember we use flat properties to override sub-parts of a property elsewhere in Vega. (I could be wrong, but it'd be great to see a counterexample if you have one.)

From what I see, we actually provide object forms for such properties:

- The top-level padding property accepts an object in the format: `{"left": 5, "top": 5, "right": 5, "bottom": 5}`. 

- [ColorValue](https://vega.github.io/vega/docs/types/#ColorValue) types also breaks fill/stroke into nested subparts (e.g., "r", "g", "b") -- though slightly different with additional "color" nesting. 

2) In practice, we figure that people would use either the fix value for all corners or setting both of the top corners like in the image below (or, similarly, "right" corners for horizontal bars):

image

Thus, we figure that `top` would be more useful than `topRight` + `topLeft`.  

(I can't even think when would users would like to change topRight alone without either `topLeft` or `bottomRight`.)

jheer:
You asked for examples of inter-property dependencies:
- `x`/`x2`/`xc`/`width` interact with each other, and if one tries to specify more than one, a system of precedence rules takes over. For example, `x2` takes precedence over width, and one can use `x`/`x2`/`xc` in conjunction: width = x2 - x, then x is recentered to xc.
- The `orient` property on area changes which spatial properties are used.

While `padding` supports an object property, it is a top-level property and not part of an encoding channel. On the other hand, no encoding channels (except for `text`) use union types (i.e., in terms of the values written to the scenegraph):
- Colors (`fill`, `stroke`) are a special case with an entirely different syntax for per-channel specification, and do not involve a union type for the underlying scenegraph value.
- `fontWeight` accepts string and numbers, but is actually an enum type, not a union type.
- `text` is the only exception in that it takes both a string and a string array (for multi-lines).

In addition, my approach was loosely modeled on CSS, in which the border-radius property actually resolves down to four properties (one for each corner). Obviously CSS also provides multi-value convenience options, but there are no "border-top-radius" or "bottom-bottom-radius" properties (nor any similar convenience patterns!). Somewhat following CSS, I'd be happier with support for an array-valued `cornerRadius` property with the following options:

- number or number[] (length 1): set radius for all corners
- number[] (length 4): sets [tl, tr, br, bl]
- number[] (length 2): sets [tl+tr, br+bl] 

The length 2 option differs from CSS: in CSS two element specs map to tl+br, tr+bl. But I think it is reasonable to differ here given your arguments.

Though this approach does introduce a union type, it still keeps the same base type, maintains backwards compatibility, doesn't involve complicated specifier objects, and one can implement a single consistent method for handling corner values by coercing the number case to an array (using vega-util's `array` method, for example).

If you agree, I'd be OK moving ahead with a single `cornerRadius` property that accepts either a number or number array.

In addition, the work so far seems to focus on `rect` marks. We should be sure that this works and is tested for `group` marks as well, which also support `cornerRadius`. Of course, the `arc` mark also has a cornerRadius property, but (a) I don't see a strong need for more control there, and (b) d3-shape only supports singular values anyway.

kanitw:
Ah, I can see why you won't like the object as it would look like this in Vega:

```
cornerRadius: {
  value: {
    topLeft: 1
    topRight: 2
  }
}
```

which I agree is quite weird.  Originally, I thought about it in Vega-Lite mark definition, which wouldn't look as weird: 

```
cornerRadius: {
  topLeft: 1
  topRight: 2
}
```

Given that we better have it be consistent in both VG and VG, maybe let's do the flattened version. 

The question is whether we want the convenient top/bottom/right/left version at all. (Again, I can't even think when would users would like to change `topRight` alone without either `topLeft` or `bottomRight`.)

----

Note that I think the array approach similar to CSS is also bad because people who use Vega-Lite/Altair/other wrappers aren't necessary CSS users and they would find this array order quite confusing (esp. without prior exposure to CSS).  

- People can read `cornerRadius: {topLeft: 1, topRight: 2, bottomRight: 3, bottomLeft: 4}` or the flat versions and figure out what happens right away. 
- On the other hand, `cornerRadius: [1, 2, 3, 4]` is quite confusing (unless people know how CSS works).  

In fact, the length-2 borderRadius in CSS actually sets top-left-and-bottom-right and top-right-and-bottom-left (according to https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius), 

```
/* top-left-and-bottom-right | top-right-and-bottom-left */
border-radius: 10px 5%;
```

So this is quite confusing and doesn't match what we would want for vertical:

> number[] (length 2): sets [tl+tr, br+bl]

donghaoren:
Thanks for the comments!

I agree with @kanitw that the number array is quite confusing.

I'd vote for having the convenient versions with a proper priority order: cornerRadius < cornerRadiusLeft/Right < cornerRadiusTop/Bottom < cornerRadiusTopLeft/TopRight/BottomLeft/BottomRight. Because changing a single corner alone should be a pretty rare case. The issue I can see is whether cornerRadiusLeft should override cornerRadiusTop or vice-versa, but that should rarely occur.


Regarding this:

> Cubic curves are not good approximations to arcs, for large cornerRadius values, it doesn't look correct. We may consider using Bezier curves. Here is a good approximation: http://spencermortensen.com/articles/bezier-circle/.

Do you think it's a good idea to revise the corner curves in a separate PR? The use case is when people want to make one end of a rect completely round for stylistic reasons, it should give a nicer hemicircle.

jheer:
OK. FYI I'm working on a version of this today so no need for any additional work here.

jheer:
I just merged #2127 into the v5.8.0 branch, which adds the individual flat corner radius options discussed above - `cornerRadiusTopLeft`, `cornerRadiusTopRight`, `cornerRadiusBottomRight`, `cornerRadiusBottomLeft` - for both `rect` and `group` marks. That PR also updates rounded corner rendering to use cubic Bezier curves instead of quadratic Bezier curves. Thank you @kanitw for moving this forward and @donghaoren for the initial implementation approach!

Ultimately I decided not to add convenience options for top/left/bottom/right, as (1) they bloat the schema and implementation, (2) the total specification effort difference in Vega is just one channel value, (3) they require rather arbitrary precedence decisions, and (4) they can be supported in higher-level APIs like Vega-Lite and mapped to the primitives now provided by Vega.

Issue #1085, opened 2017-12-04
No support for net (radial) charts?


9662:
Looking at #312 it would appear that vega does not currently support radial aka net aka radar aka [spider charts](http://bl.ocks.org/tpreusse/2bc99d74a461b8c0acb1)?

domoritz:
You can create radar charts similarly to how you can create them in D3. Here is an example that I create a few months ago. 

<img width="584" alt="screen shot 2017-12-04 at 20 27 46" src="https://user-images.githubusercontent.com/589034/33571901-b6e1c7ba-d931-11e7-98d7-004a0b58da54.png">

```
{
  "$schema": "https://vega.github.io/schema/vega/v3.0.json", 
  "config":{"background":"#ffffff"},
  "padding": 5,
  "signals": [
    {"name": "radius","value": 180},
    {"name": "radiusRange","value": [0,180]}
  ],
  "data": [
    {
      "name": "cars",
      "url": "data/cars.csv",
      "format": {
        "type": "csv",
        "parse": {
          "city-mpg": "number",
          "highway-mpg": "number",
          "price": "number",
          "horsepower": "number",
          "num-of-cylinders": "number"
        }
      },
      "transform": [
        {
          "type": "aggregate",
          "groupby": ["body-style"],
          "fields": [
            "city-mpg",
            "highway-mpg",
            "price",
            "horsepower",
            "num-of-cylinders"
          ],
          "ops": [
            "average",
            "average",
            "average",
            "average",
            "average"
          ]
        }
      ]
    },
    {
      "name": "folded",
      "source": "cars",
      "transform": [
        {
          "type": "fold",
          "fields": [
            "average_city-mpg",
            "average_highway-mpg",
            "average_price",
            "average_horsepower",
            "average_num-of-cylinders"
          ]
        }
      ]
    },
    {
      "name": "fields",
      "source": "folded",
      "transform": [
        {
          "type": "aggregate",
          "groupby": ["key"],
          "fields": ["value","value"],
          "ops": ["min","max"]
        }
      ]
    }
  ],
  "scales": [
    {
      "name": "angle",
      "type": "point",
      "range": [6.28,0],
      "padding": 0.5,
      "domain": {"data": "fields","field": "key"}
    },
    {
      "name": "color",
      "type": "ordinal",
      "domain": {"data": "cars","field": "body-style","sort": true},
      "range": "category"
    },
    {
      "name": "average_city-mpg",
      "type": "linear",
[...]

PR #2128, opened 2019-11-05
add radar chart (aka spider chart, polar chart) example

benib:
This PR adds an example of a radar chart (aka spider chart, polar chart, web chart).
See https://en.wikipedia.org/wiki/Radar_chart for descriptions of usescases.

The example posted by @domoritz in https://github.com/vega/vega/issues/1085 never made it to the examples as far as I see.

- This example tries to be as easy to understand as possible by not including any data transforms but the data in long format as part of the spec. (I am not sure if this is really helpful though)
- For the labels and the outer line around the chart the reactive geometry feature is used to show how this can help to keep specs short.

If you have any suggestions to improve the spec, please let me know.


jheer:
Thanks! I went ahead and refined the JSON spec a bit. I plan to add this example after the big v5.8.0 branch merge.

benib:
@jheer cool, I just briefly checked your changes: Some nice refinements, thanks for this. But there are some details slightly off now.
- the `key-label` text marks are now on top of the radial grid lines (the additional 20px in the dx/dy on the labels were there to push them outside the radial grid)
- the additional 20px on the radius for the grid lines makes the max value not sit at the max extent of the gridlines. Probably a thing to argue about...
- the removed autosize makes the chart not exactly fit the specified width. can you elaborate the reason to remove this?

--

Anyways: No stress, take your time with v5.8.0 release. And thanks for all the amazing work you put into vega!

jheer:
Regarding the first two changes, I'm happy to restore the grid lines to the original design to have them coincide with the maximum value. I had made those changes as having the labels anchored on the grid lines felt visually better to me. With the offset, it felt less clear that the text center point was the anchor, causing some label placements to then feel more haphazard... but I don't feel particularly strongly about it!

As for the third point, by using autosize `"none"`, the width and height are simply constants (width/height plus padding), avoiding the need for engaging the auto-sizing logic and corresponding multiple evaluations of the dataflow. Alternatively, we could use `"autosize": {"type": "none", "contains": "padding"}` and set the width and height back to 400 and get an identical result.

benib:
Thanks for the explanations.

I tried something else regarding the label placement. There is now a production rule in place to decide about the text anchor (left/right, top/middle/bottom). My first idea with the label center as the anchor was indeed pretty weird. Now the labels have about the same distance to the grid lines.

Feel free to merge it like this if you agree, or (tell me to) change if you don't.


In case you have some time and energy to explain something I got two more questions. As you probably have better things to do, it's also fine if you ignore these questions.

1. You changed the `key-label` to be based on the `keys` data object instead of the `radial-grid` marks. Since I learned about using this reactive geometry thing I find it pretty powerful especially for things like this. Is there a performance reason to not overuse it? Or something else? I totally see that the way you did it now is perfectly valid, really just wondering...

2. Regarding the autosize, I do not really understand it exactly yet:
If I look at the spec I pushed originally, the autosize makes the whole chart including the labels do fit within the defined width (without a lot of whitespace around). If the labels take up more space, the chart gets smaller, so everything still fits (one could argue about the desired behaviour though). 
With the spec as it is now, setting autosize to `fit/fit-x` doesn't have this effect anymore but the labels are cut off if they get too long.
As far as I see, setting this encode: `"enter": {"x": {"signal": "radius"}, "y": {"signal": "radius"}}` on the top level instead of an additional group mark wrapper makes the difference between the two behaviors.

jheer:
Sure, feel free to merge the updates with dynamic text anchors.

1. Reactive geometry is great, but in this case we still need to recalculate polar to Cartesian coordinates, so in the end it is slightly more efficient to generate marks directly rather than chain them off of another mark (it avoids a little bit of internal overhead, nested property lookups, etc). Not a big deal either way; the more important change was to create a summary dataset with one sets of keys to drive these guide marks.

2. Yes, your original version was more dynamic. I made the changes to provide a simpler example that is more efficient on average. I'd have to dig deeper to see why using the top-level group results in different output (one difference is that the top-level group by default gets its width and height set to the `width` and `height` signal values, but I don't know offhand if that is the root cause).

PR #2136, opened 2019-11-08
Overhaul contour and 2D density support, add heatmap.

jheer:
**vega**
- Update test specifications.

**vega-geo**
- Add `isocontour`, `heatmap`, and `kde2d` transforms.
- Drop d3-contour dependency, use local implementation.
- Deprecate `contour` transform.

**vega-scenegraph**
- Add `image` mark support for pre-loaded `image` property.
- Drop Image `loaded` flag, use built-in `complete` flag.

**vega-statistics**
- Add exposed `bandwidthNRD` method.

**vega-typings**
- Add new geo transform typings.


jheer:
To summarize, this PR contributes the following:
- Drop d3-contour in favor of our own, more granular, adaptation.
- Deprecate the `contour` transform, but preserve it for backwards-compatibility.
- Add the `contours`, `heatmap`, and `kde2d` transforms.
- Add pre-loaded image mark support for dynamically rendered images.

I'd appreciate thoughts on a couple of aspects:
- The existing `contour` transform supports both raw raster grid data and 2D density estimation. The new transforms take more granular approach: `kde2d` for density estimation and raster grids as output, `contours` for generating contours from raster grids, and `heatmap` for generating canvas bitmaps from raster grids.
- The `heatmap` transform supports per-pixel `color` and `opacity` functions, similar to encoding channels. This is accomplished through expressions that are given a pseudo-datum as input, with both the top-level datum and  `$x`, `$y`, `$value`, and `$max` properties specific to the raster grid as input. Can this scheme be successfully handled in a custom Vega-Lite `heatmap` mark type?
- I deprecated the `contour` transform and added the `contours` (plural) transform as a replacement. In an ideal world, we'd simply replace `contour`, but then we'd violate backwards-compatibility. The `contours` name violates our normal scheme of using verbs as transform names. Should we find a different name? Or should we delay this PR until a major version update? Or (if even possible) should we try to "shoehorn" new and old functionality into a single `contour` transform, and deprecate just the old code pathways?

domoritz:
Isoline, isopleth, or isarithm are alternative names (https://en.wikipedia.org/wiki/Contour_line). `isoline` seems like a good alternative to the name `contours`.

jheer:
Thanks @dom. I decided to go with `isocontour`, which follows your suggestion while also maintaining naming continuity with the now deprecated `contour` transform. I've also made enough new examples that I'm feeling comfortable with the new transform parameters, etc. I am now planning to include these updates in v5.8.0.

PR #2137, opened 2019-11-08
Remove gender stereotypical colors from population pyramid

domoritz:
Screen Shot 2019-11-08 at 14 32 19

### DIFF
--- a/docs/examples/population-pyramid.vg.json
+++ b/docs/examples/population-pyramid.vg.json
@@ -58,7 +58,7 @@
       "name": "c",
       "type": "ordinal",
       "domain": [1, 2],
-      "range": ["#1f77b4", "#e377c2"]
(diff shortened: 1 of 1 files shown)


jheer:
Thanks! I'm very supportive of the change, and your suggestion is in line with [this article](https://blog.datawrapper.de/gendercolor/) about alternative colors for gender. However, in this case I find the new colors to be a bit unaesthetic. Another alternative might be #6c4e97, #d5855a (or vice versa!), as in:

image

jheer:
I added a [color change](https://github.com/vega/vega/commit/311f89aec0424f475d2eeae33d334132c2cb9743#diff-5251342b1da855329cc8d8063b89a563) in 5.8.0 branch, in commit 311f89a. Thanks!

PR #2150, opened 2019-11-12
Correctly support signals in configs

domoritz:
This pull request corrects the types of configs. Most config properties can be signals but we previously didn't allow signals.


domoritz:
Replaced by https://github.com/vega/vega/pull/2206

PR #2154, opened 2019-11-12
Add documentation to projection types, correctly mark optional properties, add missing properties

domoritz:
For https://github.com/vega/vega-lite/pull/5542

### DIFF
--- a/packages/vega-typings/types/spec/projection.d.ts
+++ b/packages/vega-typings/types/spec/projection.d.ts
@@ -30,9 +30,12 @@ export interface BaseProjection {
    */
   clipAngle?: number | SignalRef;
   /*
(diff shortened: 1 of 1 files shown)


packages/vega-typings/types/spec/projection.d.ts
```diff
@@ -22,17 +22,20 @@ export type ProjectionType =
   | 'transverseMercator';
 export interface BaseProjection {
   /*
-   * The type of the projection.
+   * The type of the projection. The default is `"mercator"`.
    */
-  type: ProjectionType | SignalRef;
+  type?: ProjectionType | SignalRef;
```
domoritz: Type has a default so it is optional.

PR #2164, opened 2019-11-17
Add multiple implementations of area labeler

chanwutk:
In this PR, an option to choosing the `method` used to place labels in area chart.
The option contains:
- `floodfill`: using flood fill to find the position that can fit the largest rectangle with the same proportion to the placing label without any collision.
- `reduced-search`: similar to `floodfill` but only search along the line along each pair of data points that represents the width of each part of the area.
- `naive`: similar to `reduced-search` but get the first position that can fit the label instead of finding the largest one.

Merges into https://github.com/vega/vega/pull/1708


chanwutk:
# Comparing 3 area-labeler implementations
**Here are the rendering results from the three implementations:**
### With floodfill:
floodfill

### With reduced search space:
reduced-search

### With naive approach:
naive

**Here is the performance testing results for the 3 implementations of area-labelers:**
Screen Shot 2020-02-03 at 5 32 47 PM



The `floodfill` implementation performs significantly slower than the two implementations. Even though it  looks for the largest area to fit each label, the labels placements does not look that better if not better at all. So, I think that using the floodfill approach does not give more benefit than its drawback.
For the naive approach, there is no runtime benefit over the reduced search space approach. And visually, the chart from reduced search space approach does look better since the labels are more well spaced.

If the `label-transform` are going to have multiple algorithms for placing labels, `floodfill` and `reduced-search` would be the two options to put in. If we only need one algorithm, I would prefer the `reduced-search`

chanwutk:
Since the naive approach that we had did not improve any performance, I tested an another implementation of naive approach by not using bitmap at all, and the only condition for placing labels is on the width of each area, following the similar method as the original [job voyager example](https://vega.github.io/vega/examples/job-voyager/). This algorithm will not avoid any collision between any label in the chart, but it gives a significant performance improvement. Here is the performance result:
visualization

Now, if we would like to have 2 choices of implementations for area-labeler for the users to choose, I would think that `"reduced-search"` and the new `"naive"` methods would be the two good options.

domoritz:
Can you fix the travis builds?

chanwutk:
@domoritz the travis test is fixed.

domoritz:
How does this pull request relate to https://github.com/vega/vega/pull/1708?

domoritz:
The job voyager example is a good start but may not be the most representative example. 

Can you try to annotate a map with country names? I think that would be a good test case as country shapes have a large variety.

domoritz:
Actually, labeling countries (i.e. shapes) is tricker. Let's find out what the state of the art is (@chanwutk). For another example, maybe use a streamgraph or a stacked area chart (potentially normalized).

PR #2177, opened 2019-11-20
Fix hanging extent function

tuner:
Hi!

This patch fixes an infinite loop in vega-util's extent function. The bug appeared with input that is all undefined.

Vega editor froze as I was playing with density transform. I had to clear local storages, cookies, etc. to get it back up and running. Somewhat inconvenient...

To reproduce the problem, just try the following spec:

```json
{
  "data": {"url": "data/cars.json"},
  "transform": [
    { "density": "Missing_Field_Causes_a_Hang" }
  ],
  "mark": "area",
  "encoding": {
    "x": {"field": "value", "type": "quantitative"},
    "y": {"field": "density", "type": "quantitative"}
  }
}
```

### DIFF
--- a/packages/vega-util/src/extent.js
+++ b/packages/vega-util/src/extent.js
@@ -8,7 +8,7 @@ export default function(array, f) {
   if (array && (n = array.length)) {
     if (f == null) {
       // find first valid value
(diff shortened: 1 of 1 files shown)


jheer:
Thanks!

Issue #2207, opened 2019-11-28
Separate group translate from moving the clip mask.


domoritz:
It would be great if Vega had a way to move the clip mast for a group independent of translating the marks inside a group. Right now, moving a clip mask means that all marks also get translated. Because of this, Vega-Lite has to generate complex expressions. See https://github.com/vega/vega-lite/pull/5583#discussion_r351919491.

We could introduce new properties `clipX` and `clipY`. Alternatively, we could add a property `translateChildren` that can be set to `false`.

jheer:
After more thought, I don't think we want to go down this route. The boolean-valued group-level `clip` encode property is specifically for clipping to the group width/height/corners. Other marks contained within the group can also specify their own custom clips.

If we want to define a group-level custom clip, we could perhaps do so by allowing values other than boolean for the `clip` encode property, but in the end that is still functionally equivalent to custom clips defined on the marks within the group (which again we already support).

domoritz:
> If we want to define a group-level custom clip, we could perhaps do so by allowing values other than boolean for the clip encode property, but in the end that is still functionally equivalent to custom clips defined on the marks within the group (which again we already support).

I think we could use this in Vega-Lite if there was an easy way to clip by another mark definition. Custom clips have to be SVG paths or spheres, right (https://vega.github.io/vega/docs/marks/#clip)? Generating paths isn't practical for Vega-Lite, though, since we would have to dynamically create paths in Vega expressions (not having access to the data).

domoritz:
I agree that my proposal wasn't good: it only fixes a specific problem without enabling a lot more things more elegantly.

PR #2208, opened 2019-11-29
Adjust clipping region for stroked group elements

### LINKED ISSUE #2197: Clipped group elements should be behind border
When group marks are in a clipped, I would expect them to behind the group's border. However, it looks like the group elements are clipped but in front of the border. 

Screen Shot 2019-11-26 at 11 09 55

I think it would be better if Vega either put the group elements behind the border or clipped off elements along the inner stroke.

donghaoren:
* Adjust the clipping region for group elements when a stroke is used.
* Fixes an SVG renderer issue that the clip region doesn't update when `cornerRadius` is changed from zero to non-zero and vice-versa.

Fixes #2197

### DIFF
--- a/packages/vega-scenegraph/src/SVGRenderer.js
+++ b/packages/vega-scenegraph/src/SVGRenderer.js
@@ -201,10 +201,14 @@ function updateClipping(el, clip, index) {
     mask.setAttribute('d', clip.path);
   } else {
     mask = domChild(el, 0, 'rect', ns);
-    mask.setAttribute('x', 0);
-    mask.setAttribute('y', 0);
+    mask.setAttribute('x', clip.x || 0);
+    mask.setAttribute('y', clip.y || 0);
     mask.setAttribute('width', clip.width);
     mask.setAttribute('height', clip.height);
+  }
+
+  if(el.childNodes.length > 1) {
+    el.removeChild(el.childNodes[1]);
   }
 
   return index + 1;
--- a/packages/vega-scenegraph/src/marks/group.js
+++ b/packages/vega-scenegraph/src/marks/group.js
@@ -22,7 +22,8 @@ function background(emit, item) {
 }
 
 function foreground(emit, item, renderer) {
-  var url = item.clip ? clip(renderer, item, item) : null;
+  var offset = item.stroke ? StrokeOffset : 0;
+  var url = item.clip ? clip(renderer, item, item, offset) : null;
   emit('clip-path', url);
 }
 
@@ -78,7 +79,10 @@ function draw(context, scene, bounds) {
     }
 
     // set clip and bounds
-    if (group.clip) clipGroup(context, group);
+    if (group.clip) {
+      var offset = group.stroke ? StrokeOffset : 0;
+      clipGroup(context, group, offset);
+    }
     if (bounds) bounds.translate(-gx, -gy);
 
     // draw group contents
--- a/packages/vega-scenegraph/src/path/shapes.js
+++ b/packages/vega-scenegraph/src/path/shapes.js
@@ -77,6 +77,24 @@ export function rectangle(context, item, x, y) {
   return rectShape.context(context)(item, x, y);
 }
 
+export function rectangleWithContract(context, item, x, y, contract) {
+  if(contract) {
+    let rs = {
+      width: Math.max(0, item.width - contract),
+      height: Math.max(0, item.height - contract)
+    };
+    // Modify cornerRadius properties appropriately
+    if(item.cornerRadius) rs.cornerRadius = Math.max(0, item.cornerRadius - contract / 2);
+    if(item.cornerRadiusTopLeft) rs.cornerRadiusTopLeft = Math.max(0, item.cornerRadiusTopLeft - contract / 2);
+    if(item.cornerRadiusTopRight) rs.cornerRadiusTopRight = Math.max(0, item.cornerRadiusTopRight - contract / 2);
+    if(item.cornerRadiusBottomLeft) rs.cornerRadiusBottomLeft = Math.max(0, item.cornerRadiusBottomLeft - contract / 2);
+    if(item.cornerRadiusBottomRight) rs.cornerRadiusBottomRight = Math.max(0, item.cornerRadiusBottomRight - contract / 2);
+    return rectangle(context, rs, x + contract / 2 , y + contract / 2);
+  } else {
+    return rectangle(context, item, x, y);
+  }
+}
+
 export function shape(context, item) {
   return (item.mark.shape || item.shape)
     .context(context)(item);
--- a/packages/vega-scenegraph/src/util/canvas/clip.js
+++ b/packages/vega-scenegraph/src/util/canvas/clip.js
@@ -1,4 +1,4 @@
-import {hasCornerRadius, rectangle} from '../../path/shapes';
+import {hasCornerRadius, rectangleWithContract} from '../../path/shapes';
 import {isFunction} from 'vega-util';
 
 export default function(context, scene) {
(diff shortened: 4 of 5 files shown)


jheer:
Thanks! I added a separate fix for the SVG update bug (using a cleaner approach with the `domClear` utility) in #2209, so that is no longer needed here.

Regarding the other parts of the PR, I'm a little uncomfortable with the amount of code bloat / special casing this solution adds. This also leaves me wondering how necessary it really is to address #2197. Is this a truly pressing issue or just a "nice to have"?

In addition, the stroke offset calculations in this PR do not follow the newer approach introduced by #2201, and the strokeWidth calculations should respect widths of zero, not override them with a stroke width of 1.

donghaoren:
Wow, I didn't realize `domClear` has an additional parameter.

When implementing corner radius for stroked stacked bars in Vega-Lite:


image


We have to create a separate group with the same `facet` transform just for the stroke. That's a lot of overhead. 

```js
{
  "background": "white",
  "padding": 5,
  "width": 200,
  "style": "cell",
  "data": [
    {
      "name": "source_0",
      "url": "data/barley.json",
      "format": {"type": "json"},
      "transform": [
        {
          "type": "aggregate",
          "groupby": ["variety", "site"],
          "ops": ["sum"],
          "fields": ["yield"],
          "as": ["sum_yield"]
        },
        {
          "type": "stack",
          "groupby": ["variety"],
          "field": "sum_yield",
          "sort": {"field": ["site"], "order": ["descending"]},
          "as": ["sum_yield_start", "sum_yield_end"],
          "offset": "zero"
        }
      ]
    }
  ],
  "signals": [
    {"name": "y_step", "value": 20},
    {
      "name": "height",
      "update": "bandspace(domain('y').length, 0.1, 0.05) * y_step"
    }
  ],
  "marks": [
    {
      "type": "group",
      "from": {
        "facet": {
          "data": "source_0",
          "name": "stack_group_main",
          "groupby": ["variety"],
          "aggregate": {
            "fields": [
              "sum_yield_start",
              "sum_yield_start",
              "sum_yield_end",
              "sum_yield_end"
            ],
            "ops": ["min", "max", "min", "max"]
          }
        }
      },
      "encode": {
        "update": {
          "y": {"scale": "y", "field": "variety"},
          "height": {"scale": "y", "band": true},
          "cornerRadius": {"value": 10},
          "x": {
            "signal": "min(scale('x',datum[\"min_sum_yield_start\"]),scale('x',datum[\"max_sum_yield_start\"]),scale('x',datum[\"min_sum_yield_end\"]),scale('x',datum[\"max_sum_yield_end\"]))"
          },
          "x2": {
            "signal": "max(scale('x',datum[\"min_sum_yield_start\"]),scale('x',datum[\"max_sum_yield_start\"]),scale('x',datum[\"min_sum_yield_end\"]),scale('x',datum[\"max_sum_yield_end\"]))"
          },
          "clip": {"value": true}
        }
      },
      "marks": [
        {
          "type": "group",
          "encode": {
            "update": {
              "x": {
                "signal": "-min(scale('x',parent[\"min_sum_yield_start\"]),scale('x',parent[\"max_sum_yield_start\"]),scale('x',parent[\"min_sum_yield_end\"]),scale('x',parent[\"max_sum_yield_end\"]))"
              },
              "height": {"field": {"group": "height"}}
            }
          },
          "marks": [
            {
              "name": "marks",
              "type": "rect",
              "style": ["bar"],
              "from": {"data": "stack_group_main"},
              "encode": {
                "update": {
                  "stroke": {"value": "black"},
                  "fill": {"scale": "color", "field": "site"},
                  "x": {"scale": "x", "field": "sum_yield_end"},
                  "x2": {"scale": "x", "field": "sum_yield_start"},
                  "height": {"field": {"group": "height"}}
                }
              }
            }
          ]
        }
      ]
    },
    {
      "type": "group",
      "from": {
        "facet": {
          "data": "source_0",
          "name": "stack_group_main_copy",
          "groupby": ["variety"],
          "aggregate": {
            "fields": [
              "sum_yield_start",
              "sum_yield_start",
              "sum_yield_end",
              "sum_yield_end"
            ],
            "ops": ["min", "max", "min", "max"]
          }
        }
      },
      "encode": {
        "update": {
          "y": {"scale": "y", "field": "variety"},
          "height": {"scale": "y", "band": true},
          "cornerRadius": {"value": 10},
          "x": {
            "signal": "min(scale('x',datum[\"min_sum_yield_start\"]),scale('x',datum[\"max_sum_yield_start\"]),scale('x',datum[\"min_sum_yield_end\"]),scale('x',datum[\"max_sum_yield_end\"]))"
          },
          "x2": {
            "signal": "max(scale('x',datum[\"min_sum_yield_start\"]),scale('x',datum[\"max_sum_yield_start\"]),scale('x',datum[\"min_sum_yield_end\"]),scale('x',datum[\"max_sum_yield_end\"]))"
          }
        }
      },
      "marks": [
        {
          "type": "rect",
          "encode": {
            "update": {
              "width": {"field": {"group": "width"}},
              "height": {"field": {"group": "height"}},
              "cornerRadius": {"value": 10},
              "stroke": {"value": "black"},
              "fill": {"value": null}
            }
          }
        }
      ]
    }
  ],
  "scales": [
    {
      "name": "x",
      "type": "linear",
      "domain": {
        "data": "source_0",
        "fields": ["sum_yield_start", "sum_yield_end"]
      },
      "range": [0, {"signal": "width"}],
      "nice": true,
      "zero": true
    },
    {
      "name": "y",
      "type": "band",
      "domain": {"data": "source_0", "field": "variety", "sort": true},
      "range": {"step": {"signal": "y_step"}},
      "paddingInner": 0.1,
      "paddingOuter": 0.05
    },
    {
      "name": "color",
      "type": "ordinal",
      "domain": {"data": "source_0", "field": "site", "sort": true},
      "range": "category"
    }
  ],
  "axes": [
    {
      "scale": "x",
      "orient": "bottom",
      "gridScale": "y",
      "grid": true,
      "tickCount": {"signal": "ceil(width/40)"},
      "domain": false,
      "labels": false,
      "maxExtent": 0,
      "minExtent": 0,
      "ticks": false,
      "zindex": 0
    },
    {
      "scale": "x",
      "orient": "bottom",
      "grid": false,
      "title": "Sum of yield",
      "labelFlush": true,
      "labelOverlap": true,
      "tickCount": {"signal": "ceil(width/40)"},
      "zindex": 0
    },
    {
      "scale": "y",
      "orient": "left",
      "grid": false,
      "title": "variety",
      "zindex": 0
    }
  ],
  "legends": [
    {
      "fill": "color",
      "gradientLength": {"signal": "clamp(height, 64, 200)"},
      "symbolType": "square",
      "title": "site",
      "encode": {"symbols": {"update": {"stroke": {"value": "black"}}}}
    }
  ]
}
```

#2201 alone can't fix this. Here is the result:

image

This is because clipping is done halfway through the stroke.

---

`rectangleWithContract` is the part I don't like, because we are using an approximate arc for corner radius, it's not going to perfectly align with the inner stroke; and clipping can produce 1/2 pixel artifacts  due to antialiasing (can be seen in the canvas renderer). Having an actual `foreground` element seems the right solution to me, and probably less bloat.

donghaoren:
We may be able to find a better solution in Vega-Lite once #2207 is implemented. Perhaps we can leave this on hold.

donghaoren:
Changing the rendering order won't work, because it's going to break existing charts. In the example below, the `x`, `y` domain lines are overdrawn:

image

So, if we use this approach, we have to add an option to enable it.

However, clipping the inner stroke also generates undesirable artifacts, especially in the canvas renderer. If you look closely at the picture below, there are white pixels around the border between the black border and the blueish bar.

image

There are problems with both approaches. This makes me also wonder whether we should address this in Vega. Perhaps in Vega-Lite, we can go with the workaround we already have (see https://github.com/vega/vega-lite/pull/5583#issuecomment-559912907 for an improved version). What do you think @domoritz?

domoritz:
I generally prefer to fix things in Vega rather than working around bugs in Vega-Lite.

jheer:
This PR is superceded by #2220, which instead adds foreground drawing support for group strokes via the group mark `strokeForeground` property.

PR #2222, opened 2019-12-03
Fix black group when strokeForeground is applied

donghaoren:
**vega-scenegraph**
- Fix black group when strokeForeground is applied.

This is because the fill style is only set to null when the fill encoding is supplied, but we always need to set to null.


jheer:
Thanks! The problem is that the `values` hash for the foreground (used to avoid unnecessary DOM writes) was not being properly initialized with a default fill value. As a result, undefined fills (which should result in no fill) were `===` matching against the empty values hash.

PR #2228, opened 2019-12-04
Use infinite bin values outside extent bounds.

### LINKED ISSUE #2227: Should bins implicitly filter data that is out of range?
Vega-lite's default for binning is to assign out-of-range data to the nearest defined bin. This is not a default I've come across in other charting libraries, and is potentially *very* misleading. For example, here is a zoomed-in normal distribution ([vega editor](https://vega.github.io/editor/#/url/vega-lite/N4KABGBEAmCGAutIC4zEgZwKYEcCuWAdgMZYpqaIBO85ADADRQbwD2ADuQIx28C+fBuCjwqsQhgBmrKgFtyAbXTFYAG2J5VCMqkyxZ7VVgByM2WoAUASkhNIsDOUgAPSHwC6QiJHNUA1k4ARrBUtsKQRMSs0ACWhADm5KAQ3q6o6PAAnuw6UPji8DGIhQBuZHaBcUkRzvBEtKgKALRcTFzuglCSMViq0E6ugsLemdVZOU75hIXFMWW2ULDx8VRY8dpOUXjTbsJ8IHxAA)):
```json
{
  "data": {"sequence": {"start": 0, "stop": 1000}},
  "transform": [{"calculate": "sampleNormal()", "as": "x"}],
  "mark": "bar",
  "encoding": {
    "x": {"field": "x", "type": "quantitative", "bin": {"extent": [-1, 1]}},
    "y": {"type": "quantitative", "aggregate": "count"}
  }
}
```
visualization (57)

If you don't know to look for this (and my guess is most users wouldn't) it would lead you to make erroneous conclusions about the content of the data.

jheer:
**vega-transforms**
- Use infinite bin values outside extent bounds.

This PR changes the `bin` transform output such that values outside the bin `extent` (modulo the step size) are given infinite bin values. If the value is less than the lower extent, the bin boundaries are given as `[-Infinity, -Infinity]`; if greater than the upper extent, `[+Infinity, +Infinity]` is returned.

An alternative scheme would be to return `[-Infinity, start]` and `[stop, +Infinity]`, which is a more mathematically accurate representation. However, the scheme above has several advantages: it allows quick testing of out-of-range bins (check one value, not two), it still results in infinite values when requesting bin point values (`interval: false`), and involves significantly simpler code paths. Moreover, one can easily recover "true" ranges by checking the sign of the Infinity and consulting the computed `bin.start` and `bin.stop` values.

Fixes #2227.

### DIFF
--- a/packages/vega-transforms/src/Bin.js
+++ b/packages/vega-transforms/src/Bin.js
@@ -64,6 +64,7 @@ prototype.transform = function(_, pulse) {
         // maximum bin value (exclusive)
         // use convoluted math for better floating point agreement
         // see https://github.com/vega/vega/issues/830
+        // infinite values propagate through this formula! #2227
         t[b1] = v == null ? null : start + step * (1 + (v - start) / step);
       }
     : function(t) { t[b0] = bins(t); }
@@ -92,12 +93,13 @@ prototype._bins = function(_) {
 
   var f = function(t) {
     var v = field(t);
-    if (v == null) {
-      return null;
-    } else {
-      v = Math.max(start, Math.min(+v, stop - step));
-      return start + step * Math.floor(EPSILON + (v - start) / step);
-    }
+    return v == null ? null
+      : v < start ? -Infinity
+      : v > stop ? +Infinity
+      : (
+          v = Math.max(start, Math.min(+v, stop - step)),
+          start + step * Math.floor(EPSILON + (v - start) / step)
+        );
   };
 
   f.start = start;
--- a/packages/vega-transforms/test/bin-test.js
+++ b/packages/vega-transforms/test/bin-test.js
@@ -65,10 +65,13 @@ function testBin(t, b, extent, step) {
   }
 
   // test bins that precede extent
-  t.equal(b({v: f(-1)}), f(0));
+  t.equal(b({v: f(-1)}), -Infinity);
 
   // test very last, inclusive bin
   t.equal(b({v: f(steps)}), f(steps - 1));
+
+  // test bins that exceed extent
+  t.equal(b({v: f(steps+1)}), Infinity);
 }
 
 tape('Bin handles tail aggregation for last bin', function(t) {
@@ -87,7 +90,7 @@ tape('Bin handles tail aggregation for last bin', function(t) {
   t.equal(bin.value({v:28}), 25);
   t.equal(bin.value({v:29}), 25);
   t.equal(bin.value({v:30}), 25);
-  t.equal(bin.value({v:35}), 25);
+  t.equal(bin.value({v:31}), Infinity);
 
   t.end();
 });


domoritz:
I'm not sure I like `[-Infinity, -Infinity]` and `[+Infinity, +Infinity]`. The tuples outside the extent are in `[-Infinity, start]` and `[stop, +Infinity]` and so I feel that those are the bins we should assign. 

> it allows quick testing of out-of-range bins (check one value, not two)

I don't understand this. Can you elaborate? 

> Moreover, one can easily recover "true" ranges by checking the sign of the Infinity and consulting the computed bin.start and bin.stop values.

If it's easy to recover the true bins, why does supporting it require much more code in Vega?

kanitw:
Here are my answers to @domoritz's points.  


## Easy to Filter 

> > it allows quick testing of out-of-range bins (check one value, not two)


> I don't understand this. Can you elaborate?


With `[-Infinity, start]` and `[stop, +Infinity]`, we need to check both `bin0` and `bin1` to filter out-of-range values (`bin0` is `-Inf` or `bin1` is `Inf`), in addition to null checks.

Jeff's proposed scheme means that VL's output VG specs can just check if `bin0` is a valid number (since we also want to filter null values). 

This scheme also consistently works when we set `interval: false` (which only outputs `bin0`).

Given we probably will just filter for now, this is best for our current solution. 

## Simpler code path

> involves significantly simpler code paths

Looking at the code, if I understand correctly, this scheme is also simpler because we don't need to convert `+Inf` in the input data to "stop" since we will output `+Inf` anyway.  


## What if we wanna show LHS/RHS 

> > Moreover, one can easily recover "true" ranges by checking the sign of the Infinity and consulting the computed bin.start and bin.stop values.

> If it's easy to recover the true bins, why does supporting it require much more code in Vega?

This is a good question.  If we want to show LHS/RHS instead of filtering, the proposed scheme requires less code in some cases, but more code in certain cases.

Here I enumerate different cases for further considerations:

### Quantitative Scales

1a) Adding add left/right-side to quantitative-scale bin would require conditional logic to deal with positioning `-Inf`, `Inf` at `[bin.start-bin.step, bin.start]`, `[bin.stop, bin.stop+bin.step]` (which is more natural to read from the bin signal anyway) and customizing axis labels for LHS/RHS (conditional logic to replace `bin.start-bin.step` and `bin.stop+bin.step` with `-Inf` and `Inf`).   

image

1b) Similarly, for binned size with linear scales and binned color with bin-ordinal scale (which we use for type=Q in VL). 

For encoding, we probably need to replace `-Inf` with `bin.start-bin.step` with any binning scheme.  If we do `[Infinity, Infinity]`, then we need to similarly replace `Inf` with `bin.stop`.  The proposed scheme will require more code, but would make the code looks more consistent as it needs replacements on both sides.

For legends, the legend for colors seems fine as is: 

image

The legend for size is a bit tricky: (How do we make `-Inf - start`?)

image


### Ordinal Scales

However, this scheme would be more complex to include left-side/right-side in ordinal scales (which is less commonly used).  Imagine we can add `-Inf - start` and `stop - Inf` to the bins in these two plots.


image


image

With this scheme, we need to make conditional logic to recover `bin.start`, `bin.stop` from the signal instead of reading directly from `bin0`, `bin1`.

Basically, we need to add logic to read signals to the following code: 

```
{
          "type": "formula",
          "expr": "!isValid(datum[\"bin_maxbins_6_Acceleration\"]) ? \"null\" : format(datum[\"bin_maxbins_6_Acceleration\"], \"\") + \" - \" + format(datum[\"bin_maxbins_6_Acceleration_end\"], \"\")",
          "as": "bin_maxbins_6_Acceleration_range"
        }
```

This is not the end of the world, but clearly adds a bit more complexity.
(Another alternative would be to show `null` / `-Infinity` / `Infinity` without including start / stop for LHS/RHS.

Given that people would normally use bin with type=quantitative in VL and the Vega compiler code is much simpler, it is ok to prioritize quantitative scale case?

domoritz:
For 1a), instead of supporting this in Vega-Lite, shouldn't Vega axes for bin scales support -inf and +inf?

jheer:
> For 1a), instead of supporting this in Vega-Lite, shouldn't Vega axes for bin scales support -inf and +inf?

FWIW Vega doesn't really have any separate concept of "bin scales" (other than `bin-ordinal`). For example, in 1a we have just a linear scale with additional bin metadata used to guide which ticks to include. The scale transform itself is unchanged.

(While it's quite valuable to think through the ramifications of these design decisions, we should also keep in mind that people rarely change the bin extent, and if they do, they probably nearly always want to filter out-of-extent values...)

An additional advantage of my proposed scheme is that (AFAICT) it is backwards compatible: compiled Vega-Lite specs already perform a finite-value check, but for the start bin value only. If necessary we could consider "breaking" changes (i.e., a different encoding of out-of-extent values) in a future major release of Vega.

domoritz:
Sounds good. I don't have any strong opinions here.

kanitw:
> (While it's quite valuable to think through the ramifications of these design decisions, we should also keep in mind that people rarely change the bin extent, and if they do, they probably nearly always want to filter out-of-extent values...)

Yep, I don't think we should implement the LHS/RHS right now, I was just thinking to make sure that this solution would be ok for future extension.

I think it seems fine. :)

Issue #2255, opened 2019-12-18
Allow CSS classes to be set by encoding


domoritz:
In some applications, one might want to style a Vega spec using CSS. To also enable dynamic styling, it would be necessary that we support setting the CSS class dynamically based on a signal. Right now, the only way to set a CSS class is via `name` or `role` but neither can be driven by signals.

cc @mathiastiberghien

mathiastiberghien:
I would like to add some information about this.
The problem when integrating vega visualizations in a themed application (theme can be changed at runtime) are the following:

- using config property in vega spec to update style require a javascript implementation: if we want to read theme colors, we have to read into document.stylesheet which is not a good practice becasue unpredictable
- If we want to use the css approach we can rely on vega role classes which are great to style axis, legends and titles, but not for marks. For the marks, we can use the name to inject css class when rendering in svg but it doesn't seem a good practice because it's a hack.
- Whatever the approach there still a problem: let say that I have rect mark for representing columns and that I want to highlight the mark when mouse hover. I will have something like that

`{
      "type": "rect",
      "encode": {
        "enter": {
          ...
        },
        "update": {
          "fill": {"value": "steelblue"}
        },
        "hover": {
          "fill": {"value": "red"}
        }
      }
    },
`
With that configuration, at runtime, the columns will always be red when the user move the mouse over the column.

Having a 'css' property that is part of the encoding with a unique role to add the class in 'enter' mode and remove the old class then add the new class in 'update' or 'hover' mode.

This will allow to write something like this

`{
      "type": "rect",
      "encode": {
        "enter": {
          ...
        },
        "update": {
          "css": {"value": "mydefaultfillcssselector"}
        },
        "hover": {
          "css": {"value": "myhighlightfillcssselector"}
        }
      }
    },`

Adding that feature should have a minimal implementation impact and allow a standard themification of vega visualizations.

I would be happy to work on it if it appear to be an acceptable change

see-mike-out:
One potentially useful addition to this suggestion could be allowing `expr` for class names (like axis label expressions), to make it further customizable with custom scripts.

PR #2258, opened 2019-12-20
Added tests and handle css for group (background and foreground)

mathiastiberghien:
This is a proposition for feature request #2255

Creating test for the PR #2257, I noticed that group have background and foreground elements.
I used the attr method of marks to handle css and added a cssForeground for groups to be able to style foreground part of a group.

I updated documentation, schemas, and modified SVGRenderer tests to ensure every mark is rendered correctly


mathiastiberghien:
Any comment/review on this one? I would like to know if that change will be integrated or not.
Thanks!

jheer:
Thanks for the PR! The specific edits made here make sense given the goals (modulo some deviations from the project's syntax styles). However, I'm a bit uncomfortable with the overall approach, as adding a specific CSS channel directly breaks Vega's scenegraph abstraction. This would seem to encourage the creation of specifications that work in some situations (SVG rendering) but not others, undercutting the philosophy of a coherent declarative specification language. If the goal is to provide a "standard" method for dynamic theming in Vega, we need to do so in a way that works across renderers.

While we do permit some level of leaky abstraction (e.g., mark `name`, `role`, and `type` properties are passed through to CSS classes), those properties are nevertheless meaningful regardless of renderer. Is there a similar move we might make here to enable richer CSS use in custom cases without introducing strong inconsistencies into the language design?

(Flagging @arvind as well, in case he has thoughts on the matter.)

mathiastiberghien:
Hi,

Thanks for the anser. I understand your concern. I see another approach that preserve abstraction but requires control on dom elements: The marks should fire events providing their document element and contextual information (style settings). The uses will be then be able to use to add their own implementation of the themification directly on the elements. I’m a little worried with that approach because user might break the rendering process but at least the inconsistency is preserved.

 

I was hoping that the css attribute won’t be a problem because of the use of role property and name property. We exploit these properties to inject css class but, the problem with these properties is that they have another purpose and they are updated only when rendering the element the first time, not on update. Rename the property into “runtimeContext” or “theme” is not acceptable ?

jheer:
(Closing at request of author.)

PR #2283, opened 2020-01-08
Upgrade to @rollup/plugin-node-resolve

domoritz:
rollup-plugin-node-resolve is deprecated

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "rimraf": "^3.0.0",
     "rollup": "^1.27.8",
     "@rollup/plugin-json": "^4.0.0",
-    "rollup-plugin-node-resolve": "^5.2.0",
+    "@rollup/plugin-node-resolve": "^7.0.0",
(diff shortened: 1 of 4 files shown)


jheer:
Thanks!

jheer:
Unfortunately this breaks the browser build, which our CI tests fail to realize. With the new plug-in the package.json browser redirects within Vega packages are not followed, which (on first glance) appears to happen due to an interaction with our use of Yarn workspaces and the corresponding node_modules symlinks. We may need to revert this PR and/or submit fixes to the rollup plugin...

jheer:
Still need to check more thoroughly, but is looks like adding the following to the rollup node-resolve options fixes the issue:

```
customResolveOptions: {preserveSymlinks: false}
```

I'm still not sure why this differs from the previous (deprecated) plug-in, as the code is very similar between the two. I haven't yet isolated a difference that would explain the discrepancy.

PR #2289, opened 2020-01-09
Fix splitAccessPath escape handling.

### LINKED ISSUE #2287: Split access path does not respect escaped brackets
This causes a bug in Vega-Lite (https://github.com/vega/vega-lite/issues/5660). 

```
$ vega.splitAccessPath('y\\[foo\\]')
["y[foo]"]
```

I would expect

```
$ vega.splitAccessPath('y\\[foo\\]')
["y\\[foo\\]"]
```

or is that not what this function is supposed to do?

jheer:
**vega-util**

- Fix `splitAccessPath` escape handling, add tests.

Fix #2287.

### DIFF
--- a/packages/vega-util/src/splitAccessPath.js
+++ b/packages/vega-util/src/splitAccessPath.js
@@ -20,7 +20,8 @@ export default function(p) {
     c = p[j];
     if (c === '\\') {
       s += p.substring(i, j);
-      i = ++j;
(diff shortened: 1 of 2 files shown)


domoritz:
Thank you

domoritz:
Hmm, this doesn't seem to fix https://github.com/vega/vega-lite/pull/5730. I need to look more into this.

PR #2305, opened 2020-01-14
BUG: fix r-squared computation for poly and quad regression

jakevdp:
I discovered in the course of implementing regression transforms in [altair-transform](https://github.com/altair-viz/altair-transform) that in vega's ``poly`` and ``quad`` regression, the r-squared values are being computed with respect to zero, rather than with respect to the mean.

This fixes the issue in the source files... I suspect there's some sort of build step I need to complete this pull request, but I don't have my local dev environment set up for building vega.

### DIFF
--- a/packages/vega-statistics/src/regression/poly.js
+++ b/packages/vega-statistics/src/regression/poly.js
@@ -49,7 +49,7 @@ export default function(data, x, y, order) {
   return {
     coef: uncenter(k, coef, -ux, uy),
     predict: predict,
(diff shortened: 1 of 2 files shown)


jheer:
I've incorporated the fix into the v5.9.1 branch for inclusion in that release. Looks like this error was inadvertently added when we added mean-centering to improve the robustness of quad and poly regression, but it didn't change the results in the unit tests. Thank you for catching this!

PR #2344, opened 2020-01-27
performance: memoize measurements of text width

domoritz:
Measuring the width of some text is a somewhat expensive operation that runs very often. This pull request adds simple memoization to speed up most interactive Vega charts.

I can provide some more detailed benchmark results but first I wanted to see whether there are any concerns. Does it make sense to use `fastmap` instead of an object here?


jheer:
The approach looks reasonable to me. Fastmap should not be necessary, as it is designed to help with deletions (really, cases where objects can be inserted / removed / re-inserted such as in interactive filters and data joins).

Memory pressure is a potential issue here, so we might want to consider a cache with a proper eviction policy (e.g., LRU). We should think about the cache dynamics for visualizations with lots of text (thousands of strings) and especially text that is dynamically updated (e.g., label resizing or interactive wordclouds).

A minor issue is that I don't know which is faster for cache key creation: creating the font string directly or accessing it from the graphics context.

I agree that we should run benchmarks prior to accepting changes here. Thanks!

domoritz:
I tested this to compare time before and after the change. 

```js
// width computation is expensive so we are memoizing it
const widthCache = lru(TEXT_WIDTH_CACHE_SIZE);

let timeBefore = 0;
let timeAfter = 0;

function measure(text) {
  const start = new Date();
  const key = context.font + text;
  let width = widthCache.get(key);
  if (width === undefined) {
    width = context.measureText(text).width;
    widthCache.set(key, width);
  }
  timeAfter += (new Date()).getTime() - start.getTime()

  const start2 = new Date();
  context.measureText(text).width;
  timeBefore += (new Date()).getTime() - start2.getTime()

  return width;
}
```

I measured times by opening the chart (after reloading the window) and playing with the sliders for a few seconds. 

In this example, times went down from 12 to 6.

Screen Shot 2020-01-29 at 22 41 59

For this example, they went from 17 to 8.

Screen Shot 2020-01-29 at 22 43 25

Here, we went from 58 to 32.

Screen Shot 2020-01-29 at 22 47 23

I'm pretty convinced at this point. Do you want to see more?

domoritz:
And I think you're right. Setting the font on the canvas takes a while so I refactored the code a bit to delay when we assign the font to the canvas.

jheer:
This looks good, thanks! My plan is to move the cache support to vega-util, and include font metrics memoization as part of v5.10.

PR #2353, opened 2020-01-31
fix: add maxbins and extent to timeunit transform schema and parser

haldenl:
Adding maxbins and extent to the time unit transform definition s.t. they are recognized by the schema and parser. They are currently in the [docs](https://vega.github.io/vega/docs/transforms/timeunit/) but aren't functional.

### DIFF
--- a/packages/vega-transforms/src/TimeUnit.js
+++ b/packages/vega-transforms/src/TimeUnit.js
@@ -22,6 +22,8 @@ TimeUnit.Definition = {
     { "name": "interval", "type": "boolean", "default": true },
     { "name": "units", "type": "string", "array": true },
     { "name": "step", "type": "number", "default": 1 },
+    { "name": "maxbins", "type": "number", "default": 40 },
+    { "name": "extent", "type": "string", "array": true},
     { "name": "timezone", "type": "enum", "default": "local", "values": ["local", "utc"] },
     { "name": "as", "type": "string", "array": true, "length": 2, "default": OUTPUT }
   ]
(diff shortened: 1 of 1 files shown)


packages/vega-transforms/src/TimeUnit.js
```diff
@@ -76,7 +78,7 @@ prototype._floor = function(_, pulse) {
   let {units, step} = _.units
     ? {units: _.units, step: _.step || 1}
     : timeBin({
-      extent:  extent(pulse.materialize(pulse.SOURCE).source, _.field),
+      extent:  _.extent ? _.extent.map(d => new Date(d)) : extent(pulse.materialize(pulse.SOURCE).source, _.field),
```
haldenl: Was not sure how to deal with `extent` here. The [docs](https://vega.github.io/vega/docs/transforms/timeunit/) suggest a `Date` object. Is this an appropriate way of dealing with that?

jheer: These date values are being handed in from the specification, so might be a bit dicey. `new Date` with a string array is definitely not what we want. Internally, Vega is happy if the values are `Date` objects or numbers corresponding to UNIX timestamps. The safest option would be for a user to use signals / expression functions to define proper dates. (We could in theory support strings, but then we need proper parsing, too. The `toDate` function in vega-util defaults to calling `Date.parse` for non-numeric, non-Date inputs, but this method varies significantly across browsers.) Following the [docs](https://vega.github.io/vega/docs/types/#Date) I think we should expect a date type here, not a string.

jheer: I just double-checked the vega-schema package and we don't currently have support for a `"date"` type in our transform schema generation code. So I will need to add that to complete this PR. @haldenl don't worry about doing more here, I'll take care of it.

haldenl: Ahh I see. Sounds good, thanks Jeff!

jheer:
Merged modified branch via command line.

PR #2367, opened 2020-02-05
fix: Vega shouldn't fix Vega typings version

kanitw:
Since there might be error in the typings, we should allow the patch version to change.

The current "fixed" version means that Vega-Lite can't use a patch update in Vega-typings without a Vega release.

### DIFF
--- a/packages/vega/package.json
+++ b/packages/vega/package.json
@@ -49,7 +49,7 @@
     "vega-statistics": "1.7.2",
     "vega-time": "1.0.0",
     "vega-transforms": "4.6.0",
-    "vega-typings": "0.12.0",
+    "vega-typings": "~0.12.1",
     "vega-util": "1.12.2",
     "vega-view": "5.4.0",
     "vega-view-transforms": "4.5.0",


domoritz:
Maybe we shouldn't fix any version and always allow updates to the patch version. IMHO, lockfiles are for fixing versions. 

The previous issue that made up fix versions was https://github.com/vega/vega/issues/2089

kanitw:
From that discussion

>  between those two minor version the minimum Typescript support went from 2.7 (fine for me) to 3.4 (my project can't build)


I think we can use `~` for vega-typings, but require at least a minor version bump when we upgrade TS. (We should similarly require minor bump for major TS upgrade in vega-util and Vega-Lite.)

domoritz:
Makes sense. Do you want to update this pull request to allow patch version updates for all dependencies?

kanitw:
> Makes sense. Do you want to update this pull request to allow patch version updates for all dependencies?

Let's see what @jheer says.

jheer:
Superceded by #2375.

PR #2376, opened 2020-02-11
vega-typings: ProjectTransform "as" takes array

danmarshall:


domoritz:
What does `as: ["foo", null, "bar"]` do?

danmarshall:
https://vega.github.io/vega/docs/transforms/project/

## Transform Parameters

| Property            | Type                           | Description   |
| :------------------ | :----------------------------: | :------------ |
| fields              | Field[] | The data fields that should be copied over in the projection. If unspecified, all fields will be copied using their existing names.|
| as                  | String[] | For each corresponding field in the _fields_ array, indicates the output field name to use for derived data objects.|

domoritz:
What about the null? I think it should be outside the array.

danmarshall:
Oh, sorry I misread your question @domoritz . Seems like a null will use the original name:

```json
{
  "data": [
    {
      "name": "table",
      "values": [
        {"category": "A", "a really long name": 28, "value": 5},
        {"category": "B", "a really long name": 55, "value": 5},
        {"category": "C", "a really long name": 87, "value": 5}
      ],
      "transform": [
        {
          "type": "project",
          "fields": ["category", "a really long name", "value"],
          "as": ["cat", null, "val"]
        }
      ]
    }
  ]
}
```
image

[link to vega editor](https://vega.github.io/editor/#/url/vega/N4KABGBEAkDODGALApgWwIaQFxUQFzwAdYsB6UgN2QHN0A6agSz0QFcAjOxge1IRQyUa6SgFY6AK1jcAdpAA04KABN0eTDgDaSiKAj6oM9KmTYo69gBtTigxEgV0l1slhntd-cEjw1NbgBOAJ5mkACCClDoYAHITpZBYJay1GBGJmYATAAc8lCOzqY4ogC+tp5g3r54-sGhAEKRkNGx8YnJMqnpRWCionkOTi5mpeWeVX7UgSE4kADCTS1xlglJKWnGPdkA7AMFw8UlOvoAumP6kHgB6DKwAGaBqO7HBnoV9nhBhD2QhAHcEmQ8DwChedkgd0YyEsyjcWh8k2mixiy1WHS6mya+1MZzBBmacLAmgRILyMlYKz2TkgJzxYCOnlpBgZYFpJSAA)

domoritz:
Okay. Makes sense. This is consistent with the schema then.

PR #2380, opened 2020-02-16
feat: add support for aria properties with defaults for guides

domoritz:
**vega-parser**

- Add support for `ariaLabel`, `ariaRole`, `ariaRoleDescription`, and `ariaHidden` to encodings, axes, legends, and titles.
- Automatically set the role of guide groups to `graphics-object`.
- Create default descriptions for axes and legends.
- Set default aria role and aria role description for marks.

**vega-schema**

- Add support for `ariaLabel`, `ariaRole`, `ariaHidden`, and `ariaRoleDescription` to encodings, axes, legends, and titles.

**vega-scenegraph**

- Add support for `ariaLabel`, `ariaRole`, `ariaHidden`, and `ariaRoleDescription` to SVG renderer.

**vega-typings**

- Add `ariaLabel`, `ariaRole`, `ariaHidden`, and `ariaRoleDescription` to encodings, axes, legends, and titles.

Now we can support https://blog.tenon.io/accessible-charts-with-aria and https://www.w3.org/TR/graphics-aria-1.0/.

To document:
* What aria properties are exposed, what are the defaults. 
* Best practices for adding aria labels to marks.
* Best practices for adding aria labels to groups to allow screen reader users to skip over large numbers of marks.


domoritz:
I've been playing with this pull request to make a chart accessible and here is what I have been working with

```json
{
  "width": 400,
  "height": 200,
  "padding": 5,
  "description": "Bar chart showing amounts per category.",
  "title": {
    "text": "Interesting chart title",
    "subtitle": "An even more interesting subtitle",
    "encode": {
      "title": {"update": {"ariaLabel": {"value": "chart title"}}},
      "subtitle": {"update": {"ariaLabel": {"value": "chart subtitle"}}}
    }
  },
  "data": [
    {
      "name": "table",
      "values": [
        {"category": "A", "amount": 28},
        {"category": "B", "amount": 55},
        {"category": "C", "amount": 43},
        {"category": "D", "amount": 91},
        {"category": "E", "amount": 81},
        {"category": "F", "amount": 53},
        {"category": "G", "amount": 19},
        {"category": "H", "amount": 87}
      ]
    }
  ],
  "scales": [
    {
      "name": "xscale",
      "type": "band",
      "domain": {"data": "table", "field": "category"},
      "range": "width",
      "padding": 0.05,
      "round": true
    },
    {
      "name": "yscale",
      "domain": {"data": "table", "field": "amount"},
      "nice": true,
      "range": "height"
    },
    {
      "name": "qcolor",
      "domain": {"data": "table", "field": "amount"},
      "range": "heatmap"
    },
    {
      "name": "ncolor",
      "type": "ordinal",
      "domain": {"data": "table", "field": "category"},
      "range": "category"
    }
  ],
  "axes": [
    {
      "orient": "bottom",
      "scale": "xscale",
      "title": "Category",
      "encode": {
        "axis": {
          "update": {
            "ariaLabel": {
              "signal": "'x axis showing ' + 'Category' + ' with values ' + domain('xscale')"
            }
          }
        },
        "title": {"update": {"ariaHidden": {"value": true}}},
        "labels": {"update": {"ariaHidden": {"value": true}}}
      }
    },
    {
      "title": "Amount",
      "orient": "left",
      "scale": "yscale",
      "grid": true,
      "encode": {
        "axis": {
          "update": {
            "ariaLabel": {
              "signal": "'y axis showing ' + 'Amount' + ' from ' + domain('yscale')[0] + ' to ' + domain('yscale')[1]"
            }
          }
        },
        "title": {"update": {"ariaHidden": {"value": true}}},
        "labels": {"update": {"ariaHidden": {"value": true}}}
      }
    }
  ],
  "legends": [
    {
      "fill": "qcolor",
      "title": "Amount",
      "encode": {
        "legend": {
          "update": {
            "ariaLabel": {
              "signal": "'legend showing ' + 'Amount' + ' from ' + domain('qcolor')[0] + ' to ' + domain('qcolor')[1]"
            }
          }
        },
        "title": {"update": {"ariaHidden": {"value": true}}},
        "labels": {"update": {"ariaHidden": {"value": true}}},
        "gradient": {"update": {"ariaHidden": {"value": true}}}
      }
    },
    {
      "fill": "ncolor",
      "title": "Categories",
      "encode": {
        "legend": {
          "update": {
            "ariaLabel": {
              "signal": "'legend showing ' + 'Categories' + ' with values ' + domain('ncolor')"
            }
          }
        },
        "title": {"update": {"ariaHidden": {"value": true}}},
        "labels": {"update": {"ariaHidden": {"value": true}}},
        "symbols": {"update": {"ariaHidden": {"value": true}}}
      }
    }
  ],
  "marks": [
    {
      "type": "group",
      "role": "figure",
      "encode": {"enter": {"ariaLabel": {"value": "bars of the chart"}}},
      "marks": [
        {
          "type": "rect",
          "from": {"data": "table"},
          "encode": {
            "enter": {
              "x": {"scale": "xscale", "field": "category"},
              "width": {"scale": "xscale", "band": 1},
              "y": {"scale": "yscale", "field": "amount"},
              "y2": {"scale": "yscale", "value": 0},
              "fill": {"scale": "qcolor", "field": "amount"},
              "ariaLabel": {"signal": "datum.category + ' ' + datum.amount"}
            }
          }
        }
      ]
    }
  ]
}
```

domoritz:
Basic spec to start with

```
{
  "width": 400,
  "height": 200,
  "padding": 5,
  "description": "Bar chart showing amounts per category.",
  "title": {
    "text": "Interesting chart title",
    "subtitle": "An even more interesting subtitle"
  },
  "data": [
    {
      "name": "table",
      "values": [
        {"category": "A", "amount": 28},
        {"category": "B", "amount": 55},
        {"category": "C", "amount": 43},
        {"category": "D", "amount": 91},
        {"category": "E", "amount": 81},
        {"category": "F", "amount": 53},
        {"category": "G", "amount": 19},
        {"category": "H", "amount": 87}
      ]
    }
  ],
  "scales": [
    {
      "name": "xscale",
      "type": "band",
      "domain": {"data": "table", "field": "category"},
      "range": "width",
      "padding": 0.05,
      "round": true
    },
    {
      "name": "yscale",
      "domain": {"data": "table", "field": "amount"},
      "nice": true,
      "range": "height"
    },
    {
      "name": "qcolor",
      "domain": {"data": "table", "field": "amount"},
      "range": "heatmap"
    },
    {
      "name": "ncolor",
      "type": "ordinal",
      "domain": {"data": "table", "field": "category"},
      "range": "category"
    }
  ],
  "axes": [
    {
      "orient": "bottom",
      "scale": "xscale",
      "title": "Category"
    },
    {
      "title": "Amount",
      "orient": "left",
      "scale": "yscale",
      "grid": true
    }
  ],
  "legends": [
    {
      "fill": "qcolor",
      "title": "Amount"
    },
    {
      "fill": "ncolor",
      "title": "Categories"
    }
  ],
  "marks": [
    {
      "type": "group",
      "role": "figure",
      "marks": [
        {
          "type": "rect",
          "from": {"data": "table"},
          "encode": {
            "enter": {
              "x": {"scale": "xscale", "field": "category"},
              "width": {"scale": "xscale", "band": 1},
              "y": {"scale": "yscale", "field": "amount"},
              "y2": {"scale": "yscale", "value": 0},
              "fill": {"scale": "qcolor", "field": "amount"}
            }
          }
        }
      ]
    }
  ]
}
```

domoritz:
```
{
  "width": 400,
  "height": 200,
  "padding": 5,
  "description": "Bar chart showing amounts per category.",
  "title": {
    "text": "Interesting chart title",
    "subtitle": "An even more interesting subtitle"
  },
  "layout": {

  },
  "data": [
    {
      "name": "table",
      "values": [
        {"category": "A", "amount": 28},
        {"category": "B", "amount": 55},
        {"category": "C", "amount": 43},
        {"category": "D", "amount": 91},
        {"category": "E", "amount": 81},
        {"category": "F", "amount": 53},
        {"category": "G", "amount": 19},
        {"category": "H", "amount": 87}
      ]
    }
  ],
  "scales": [
    {
      "name": "xscale",
      "type": "band",
      "domain": {"data": "table", "field": "category"},
      "range": "width",
      "padding": 0.05,
      "round": true
    },
    {
      "name": "yscale",
      "domain": {"data": "table", "field": "amount"},
      "nice": true,
      "range": "height"
    },
    {
      "name": "qcolor",
      "domain": {"data": "table", "field": "amount"},
      "range": "heatmap"
    },
    {
      "name": "ncolor",
      "type": "ordinal",
      "domain": {"data": "table", "field": "category"},
      "range": "category"
    }
  ],
  "axes": [
    {
      "orient": "bottom",
      "scale": "xscale",
      "title": {"signal": "'Category'"}
    },
    {
      "title": "Amount",
      "orient": "left",
      "scale": "yscale",
      "grid": true
    }
  ],
  "legends": [
    {
      "fill": "qcolor",
      "title": "Amount",
      "direction": "horizontal",
      "encode": {
        "legend": {
          "update": {
            "ariaLabel": {"value": "an alternative description of the legend"}
          }
        }
      }
    },
    {
      "fill": "ncolor",
      "title": "Categories",
      "direction": "horizontal"
    }
  ],
  "marks": [
    {
      "type": "group",
      "role": "figure",
      "marks": [
        {
          "type": "rect",
          "from": {"data": "table"},
          "encode": {
            "enter": {
              "x": {"scale": "xscale", "field": "category"},
              "width": {"scale": "xscale", "band": 1},
              "y": {"scale": "yscale", "field": "amount"},
              "y2": {"scale": "yscale", "value": 0},
              "fill": {"scale": "qcolor", "field": "amount"},
              "ariaRole": {"value": "graphics-symbol"},
              "ariaRoleDescription": {"value": "bar"},
              "ariaLabel": {"signal": "datum.category + ' ' + datum.amount"}
            }
          }
        }
      ]
    }
  ]
}
```

packages/vega-parser/src/parsers/guides/guide-util.js
```diff
@@ -86,3 +86,46 @@ export function extendOffset(value, offset) {
     : !isObject(value) ? { value, offset }
     : { ...value, offset: extendOffset(value.offset, offset) };
 }
+
+const DISCRETE_SCALES = new Set(['ordinal', 'band', 'point']);
+
+function domainText(scaleType, scaleName) {
+  return DISCRETE_SCALES.has(scaleType)
+    ? ` with values " + domain('${scaleName}')`
```
donghaoren: How should we support localization for these?

packages/vega-parser/src/parsers/title.js
```diff
@@ -66,7 +66,7 @@ function groupEncode(_, userEncode) {
     limit:      _('limit'),
     frame:      _('frame'),
     offset:     _('offset') || 0,
-    padding:    _('subtitlePadding')
+    padding:    _('subtitlePadding'),
```
donghaoren: Remove extra `,`

packages/vega-parser/src/parsers/title.js
```diff
@@ -94,7 +94,12 @@ function buildTitle(spec, _, userEncode, dataRef) {
     fontSize:   _('fontSize'),
     fontStyle:  _('fontStyle'),
     fontWeight: _('fontWeight'),
-    lineHeight: _('lineHeight')
+    lineHeight: _('lineHeight'),
+    ariaHidden:   _('ariaHidden'),
+    ariaLabel:    _('ariaLabel'),
+    ariaRole:     _('ariaRole'),
+    ariaRoleDescription: _('ariaRoleDescription') || 'title',
+    tabindex:     _('tabindex'),
```
donghaoren: Remove extra `,`

packages/vega-parser/src/parsers/title.js
```diff
@@ -127,7 +132,12 @@ function buildSubTitle(spec, _, userEncode, dataRef) {
     fontSize:   _('subtitleFontSize'),
     fontStyle:  _('subtitleFontStyle'),
     fontWeight: _('subtitleFontWeight'),
-    lineHeight: _('subtitleLineHeight')
+    lineHeight: _('subtitleLineHeight'),
+    ariaHidden:   _('subtitleAriaHidden'),
+    ariaLabel:    _('subtitleAriaLabel'),
+    ariaRole:     _('subtitleAriaRole'),
+    ariaRoleDescription: _('subtitleAriaRoleDescription') || 'subtitle',
+    tabindex:     _('subtitleTabindex'),
```
donghaoren: Remove extra `,`

packages/vega-scenegraph/src/SVGStringRenderer.js
```diff
@@ -330,10 +330,9 @@ function applyStyles(o, mark, tag, defs) {
     if (o.fontWeight) s += 'font-weight: ' + o.fontWeight + '; ';
   }
 
-  for (i=0, n=styleProperties.length; i<n; ++i) {
-    prop = styleProperties[i];
-    name = styles[prop];
-    value = o[prop];
+  for (const prop in styles) {
```
donghaoren: Is this going to break compatibility with older browsers?

packages/vega/test/scenegraphs/scatter-brush-filter.json
```diff
@@ -225,54 +244,65 @@
...
                     }
                   ],
                   "zindex": 0
                 }
               ],
               "x": 0.5,
-              "y": 200.5
+              "y": 200.5,
+              "ariaHidden": null,
+              "ariaLabel": "x axis from Sat Jan 01 2000 00:00:00 GMT-0800 (Pacific Standard Time) to Mon Feb 01 2010 00:00:00 GMT-0800 (Pacific Standard Time)",
```
donghaoren: Looks like the date object format is not very comprehensible here

packages/vega-parser/src/parsers/guides/guide-util.js
donghaoren: I believe we can use `encode` and override the properties there.

packages/vega/test/scenegraphs/scatter-brush-filter.json
domoritz: I should apply the format that is used for the labels here. Good catch.

domoritz:
Todo
* Apply format to labels
* Move default properties to default config

packages/vega-parser/src/parsers/title.js
domoritz: Vega does not enforce trailing commas so I'll ignore this. Like here https://github.com/vega/vega/blob/master/packages/vega-parser/src/parsers/legend.js#L152

packages/vega-scenegraph/src/SVGStringRenderer.js
domoritz: It looks like it's well supported https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in#Browser_compatibility.

packages/vega-parser/src/parsers/guides/guide-util.js
domoritz: Yes, you can override `ariaLabel` for axes. This is just a default.

domoritz: I'll move some defaults to config where it makes sense.

domoritz: I'll move some defaults to config where it makes sense.

packages/vega-parser/src/parsers/axis.js
```diff
@@ -39,12 +39,17 @@ export default function(spec, scope) {
   // encoding properties for axis group item
   axisEncode = extendEncode({
     update: {
-      offset:       encoder(_('offset') || 0),
+      offset:       encoder(_('offset', 0)),
```
domoritz: I believe this is a cleaner way to add defaults but I'm happy to revert if it's not correct.

domoritz:
@jheer Can you take a look at this and let me know what I should change?

domoritz:
todo
* don't output are properties if ariaHidden: true for marks
* consider moving default axis label to Vega-Lite

domoritz:
@jheer Do you think we can include this in the next release? We have made corresponding changes in https://github.com/vega/vega-lite/pull/6185.

packages/vega-parser/src/config.js
```diff
@@ -36,34 +36,62 @@ export default function() {
...
+      stroke: defaultColor,
+      ariaRoleDescription: 'shape'
     },
-    path: { stroke: defaultColor },
-    rect: { fill: defaultColor },
-    rule: { stroke: black },
-    shape: { stroke: defaultColor },
     symbol: {
       fill: defaultColor,
-      size: 64
```
jheer, line 57: Why is this deleted?

packages/vega-parser/src/parsers/axis.js
jheer: The prior version is safer against invalid inputs like empty string or false, but I'm not sure how critical that is.

packages/vega-parser/src/parsers/axis.js
```diff
@@ -36,15 +36,22 @@ export default function(spec, scope) {
...
       position:     encoder(value(spec.position, 0)),
       titlePadding: encoder(_('titlePadding')),
       minExtent:    encoder(_('minExtent')),
       maxExtent:    encoder(_('maxExtent')),
-      range:        {signal: `abs(span(range("${spec.scale}")))`}
+      range:        {signal: `abs(span(range("${spec.scale}")))`},
+      ariaHidden:   encoder(ariaHidden),
+      ariaLabel:    encoder(ariaHidden == true ? undefined : _('ariaLabel', axisAriaLabel(_, scope))),
```
jheer, line 51: - Should either check to be truthy or use triple (not double) equals comparison.
- The ternaries can be outside the encoder calls: `ariaHidden ? undefined : encoder(_('ariaLabel', axisAriaLabel(_, scope)))`. Perhaps a cleaner approach would be to consolidate conditional logic for ariaHidden. One could ideally use object rest destructuring, but that might introduce compatibility issues with certain browsers.

packages/vega-parser/src/parsers/guides/guide-util.js
```diff
@@ -1,6 +1,7 @@
 import {Left, Right, Center, Start, End, Vertical} from './constants';
 import {value} from '../../util';
-import {isObject, stringValue} from 'vega-util';
+import {isObject, stringValue, isArray} from 'vega-util';
```
jheer, line 3: isArray should appear first in this import list

packages/vega-parser/src/parsers/guides/guide-util.js
```diff
@@ -86,3 +87,80 @@ export function extendOffset(value, offset) {
     : !isObject(value) ? { value, offset }
     : { ...value, offset: extendOffset(value.offset, offset) };
 }
+
+const DISCRETE_SCALES = new Set(['ordinal', 'band', 'point']);
```
jheer, line 91: What about in BinOrdinal? Can we use `isDiscrete` from vega-scales instead?

packages/vega-parser/src/parsers/guides/guide-util.js
```diff
@@ -86,3 +87,80 @@ export function extendOffset(value, offset) {
...
+
+function formatType(_, scaleType) {
+  const formatType = _('formatType');
+
+  if (formatType) {
+    return formatType;
+  }
+  if (scaleType === Time) {
+    return 'time'
```
jheer, line 100: Time === 'time' and UTC === 'utc', no?

packages/vega-parser/src/parsers/guides/guide-util.js
```diff
@@ -86,3 +87,80 @@ export function extendOffset(value, offset) {
...
+  if (scaleType === UTC) {
+    return 'utc'
+  }
+}
+
+function wrapFormat(text, format, formatType) {
+  if (format == null) {
+    return text
+  }
+  const func = formatType === "time" ? "timeFormat" : formatType === "utc" ? "utcFormat" : "format";
```
jheer, line 111: Should use single quotes

packages/vega-parser/src/parsers/guides/guide-util.js
jheer: Also, these lines are missing trailing semi-colons

packages/vega-parser/src/parsers/guides/guide-util.js
```diff
@@ -86,3 +87,80 @@ export function extendOffset(value, offset) {
...
+
+  if (formatType) {
+    return formatType;
+  }
+  if (scaleType === Time) {
+    return 'time'
+  }
+  if (scaleType === UTC) {
+    return 'utc'
+  }
```
jheer, line 104: Could this method return `undefined`? If so, what happens subsequently? Should there be an error condition here?

packages/vega-parser/src/parsers/guides/guide-util.js
```diff
@@ -86,3 +87,80 @@ export function extendOffset(value, offset) {
...
+        format,
+        formatType
+      )} + " to " + ${wrapFormat(
+        `domain("${scaleName}")[1]`,
+        format,
+        formatType
+      )}`;
+}
+
+function titleText(_) {
```
jheer, line 129: Consider passing in the title value instead of the parameter object? It would make this method a clearer stand-alone utility.

packages/vega-parser/src/parsers/guides/guide-util.js
```diff
@@ -86,3 +87,80 @@ export function extendOffset(value, offset) {
...
+  let title = _('title')
+
+  if (isArray(title)) {
+    // only use the first row of multiline titles
+    title = title[0];
+  }
+
+  if (title) {
+    return isObject(title) ? ` showing " + ${title.signal} + "` : ` showing ${title}`;
```
jheer, line 138: What if a non-signal object is passed in?

packages/vega-parser/src/parsers/guides/guide-util.js
```diff
@@ -86,3 +87,80 @@ export function extendOffset(value, offset) {
...
+  }
+
+  return '';
+}
+
+export function legendAriaLabel(_, scope) {
+  const domains = [];
+  for (const legendType of ['fill', 'stroke', 'opacity', 'size', 'shape', 'strokeDash']) {
```
jheer, line 146: Instead, please use LegendScales exported by constants.js

packages/vega-parser/src/parsers/guides/guide-util.js
```diff
@@ -86,3 +87,80 @@ export function extendOffset(value, offset) {
...
+  }
+  const signal = `"legend${titleText(_)}${domains.join(',')}"`;
+  return { signal };
+}
+
+export function axisAriaLabel(_, scope) {
+  const scaleName = _('scale');
+  const scaleType = scope.scales[scaleName].params.type;
+  const orient = _('orient')
+  const axisType = orient === 'bottom' || orient === 'top' ? 'x' : 'y';
```
jheer, line 162: Again, please use defined constants in constants.js

packages/vega-parser/src/parsers/guides/guide-util.js
```diff
@@ -86,3 +87,80 @@ export function extendOffset(value, offset) {
...
+    }
+  }
+  const signal = `"legend${titleText(_)}${domains.join(',')}"`;
+  return { signal };
+}
+
+export function axisAriaLabel(_, scope) {
+  const scaleName = _('scale');
+  const scaleType = scope.scales[scaleName].params.type;
+  const orient = _('orient')
```
jheer, line 161: semi-colon

packages/vega-parser/src/parsers/guides/guide-util.js
```diff
@@ -86,3 +87,80 @@ export function extendOffset(value, offset) {
...
+        formatType
+      )} + " to " + ${wrapFormat(
+        `domain("${scaleName}")[1]`,
+        format,
+        formatType
+      )}`;
+}
+
+function titleText(_) {
+  let title = _('title')
```
jheer, line 130: semi-colon

packages/vega-parser/src/parsers/legend.js
```diff
@@ -150,6 +152,11 @@ function buildLegendEncode(_, config) {
     strokeDash:   config.strokeDash,
     x:            _('legendX'),
     y:            _('legendY'),
+    ariaHidden:   ariaHidden,
```
jheer, line 155: Same issues with ariaHidden-conditional logic noted previously. Since this pattern is repeated, perhaps it would be better to abstract these to a dedicated aria helper, and pass the config object through that helper (which could then annotate the config object).

packages/vega-parser/src/parsers/title.js
```diff
@@ -94,7 +96,12 @@ function buildTitle(spec, _, userEncode, dataRef) {
     fontSize:   _('fontSize'),
     fontStyle:  _('fontStyle'),
     fontWeight: _('fontWeight'),
-    lineHeight: _('lineHeight')
+    lineHeight: _('lineHeight'),
+    ariaHidden:   ariaHidden,
```
jheer, line 100: See above comments about abstracting aria annotations for config objects.

packages/vega-parser/src/parsers/legend.js
jheer: Maybe something like this?

```js
function addAriaAnnotations(spec, _, props) {
  const hidden = _(props.hidden);
  const aria = hidden === true
    ? {
        ariaHidden: hidden
      }
    : {
        ariaHidden: hidden,
        ariaLabel: _(props.label),
        ariaRole: _(props.role),
        ariaRoleDescription: _(props.roleDescription),
        tabindex: _(props.tabindex)
      };
      
  return Object.assign(spec, aria);
}
```

packages/vega-scenegraph/src/SVGRenderer.js
```diff
@@ -529,6 +540,20 @@ prototype._update = function(mdef, el, item) {
   // apply svg attributes
   mdef.attr(emit, item, this);
 
+  // optimization: if ariaHidden is true, we can remove other aria properties
+  if (item.ariaHidden) {
```
jheer, line 544: Is this necessary? `delete` can be a relatively costly operation. Another option might be to set these to null or undefined if it is necessary to suppress the values.

packages/vega-scenegraph/src/SVGRenderer.js
```diff
@@ -414,6 +417,14 @@ function siblingCheck(node, sibling) {
 var element = null, // temp var for current SVG element
     values = null;  // temp var for current values hash
 
+var general_extras = {
```
jheer, line 420: Could name this aria_extras for now, to be more specific and clear.

packages/vega-scenegraph/src/SVGRenderer.js
jheer: On second thought, this might be buggy. What if ariaHidden is a signal defined in the update set, when all the other properties are constants in the enter set? I think that could result in inaccurate output here. We may need to condition lookup of other aria properties on the current value of ariaHidden.

packages/vega-schema/src/axis.js
```diff
@@ -59,6 +59,11 @@ const axis = object({
   translate: numberType,
   values: arrayOrSignal,
   zindex: numberType,
+  ariaHidden: booleanValue,
```
jheer, line 62: Not something we need to do now, but it might be helpful in the future to define repeated properties (like these aria properties) in an object bundle and use object rest destructuring to insert them where needed. There would be no major compatibility issues as this is part of the compile process only.

packages/vega-typings/types/spec/encode.d.ts
```diff
@@ -252,7 +252,25 @@ export interface EncodeEntry {
   blend?: ProductionRule<ScaledValueRef<Blend>>;
   cursor?: ProductionRule<StringValueRef>;
   tooltip?: ProductionRule<StringValueRef>;
+
```
jheer, line 255: Why is there a new line break here?

jheer:
Thanks @domoritz for all your work here! I added lots of individual comments to consider. In addition, some other higher-level comments:

- I wonder how much the scenegraph (and, secondarily, output SVG) get bloated by the aria properties. For very large visualizations, does this have an appreciable performance impact? I note that memory pressure could be increased (perhaps significantly?) even when using the non-accessible Canvas renderer. Given that the properties are typically constant for all mark items of a given type, is there a more efficient way for aria information to get passed to the renderer?
- Should it be possible to disable aria property output? Basically, functioning like a feature flag?
- The ARIA properties are applied to SVG only. This makes sense, in so far as I don't know how to leverage them effectively in Canvas. However, `tabindex` is perhaps a more complicated case, as it changes the interaction model, leading to inconsistency. Probably a problem we can live with, but I'm wondering how far you've thought this through.
- On a similar note, how does `tabindex` interact with existing interaction features, such as enabling or disabling interaction events (e.g., `interactive: false`)? Have you tested this (manual tests are fine)?

packages/vega-parser/src/parsers/axis.js
domoritz: In the end, the new code uses this to use the default so yes, empty strings or false would be passed to offset but those are invalid values so 🤷‍♂. I guess there is no point in changing this so I'll revert. 

```
export function value(specValue, defaultValue) {
  return specValue != null ? specValue : defaultValue;
}
```

domoritz: The legend encode code doesn't seem to handle default values at all.

```js
addEncoders(encode, {
    orient:       _('orient'),
    offset:       _('offset'),
    padding:      _('padding'),
    titlePadding: _('titlePadding'),
    cornerRadius: _('cornerRadius'),
    fill:         _('fillColor'),
    stroke:       _('strokeColor'),
    strokeWidth:  config.strokeWidth,
    strokeDash:   config.strokeDash,
    x:            _('legendX'),
    y:            _('legendY'),
    ariaHidden:   ariaHidden,
    ariaLabel:    ariaHidden == true ? undefined : _('ariaLabel', legendAriaLabel(_, scope)),
    ariaRole:     ariaHidden == true ? undefined : _('ariaRole'),
    ariaRoleDescription: ariaHidden == true ? undefined : _('ariaRoleDescription'),
    tabindex:     _('tabindex')
  });
```

packages/vega-parser/src/config.js
domoritz: Good catch. Reverted.

packages/vega-parser/src/parsers/guides/guide-util.js
domoritz: Ahh, thanks. I didn't think of `isDiscrete`.

packages/vega-parser/src/parsers/guides/guide-util.js
domoritz: Yes. I like writing

```
if (x === 1) {
	return true;
}
```

instead of

```
return x === 1;
```

domoritz: Simplified to 

```
function formatType(_, scaleType) {
  return _('formatType') || scaleType;
}
```

I think there is value in having this in a function.

packages/vega-parser/src/parsers/guides/guide-util.js
domoritz: Done. Enforced in https://github.com/vega/vega/pull/2514.

packages/vega-parser/src/parsers/guides/guide-util.js
domoritz: It's now

```
function formatType(_, scaleType) {
  return _('formatType') || scaleType;
}
```

packages/vega-parser/src/parsers/guides/guide-util.js
domoritz: This crashes and burns. Is it ever expected to happen, though? Should we default to printing `[Object]` instead of crashing?

packages/vega-parser/src/parsers/guides/guide-util.js
domoritz: Ahh, I'm learning more about Vega. Thanks.

packages/vega-parser/src/parsers/guides/guide-util.js
domoritz: I'll add `isSignal(title)` to the condition.

packages/vega-parser/src/parsers/guides/guide-util.js
domoritz: I'm so used to just type and have prettier add all the semicolons. I've become complacent to code style.

packages/vega-parser/src/parsers/guides/guide-util.js
domoritz: Done

packages/vega-scenegraph/src/SVGRenderer.js
domoritz: I'm happy to just remove this. I assumed that there would be value in removing things from the generated SVG.

packages/vega-typings/types/spec/encode.d.ts
domoritz: No real reason. Removed.

packages/vega-parser/src/parsers/legend.js
domoritz: Done. I needed a bit more for the function to generate the default label.

domoritz: I honestly got confused by the code. The legend code uses `addEncoders` while the axis code uses `extendEncode` and expects every property to be passed through and `encoder`.

Can you help me?

domoritz: I didn't really know what do with the encoder so I'm just passing it as an argument. This feels wrong. Can you help me?

packages/vega-parser/src/parsers/axis.js
domoritz: handled below together with legend

domoritz:
> I wonder how much the scenegraph (and, secondarily, output SVG) get bloated by the aria properties. For very large visualizations, does this have an appreciable performance impact? I note that memory pressure could be increased (perhaps significantly?) even when using the non-accessible Canvas renderer. Given that the properties are typically constant for all mark items of a given type, is there a more efficient way for aria information to get passed to the renderer?

I'm sure the scenegraph and SVG gets bloated. I did not focus on performance in this pull request since the defaults are only applied to axes and legends.

Yes, we could apply constant properties to the group instead of individual properties (like the ticks, labels, etc). that should make things a lot more efficient. I didn't feel confident making these changes in Vega since there are probably some subtleties I am missing (like how do we know that a property is actually constant when it can be set in both enter and update?).

Is this something you could help with?

Another thing we may be able to do is to consolidate aria-role and role a bit since they serve similar roles (no pun intended) but I'm not sure we can do this without breaking changes.

> Should it be possible to disable aria property output? Basically, functioning like a feature flag?
The ARIA properties are applied to SVG only. This makes sense, in so far as I don't know how to leverage them effectively in Canvas. However, tabindex is perhaps a more complicated case, as it changes the interaction model, leading to inconsistency. Probably a problem we can live with, but I'm wondering how far you've thought this through.

I assume you mean a feature flag to completely disable/enable aria support in Vega. It would be a shame though if accessibility was opt-in. Maybe this can come in a follow up pull request? Would this be purely for performance reasons? 

One option I thought about was to move the guide defaults to Vega-Lite and not output any accessibility information in Vega. But then that would mean a lot more work to make a Vega chart accessible. Not ideal.  

Why is tabindex more complicated?

> On a similar note, how does tabindex interact with existing interaction features, such as enabling or disabling interaction events (e.g., interactive: false)? Have you tested this (manual tests are fine)?

I have no tested in. Why would there be issues? What would you want me to test? 

I'm also happy to just remove tabindex from this pull request completely since it's not actually needed for aria (aria ignores the tab order and goes in dom order anyway).

packages/vega-parser/src/parsers/guides/axis-grid.js
```diff
@@ -15,7 +15,10 @@ export default function(spec, config, userEncode, dataRef, band) {
       encode, enter, exit, update, tickPos, u, v, v2, s;
 
   encode = {
-    enter: enter = {opacity: zero},
+    enter: enter = {
+      opacity: zero,
+      ariaHidden: {value: true}
```
domoritz, line 20: I think this could be moved to the group. What's the best way to do that?

jheer:
I’m happy to help move to group level annotations when I can find the time. Starting with internally generated and perhaps also mark-level (not item-level) spec/config properties makes sense.

If we added a feature flag it would be for opt-out. I too want this on by default! Most likely we just include it always assuming we sidestep any significant performance issues.

It may be best to separate out tabindex for now. It would be ideal (though maybe difficult) to “simulate” tabindex support in the Canvas handler.

domoritz:
> It may be best to separate out tabindex for now. It would be ideal (though maybe difficult) to “simulate” tabindex support in the Canvas handler.

Done.

domoritz:
I've rewritten the logic for adding aria annotations for axes, legends, and titles using your suggestion. However, I am not using a `addAriaAnnotations` for axes and legends since axes use `encoder` and both have custom logic to compute default aria labels.

packages/vega-parser/src/parsers/legend.js
domoritz: I'm using the shorthand for titles and subtitles now. I couldn't figure out a clean way to make it work for guides since their implementations differ (to me) significantly.

domoritz:
I had a simple idea. Instead of passing aria hidden through the scenegraph for guide groups, I changed the SVG renderer to add aria hidden to the guide groups based on the group role. This means we can omit a bunch of `ariaHidden` from the scenegraph and only take a tiny performance hit in the renderer.

Does this look good to you @jheer?

packages/vega/test/scenegraphs/bar-time.json
```diff
@@ -207,7 +207,10 @@
                         }
                       ],
                       "x": 0.5,
-                      "y": 200.5
+                      "y": 200.5,
+                      "ariaLabel": "x axis with values Sun Jan 01 2012 00:00:00 GMT-0800 (Pacific Standard Time),Mon Jan 02 2012 00:00:00 GMT-0800 (Pacific Standard Time),Tue Jan 03 2012 00:00:00 GMT-0800 (Pacific Standard Time),Wed Jan 04 2012 00:00:00 GMT-0800 (Pacific Standard Time),Thu Jan 05 2012 00:00:00 GMT-0800 (Pacific Standard Time),Fri Jan 06 2012 00:00:00 GMT-0800 (Pacific Standard Time),Sat Jan 07 2012 00:00:00 GMT-0800 (Pacific Standard Time)",
```
jheer, line 211: We need nicer default aria text descriptions of `Date` objects.

packages/vega/test/scenegraphs/arc.json
```diff
@@ -21,7 +21,9 @@
               "startAngle": 0,
               "endAngle": 0.474202664692799,
               "innerRadius": 20,
-              "outerRadius": 58.43075691322092
+              "outerRadius": 58.43075691322092,
+              "ariaRole": "graphics-symbol",
```
jheer, line 25: While the latest revisions have helpfully limited the aria annotations produced for guides, we can still generate a lot of annotations, one for each standard mark. Bloat issues aside, is this really what we want to do? Is this what is most helpful?

An alternative might be to emit standard aria annotations not for each mark scenegraph item, but once for the mark group itself. In other words: put the annotation on the mark parent in the scenegraph (not the item entries) and then have the renderers write this annotation to the enclosing SVG `g` element, rather than the individual elements (e.g., `path`, `text`, etc.) corresponding to mark items.

Would that make sense to do?

packages/vega/test/scenegraphs/area.json
```diff
@@ -780,7 +788,9 @@
               "fill": "steelblue",
               "fillOpacity": 1,
               "interpolate": "monotone",
-              "y2": 200
+              "y2": 200,
+              "ariaRole": "graphics-symbol",
+              "ariaRoleDescription": "area"
```
jheer, line 793: Adding aria annotations to every item makes even less sense for interpolated marks like area, line, and trail, which map to a single SVG path element. Regardless of how it is ultimately rendered out, it seems better to put information that is identical for mark items in the mark parent element of the scenegraph.

packages/vega/test/scenegraphs/bar-time.json
domoritz: Yes. I think this is not possible with Vega right now. We currently use this aria label: `x axis represents values " + domain('xscale')`. The format is `timeFormat(VALUE, timeUnitSpecifier(tbin.units))` but I don't think we have a way to map the values with the formatting function right now. We would need a `map` expression function, right?

packages/vega/test/scenegraphs/arc.json
domoritz: I think that makes sense for line, trail, and area marks. For rect marks or text marks, there is value for a screen reader user to be able to navigate the individual items. How do you think we can get the best of both worlds?

packages/vega/test/scenegraphs/area.json
domoritz: Right. I think in this case the default is just telling a screen reader user that there is a line. It's on the chart creator to add points that are accessible.

So far, I have not focused much on making the marks accessible by default in Vega since it's not always clear what the right description is. In my Vega-Lite pull request, I am adding automatic labels for marks and I think Vega-Lite could automatically generate accessible lines (it does not right now). 

We could do this by default in Vega, though, which I think would be good but also increase the bloat. 

Maybe one option here (as you suggested above) would be to come up with some way to summarize the line in a single description. I just don't know of a good way to do that.

domoritz:
I updated the default guide description based on feedback from a screenreader user. I'll get more feedback on the current state.

domoritz:
I'm deploying the editor with this branch at https://vega.github.io/editor-a11y/ from time to time so we can get feedback.

domoritz:
[Open the Fruit Chart in the Accessible Vega Editor](https://vega.github.io/editor-a11y/#/url/vega/N4KABGBEAkDODGALApgWwIaQFxUQFzwAdYsB6UgN2QHN0A6agSz0QFcAjOxge1IRQyUa6SgFY6AK1jcAdpAA04KAHdGAExbYwAFgAMuxREgpG1fFoBM+w1ELo1axjOpbRNyGuQIATo0J4eORxIAEEwdnRvMCRIvDBuADMwBO9WZgUlSAC8ABtkLVAIIzxkAA88LVDwyOjEWPik9Ap0Rhz0djzk1OYwJzAWZDBYVG5uFkZB2ERuQkglAF93NXQ8TBwAbSUIQqKjGXRUfODVjvybXchmnNYvLU3d3eBIfcPKwB4NwDl9hSh4blYZCo4ADsiy2DyeLyOUDegHl976QX7-QFgACMulBDyKEIOUMgb0AMvvwxEArQANgxmLA2NewTegAV9ol-Ek4ADMFMx1Nxb0A0vuMpGWdng54496AKX2+cywOTzkLIe9AOL7EuRFkFj2FNOhgEl9pVaFHzMEQAC6YP1RvcCHQeVgdzBOwucuCpQteQymKyAE9CLiIjI1K6Hh5uBgnAUPCs1lATi75FAEhMcn7gnLVVBvOhnLjVBpEP6LnYHE4XDhdHRdG4DammYn+qlkCaZXaiurce7nWcK4Hg0EqWHVpUo2dY-HqwimRUU89GPAoXhazKjGmM5UTGYKialIb3OhSrcNraO9xfMhJZB2GM8EHc0225UnfBLe23dkXTgnrBTPscpUAOQAMW6eDfpAppFIKjbFMwL5QAAwmOV5GIeEwnnkCQVPOUA3sErb3tGHbUL41azjc65mpkGDeAA1tae6PB2eCeri+F-LM6GQN43BQZAcbUKw3iPgGx6-J4BQVkYx4lN4ImUk2kSMOgAAy7TIF+r6XJaNyVDE3hxMsfYgZi+mgax5FUTa0ngQG9FepUfHwGholNikl6qbpEZZO0LophcgncMJr4Od5ALIJJ-nSW6pShphUB3g+8JxspI7Juh0mQFmmiqVFkAxdGUA+tWerJZSkDupFOEtjeMZccOmlwV5RXuhYpWxVhFVQFcGnFnVbqyegABKHFQk87WMWmhCIFOsAALSwO6qBnl+XUBj1-V5AAIl48C+P4gShsNlQRJJi0XD1insMpkUfpalS6awqB0JCYAANRgN+L1PWAN13cS44BaBv2GbsANgMagMbiA8xAA)

domoritz:
Looks like the svg string renderer doesn’t produce annotate charts yet. I’ll take a look tomorrow.

kanitw:
We'll merge this into `next` branch first. We can further revise it before merging this into master.

majornista:
Links to charts with WAI-ARIA:

[Open the Fruit Chart in the Accessible Vega Editor](https://vega.github.io/editor/#/url/vega/N4KABGBEAkDODGALApgWwIaQFxUQFzwAdYsB6UgN2QHN0A6agSz0QFcAjOxge1IRQyUa6SgFY6AK1jcAdpAA04KAHdGAExbYwAFgAMuxREgpG1fFoBM+w1ELo1axjOpbRNyGuQIATo0J4eORxIAEEwdnRvMCRIvDBuADMwBO9WZgUlSAC8ABtkLVAIIzxkAA88LVDwyOjEWPik9Ap0Rhz0djzk1OYwJzAWZDBYVG5uFkZB2ERuQkglAF93NXQ8TBwAbSUIQqKjGXRUfODVjvybXchmnNYvLU3d3eBIfcPKwB4NwDl9hSh4blYZCo4ADsiy2DyeLyOUDegHl976QX7-QFgACMulBDyKEIOUMgb0AMvvwxEArQANgxmLA2NewTegAV9ol-Ek4ADMFMx1Nxb0A0vuMpGWdng54496AKX2+cywOTzkLIe9AOL7EuRFkFj2FNOhgEl9pVaFHzMEQAC6YP1RvcCHQeVgdzBOwucuCpQteQymKyAE9CLiIjI1K6Hh5uBgnAUPCs1lATi75FAEhMcn7gnLVVBvOhnLjVBpEP6LnYHE4XDhdHRdG4DammYn+qlkCaZXaiurce7nWcK4Hg0EqWHVpUo2dY-HqwimRUU89GPAoXhazKjGmM5UTGYKialIb3OhSrcNraO9xfMhJZB2GM8EHc0225UnfBLe23dkXTgnrBTPscpUAOQAMW6eDfpAppFIKjbFMwL5QAAwmOV5GIeEwnnkCQVPOUA3sErb3tGHbUL41azjc65mpkGDeAA1tae6PB2eCeri+F-LM6GQN43BQZAcbUKw3iPgGx6-J4BQVkYx4lN4ImUk2kSMOgAAy7TIF+r6XJaNyVDE3hxMsfYgZi+mgax5FUTa0ngQG9FepUfHwGholNikl6qbpEZZO0LophcgncMJr4Od5ALIJJ-nSW6pShphUB3g+8JxspI7Juh0mQFmmiqVFkAxdGUA+tWerJZSkDupFOEtjeMZccOmlwV5RXuhYpWxVhFVQFcGnFnVbqyegABKHHIAAIl48C+P4gShu13qRMBhXdb4ClKSpPbvtQn6VLprCoHQxJxAA1GA36HWAB2bdtyYBaBl2GbsN1gMat0biA8xAA)
[Open the Weather Chart in the Accessible Vega Editor](https://vega.github.io/editor#/url/vega/N4KABGBEAkDODGALApgWwIaQFxUQFzwAdYsB6UgN2QHN0A6agSz0QFcAjOxge1IRQyUa6SgFY6AK1jcAdpAA04KIXQATVYxnVsYUYoiQA7o1UsdAJgAMl-bmSNq+C9duRVyBACdGhPDzk4kACyyOgyYHhohMie6HisnshgYapghInwPsxx-mAAZtyeyQA2xWCosiywCkpucZg4ANpKEKAQ7VAy6KjIOnV4mLYdkAnFfar1fKEExcgAtIbTKJ508LAUNR0GBZ4YeDrAkHgAntF9axvyyuiesL04hxOR43G9AL5vQ+1HsTKwO6gdM0tu02iDvnlGMhiqoXs8viCjqd7lA-D1WDJmJtwQYMcxqk1IBUZGYALoIraQdAEsCNImVRAAfSevSu9JJTJZjOQMlhpJaW0+Ao6YJxSLOgXQ1GoiVo8OFlJl3FYhHYxyB7JYzNekHJCuG3GIGp6YQUUBNcj1OO2UJhNLpkVQhEZGAAHmbIOlkJlCNk-LJdRThtTjaEZIzHc63R6LYyvT6-f5dfqwG8FfyOmmIFbzTcANb2gWi75dHp9DCeAvY4YnCVQWDHVDsbhjCmQPKebiAh79BpQJ6YIWInnwbjuA4pkaEFkT61QSGlA6QCjoYqsFGQADEABZ4AB2AAc6APkCH1sg7p7CFXG-dbMh0NhgVjkZd6HdZ7F6qv8BvfXV962k+5phnGGRZAM-pyJ+4JUt46AAEotsgAAiHjwN4vhJj2K5rhuhDcJo+wwYiNyMOgAAy6DsNCS6wA4XRjIEaLIAAYoUewABRPKwqB0MSWoslcADkACkABCwkAJRgAA1GAwkiXJ+QcXE3FxLx-Gga+boiXQACMeTScpwlgO4MrIB4YDwNC9GsLAyS8gpykAmpPF8bG8YQTksgiZYBlGTJ8mmTZJKMD0kS3MJp5Bt8eDcC2fiEHRDGrn0LHsbsbkaR5DLapEIkSdJp4plmgoCmVOaQNesyFiKCqQKWt7VnFyJ9MUmihJ4LUGKoXboJoS4DnCgzzkB5baVEb4fm2vzUCijQ2GAhz0dQjF9MYpiIKeObDJiNk6Hgnjrm2ABeMTcIdx3IBVXzFgYTX-j1qJtYEHUyF1z1uP1g09sNgTDYBj4TWEYHet5UExQ1c0LStqVMXYDhOJ8YCWLtJaMAdOBHSdDXnZ2V3rhVShVe+HhAkWDU1c1baFFCJJ9M2BBdl9MomDoeSrncbZ+Hgsx9CEYQRFEMQaYkYCaNZPJ+DKajk22xTUdCrFrrA23Y9dCtK8UADyVCeIryUa7jiJ+PAeYAMLKgzV7w+c9jFJxm0sKQ26WFJUOIqdmjuJeqO3ZTiLU09tPeNL7XIHk+xtmzwGc8U3MNbz-OBIL4Reb6kG5JLoUsZF1RazRuv64bhPIDzmOW9b+y22taWBDZjCOygSN4K77ue5S3u8sgfuWMT2YgG8QA)

majornista:
Updated example using `description` instead of `ariaLabel` to render the `aria-label` attribute on each `path`.

[Open the Fruit Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykSArJQBWENgDsQAGhAB3GgBN6aACwAGNTPg0yWNACYNMnEgUKa4smgEyFcaACcaOTO0noAggAImSB19g-TC82ADMvUIcGOmkQV0wAGzg0UEw4AA9MNBBvX39Ah2CwryRiJBoEpCYkiKi6LwsvejgvCAQ2NnoaFogYNhwQAF9bJEwUVABtUHEkBGT0MerkmTKEhns0KZAZuezAHg3AOX3YqDYGcSzUAHZh6dn5kD3AeX3j0-O0AEY1G+27-cAZfZeZwuADZvjt7ntAAr7gLeqAAzGDfug9oBpfZhF30iN2yMAUvvotCgqS3bEPQDi+-jUJiiT8SXtAJL7FPegwAuiyZNAkEkIJtifd0hykrFMABPHD3XziBSxBRsZAWFIgBSjcZxKqCmShboJKXocFDGQOJCWe7yJQwWImMwWKyoNSUNQ2EAOV46zBROBY+7CgXLRWy8ruUBKsbZRbqkCauDa7InIH67Y0KDzN3rA1Gsj3bS6LJskBIdIbSagNhOOCwkBMTqYWWxH3ZflQTm++KC1CgCA6GYJbIAcgAYnVMD2ht8W-cAMKvLIyEvdctJULTkB19DexvhshOV3u3PIBwAa25Rbiovum9OAwNbFbEZ0DAcvrLJzsCrLaQcCr8NCQABkqlGFVWdYYxgIIvGDFBBigmQ90PHkTzFbIHygJdIhrNtFWVUM1WSYYQCfNgXwwt84A-DD0gVFcQAbJtYkjaNdV+PDTWUDCqJo8MJR1JkZGFSj1y9OsNS1HUQFjN48OFfR+No1chJAID5i+GQvyQAAla84AAEXsKAnBcNxAM5YD0DyeM7EcZxXAkSjO05bJgwYBBKHE4IAGovB7TyvA8xznL1KCoNZVkgA)
[Open the Weather Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykSArJQBWENgDsQAGhA4kAE3k1xZNAJkB3GvPpoATAAYDM+DTJZ9RmfLjQATjRyZ2k9AFk4ScQAJMiHHB2SJgMdnDeXvLeOGFQjnTBLt4AZmx2EQA2Gd4IEvQQ0iDywSioANqg4kgIcGhFJYWhGXXFmPwQntgZcAC0Gp3wdpRQEMSFqXbImGigmACeAXUjYzJydh0z9X4twbUAvnsymEHiEBMIaBUgyTRwGfI720cLtejONQzidI1fmAXlIFy4l0AF0ZEh-mVAXkYAB9Vq1GRA+jw3awuDiB4gw6zF51JBkMhhChPEBEtgMHBMOaXaHAuEIkBgkBsPC0mpeQocyTMm53eSQkB+BA4WHIAAehRicDiOASzgkTPBgu5sOFoolXM84lh0tl8pcTL22OZyDsAGtIZVqq9AUgLQVnot0BA5ggmGxmjJknY2BdUKBWqUtihDiAMVA2DZNpTGQHrjQsptiEgMgxbQBiAAsUAA7AAOJD5kBhyXx6Cp22S723e51VXqsVISVhmnlqCVuo0mv8+va3WxeJtBWSMP2mhIABKnrgABFbFAHE5DfGU2nbTg2MppmGbPZHCPNhAzFVmm8aDUAGJpKYAClaDAQlGRDN2Um8AHIAKQAIQ-AEpvAAak-d8P2AlIb2Ce9gkfZ9+0bCUwMoABGZIAIg8CbCJOBbG8KA7mPBgIAiTFPwg85oIfJ9VT1IdEgkMCDFQ9DAJA8CCOBC84D8dYPxLI42E9ZwcCPE9Uzqd44GvSYqNgmiYVRPwwN-ACSwOY0ZArborRAKoajqashTxdAMmUTw7EKeQ-SQZRNiDR4UB7Ot0AbfwmxbGQTjIV4ymMUBjzIU86i0HQYBLZkvgItBjnTGQAC9AjYGK7HTHE9JtLtCnmZ0QDM8QLKsmy7PjBz0Ac5yHlc-s6LlYdDTDbzfIC8SzwwOAzAsQ4DEimhotQWK4ASpKUrS5lm1sS4Ao7bpDMKNJbmBOoPWwP1CiJbQ0GSVMOiOOhZvcbVfH8QJYLCbxlHwjFnCJBRJpkDIkCYO5LzTCBwoG1Khryp67gAeVIOxHtEz64qFPrzQAYQpJby1apYOoyW9QvoAgswMf8BJAeLlBsMsDHS7TbW7FkHGuupumSaYZA2qrtoyXbwcwA6QA8LxokHOqGJ8S7OKk3jHR+56MgBwJgdG77nCgKGYemOGgok9ACMTW9THMTA0YxrGccxOB8eNPYgA)

PR #2387, opened 2020-02-19
fix(vega-typings): nice can be time interval with step

kanitw:


packages/vega-typings/types/spec/scale.d.ts
```diff
@@ -161,9 +161,12 @@ export type TimeInterval =
   | 'week'
   | 'month'
   | 'year';
+
+export type TimeIntervalStep = { interval: string; step: number };
```
domoritz: Instead of a string, let's make this more specific: https://github.com/vega/vega/blob/2741da3a96d4eee50069adaf6f5133ec94d46078/docs/vega-schema.json#L4696

PR #2388, opened 2020-02-19
refactor(vega-typings): avoid object-based union types for scale

kanitw:
Merge https://github.com/vega/vega/pull/2387 first

### DIFF
--- a/packages/vega-typings/types/spec/scale.d.ts
+++ b/packages/vega-typings/types/spec/scale.d.ts
@@ -56,14 +56,29 @@ export type UnionSortField =
     };
 export type ScaleField = string | SignalRef;
 
-export type ScaleBins =
-  | (number | SignalRef)[]
-  | SignalRef
-  | {
-      step: number | SignalRef;
-      start?: number | SignalRef;
-      stop?: number | SignalRef;
-    };
+export interface ScaleBinParams {
+  /**
+   * The step size defining the bin interval width.
+   */
+  step: number | SignalRef;
+
+  /**
+   * The starting (lowest-valued) bin boundary.
+   *
+   * __Default value:__ The lowest value of the scale domain will be used.
+   */
+  start?: number | SignalRef;
+
+  /**
+   * The stopping (highest-valued) bin boundary.
+   *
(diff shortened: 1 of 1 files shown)


packages/vega-typings/types/spec/scale.d.ts
```diff
@@ -78,21 +79,23 @@ export type ScaleInterpolate =
...
-      fields: ScaleField[];
-    }
-  | {
-      fields: ((string | number | boolean)[] | DataRef | SignalRef)[];
-    };
-export type ScaleData =
-  | (DataRef & { sort?: SortField })
-  | (MultiDataRef & { sort?: UnionSortField });
+
+export type ScaleMultiDataRef = {
```
domoritz: Why not multi scale data ref?

kanitw: It's single scale, not multi scale?  I'm also just treating scale as prefix here.

domoritz:
I suppose you made this change so you can import the individual types in Vega-Lite, right?

kanitw:
> I suppose you made this change so you can import the individual types in Vega-Lite, right?

Yes, and even if we import the main one, we want each of them to be named.  Otherwise, Altair won't generate a class and we can't refer to any of these objects in the docs.

packages/vega-typings/types/spec/scale.d.ts
```diff
@@ -55,44 +56,64 @@ export type UnionSortField =
...
-    }
-  | {
-      fields: ((string | number | boolean)[] | DataRef | SignalRef)[];
-    };
-export type ScaleData =
-  | (DataRef & { sort?: SortField })
-  | (MultiDataRef & { sort?: UnionSortField });
+
+export interface ScaleMultiDataRef {
+  fields: ((string | number | boolean)[] | ScaleDataRef | SignalRef)[];
```
domoritz: Do we have a type for `string | number | boolean`?

kanitw: Not that I know of.

kanitw: Let's merge first, if you find one we can fix them later.

PR #2402, opened 2020-02-26
Update dependencies.

jheer:
Update monorepo-wide project dependencies.


jheer:
Hmm, CI tests pass on Node 10 and 12, but not 13. For some reason, some canvas render tests fail there. Will need to investigate, but at this point I don't think this is specific to this PR.

jheer:
Hmm, can not replicate locally using node 13.8.0 (via brew). Travis CI is using node v13.9.0.

jheer:
Limited CI coverage to node 13.8.0 and tests pass. Merging for now, but we need to revisit this issue later.

Issue #2395, opened 2020-02-24
Allow lineHeight to affect single line text


kanitw:
Currently, lineHeight in Vega only affects multi-line text. 

However, as I've talked designers and observed how they do typography in UI design, I notice that they sometimes want to position text base on their line and line height instead of by font-size.

Thus, I would like to propose that we add two new `baseline` values: `"line-top"` and `"line-bottom"` (see below) so that one can easily place their text based on their line boxing model.

![image](https://user-images.githubusercontent.com/111269/75124534-adc04a00-5664-11ea-9d1e-618f8b0cfada.png)

(modified from https://smad.jmu.edu/shen/webtype/lineheight.html)

@jheer If you agree, I'm happy to go ahead and implement it.  

I believe that I only have to do a minor modification to the `offset()` method in https://github.com/vega/vega/blob/master/packages/vega-scenegraph/src/util/text.js#L131
and then adjust schema / types accordingly. 

cc: @donghaoren

kanitw:
I've added https://github.com/vega/vega/pull/2408, but feel free to replace it if you have better idea how to do it.

PR #2408, opened 2020-02-27
feat: allow text `baseline` to account for line-height with "line-top" and "line-bottom"

kanitw:
Fix https://github.com/vega/vega/issues/2395

From a quick skim, I'm not sure how you're testing for this part of code yet.

### DIFF
--- a/packages/vega-scenegraph/src/util/text.js
+++ b/packages/vega-scenegraph/src/util/text.js
@@ -129,14 +129,13 @@ export function offset(item) {
   // perform our own font baseline calculation
   // why? not all browsers support SVG 1.1 'alignment-baseline' :(
   var baseline = item.baseline,
-      h = fontSize(item),
-      lh = lineHeight(item);
+      h = fontSize(item);
 
   return Math.round(
-    baseline === 'line-top' ? 0.79.h + (lh-h) / 2.0 :
+    baseline === 'line-top' ? 0.79.h + (lineHeight(item)-h) / 2.0 :
     baseline === 'top'    ?  0.79*h :
     baseline === 'middle' ?  0.30*h :
     baseline === 'bottom' ? -0.21*h :
-    baseline === 'line-bottom' ? -0.21*h - (lh-h) / 2.0 : 0
+    baseline === 'line-bottom' ? -0.21*h - (lineHeight(item)-h) / 2.0 : 0
   );
 }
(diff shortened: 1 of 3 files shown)


packages/vega-scenegraph/src/util/text.js
```diff
@@ -129,10 +129,14 @@ export function offset(item) {
   // perform our own font baseline calculation
   // why? not all browsers support SVG 1.1 'alignment-baseline' :(
   var baseline = item.baseline,
-      h = fontSize(item);
+      h = fontSize(item),
+      lh = lineHeight(item);
```
jheer: This `lineHeight` call could be inlined to avoid invoking it even when it is not needed.

packages/vega-scenegraph/src/util/text.js
```diff
@@ -129,10 +129,14 @@ export function offset(item) {
...
+
   return Math.round(
+    baseline === 'line-top' ? 0.79.h + (lh-h) / 2.0 :
```
jheer: Which would be preferred: `'line-top'` or `'top-line'`? The first reads better to me, but the second groups better with the existing options. I don't have a strong opinion, but wanted to at least consider the question.

I also have a (very slight) concern that we are adding new baseline types that are not part of existing web standards. Our current values are a subset of valid [Canvas textBaseline options](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/textBaseline). Initially, my hope was to avoid this custom offset calculation altogether, but adding them made a huge improvement in text layout consistency between SVG and Canvas (and, by extension, hopefully for any other future renderers).

domoritz: `line-top` is similar to CSS properties like `padding-top` and `line-height`. Since it also reads easier, I'd prefer `line-top` over `top-line`.

packages/vega-scenegraph/src/util/text.js
```diff
@@ -130,9 +130,12 @@ export function offset(item) {
   // why? not all browsers support SVG 1.1 'alignment-baseline' :(
   var baseline = item.baseline,
       h = fontSize(item);
+
   return Math.round(
+    baseline === 'line-top' ? 0.79.h + (lineHeight(item)-h) / 2.0 :
```
jheer: I am fixing this myself, but note the `0.79.h` (which should be `0.79*h`), which suggests that this code has never been evaluated, much less tested.

jheer:
Merged via command line.

packages/vega-scenegraph/src/util/text.js
kanitw: sorry about that, I forgot to mark the PR as WIP (as I was doing it while traveling).

PR #2412, opened 2020-02-28
Add signal support for top-level spec properties.

### LINKED ISSUE #2394: Support signals for background color
It would be nice if Vega supported signals for the background color.

```json
{
  "width": 400,
  "height": 200,
  "background": {
    "signal": "'red'"
  }
}
```

### LINKED ISSUE #2417: Allow setting background, width, height, etc with signal
It would be nice to be able to write

```js
{
  padding: {signal: '10'},
}
```

instead of having to define a signal named `padding`. 

See https://github.com/vega/vega/pull/2412#issuecomment-592618807

jheer:
This PR does the following:

- The Vega parser now generates a built-in `background` signal which the view uses to set the background color. _While not technically a breaking change_ (specs will still parse and evaluate successfully), existing specs that use a signal named `"background"` may not render the same as before, in which case a different signal name should be used.
- The top-level properties `autosize`, `background`, `padding`, `width`, and `height` now accept signal references, such as `{"signal": "<expr>"}`, which map to a signal definition's `update` property. If a specifications top-level `signals` array contains an entry that matches one of these properties, the definitions will be merged, with precedence given to the properties defined in the `signals` array.

**vega**
- Update test specifications to avoid use of a signal named `"background"`.

**vega-parser**
- Add built-in `background` signal to drive view background.
- Add support for top-level properties to take signal-values.

**vega-schema**
- Add top-level signal-valued properties to schema.

**vega-typings**
- Add top-level signal-valued properties to view typings.
- Add top-level signal-valued properties to config typings.

**vega-view**
- Use `background` signal to control the view background color.
- Update `padding` method to handle numeric values.

**vega-view-transforms**
- Refine view layout resize code to avoid invocation of padding until needed.

Close #2394.
Close #2417.

### DIFF
--- a/packages/vega-parser/src/Scope.js
+++ b/packages/vega-parser/src/Scope.js
@@ -26,7 +26,6 @@ export default function Scope(config) {
   this.streams = [];
   this.updates = [];
   this.operators = [];
-  this.background = null;
   this.eventConfig = null;
 
   this._id = 0;
@@ -79,7 +78,6 @@ prototype.isSubscope = function() {
 prototype.toRuntime = function() {
   this.finish();
   return {
-    background:  this.background,
     operators:   this.operators,
     streams:     this.streams,
     updates:     this.updates,
--- a/packages/vega-parser/src/parsers/autosize.js
+++ b/packages/vega-parser/src/parsers/autosize.js
@@ -1,8 +1,5 @@
 import {isObject} from 'vega-util';
 
-export default function(spec, config) {
-  spec = spec || config.autosize;
-  return isObject(spec)
-    ? spec
-    : {type: spec || 'pad'};
+export default function(spec) {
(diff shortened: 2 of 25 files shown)


domoritz:
Thank you. Can we use the changes in this pull request to set the background to some value from the config?

```json
{
  background: {signal: 'mycolor'},
  ...
  config: {
    signals: [{
      name: 'mycolor',
      value: 'red'
    }]
  }
}
```

I think the answer is no and instead, we can only do

```json
{
  background: {signal: 'mycolor'},
  ...
  config: {
    signals: [{
      name: 'background',
      update: 'mycolor'
    }, {
      name: 'mycolor',
      value: 'red'
    }]
  }
}
```

The former would help with theming. Would it make sense to support that pattern or is there some conflict?

PR #2414, opened 2020-02-28
Update expressions.md

afdta:
I found that this example would have been useful when encountering the documentation for the first time.


jheer:
Thanks!

Issue #2418, opened 2020-02-29
Text with spaces at the beginning or end gets positioned differently in SVG and Canvas


domoritz:
```
{
  "$schema": "https://vega.github.io/schema/vega/v5.json",
  "padding": 5,
  "background": "lightblue",
  "marks": [
    {
      "type": "text",
      "encode": {
        "enter": {
          "x": {"value": 0},
          "y": {"value": 0},
          "align": {"value": "left"},
          "text": {"value": "       Hello Vega"}
        }
      }
    }
  ]
}
```

Canvas

![image](https://user-images.githubusercontent.com/589034/75616866-8f969600-5b0b-11ea-8a15-ead84c609d51.png)

SVG

<img src="https://user-images.githubusercontent.com/589034/75616868-96250d80-5b0b-11ea-8bfc-08fc8327e5b1.png">

jheer:
Ugh. That’s annoying. Personally I find the Canvas behavior more “correct”. But we probably do want to error on the side of consistency, and make sure we document this clearly on the text mark page.

domoritz:
I think the SVG behavior is consistent with HTML where spaces are often collapsed. I'm addressing the inconsistency in https://github.com/vega/vega/pull/2419.

PR #2419, opened 2020-03-01
Trim text to make rendering consistent

domoritz:
**vega-scenegraph**
- Trim text to make rendering between SVG and Canvas consistent.

Fixes https://github.com/vega/vega/issues/2418


domoritz:
cc @kanitw let's make sure not to rely on whitespace for spacing around text

jheer:
This PR is buggy (truncation should be performed on the trimmed string) and also lacks test coverage. Closing out, and will replace with a new PR instead.

PR #2438, opened 2020-03-06
Change default role from figure to graphics element

domoritz:
I think we want to use `graphics-document` and later annotate axes and legends with `graphics-object` and marks with `graphics-symbol` (based on https://blog.tenon.io/accessible-charts-with-aria). 

https://www.w3.org/TR/graphics-aria-1.0/#role_other shows that a figure is for wrapping graphics.


domoritz:
Actually, setting figure on the container makes sense.

PR #2442, opened 2020-03-09
docs: add comment to `addEncode`

kanitw:


packages/vega-parser/src/parsers/encode/encode-util.js
```diff
@@ -8,6 +8,7 @@ export function encoder(_) {
 
 export function addEncode(object, name, value, set) {
   if (value != null) {
+    // Always assign signal to update, even the signal is from the enter block
```
domoritz:
```suggestion
    // Always assign signal to update, even if the signal is from the enter block
```

jheer:
Thanks!

PR #2445, opened 2020-03-09
fix(vega-typings,vega-schema): fix schema for axis/legend properties

kanitw:
According to Slack conversation, all axis/legend properties that are transpiled via `addEncoders` do support signals. Thus, the typings we have are probably wrong. 

I haven't updated vega-schema/docs yet, wanna make sure that this should be officially supported.

In any case, it seems a bit painful that we have to update these in 3 places (vega-typings/schema/docs)

### DIFF
--- a/packages/vega-typings/types/spec/axis.d.ts
+++ b/packages/vega-typings/types/spec/axis.d.ts
@@ -182,7 +182,7 @@ export interface BaseAxis {
   /**
    * Horizontal text alignment of axis titles.
    */
-  titleAlign?: AlignValue;
+  titleAlign?: AlignValue | SignalRef;
 
   /**
    * Text anchor position for placing axis titles.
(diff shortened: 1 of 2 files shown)


domoritz:
> In any case, it seems a bit painful that we have to update these in 3 places (vega-typings/schema/docs)

We could work towards generating the schema from the typings to avoid this overhead. Similarly, the docs could be populated from the schema as we do in Vega-Lite.

packages/vega-typings/types/spec/axis.d.ts
```diff
@@ -182,7 +182,7 @@ export interface BaseAxis {
   /**
    * Horizontal text alignment of axis titles.
    */
-  titleAlign?: AlignValue;
+  titleAlign?: AlignValue | SignalRef;
```
domoritz: Isn't alignValue already a signal?

domoritz: AlignValue > AlignValueRef > ScaledValueRef > BaseValueRef > SignalRef

kanitw: ok right, then we don't need to do this?

domoritz: Yep

PR #2447, opened 2020-03-10
Restore Travis CI for node 13.

jheer:
**monorepo**

- Restore Travis CI to latest version of Node.js v13.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: node_js
 node_js:
   - "10"
   - "12"
-  - "13.8"
+  - "13"
 before_install:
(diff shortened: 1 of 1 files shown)


jheer:
Node v13 is causing failures due to differences in node-canvas rendering output. I was hoping this might be a transient issue with v13.9, but it has persisted with v13.10. Moreover, I'm unable to replicate locally on MacOS using Node v13.10.1.

domoritz:
The [latest supported node version is now node 14](https://nodejs.org/en/about/releases/).

jheer:
Superceded.

PR #2470, opened 2020-03-23
Update marks.md

afdta:
I could not find a case where the enter block is re-evaluated when the visualization is resized. 
For example, in a visualization I created that resizes based on the container width, the following encoding rule only takes effect when the marks are first instantiated. (The stroke changes color as expected when this style rule is placed in an update block.) 
{
  "enter": {
      "stroke":{"signal":"width < 400 ? 'red' : 'blue'"}
    }
}


jheer:
Thanks! The `enter` set is in fact re-run upon a resize, *but* only when that resize was forcibly created in response to the autosize calculation. As this is a special situation and not generally applied, I've merged in your edit to the documentation. Thanks for contributing!

PR #2500, opened 2020-04-02
Minify es5 build of vega

wylieconlon:
I found that the directory structure of the `build` directory was not the same as the `build-es5` directory, due to missing `.min.js` files.

Before:

```
build-es5/vega-core.js
build-es5/vega-module.js
build-es5/vega-node.js
build-es5/vega.js
```

After:

```
build-es5/vega-core.js
build-es5/vega-core.min.js
build-es5/vega-module.js
build-es5/vega-node.js
build-es5/vega.js
build-es5/vega.min.js
```


jheer:
Thanks!

Issue #1350, opened 2018-07-04
autosize:fit and resize:true results in rerendering when hovering over interactive marks


Eveliyse:
Hi there,
I've been tinkering with Vega and came across some weird behaviour where the view seems to be rerendering when I hover/mouseover interactive marks. This probably isn't an issue normally but when there are marks which are sliiightly 'miscalculated' the rerendering caues the graph to actually change and 'judder'

I've included an example spec:
```
{
  "autosize": {
    "type": "fit",
    "resize": true
  },
  "width": 175,
  "height": 200,
  "data": [
    {
      "name": "data",
      "values": [{ "category": "Bananananananananananananana", "value": 2 }]
    }
  ],
  "scales": [
    {
      "name": "categoryScale",
      "type": "band",
      "domain": { "data": "data", "field": "category" },
      "range": "width",
      "paddingInner": 0.2
    },
    {
      "name": "magnitudeScale",
      "type": "linear",
      "domain": { "data": "data", "field": "value" },
      "range": "height"
    }
  ],
  "axes": [
    { "orient": "bottom", "scale": "categoryScale" },
    { "orient": "left", "scale": "magnitudeScale" }
  ],
  "marks": [
    {
      "type": "rect",
      "name": "barMarks",
      "from": { "data": "data" },
      "encode": {
        "enter": {
          "x": { "scale": "categoryScale", "field": "category" },
          "width": { "scale": "categoryScale", "band": 1 },
          "y": { "scale": "magnitudeScale", "value": 0 },
          "y2": { "scale": "magnitudeScale", "field": "value" },
          "fill": { "value": "#1380A1" }
        }
      }
    }
  ]
}
```

When I run this in the online editor and mouseover/hover over the bar mark the view seems to rerender and you can see the word "Bananananananananananananana" slowly fit onto the graph. After doing this a few times it seems to "stabilise" and changes no longer occur when mousing over the graph (You might need to move on and off the mark several times for this to occur).

This behaviour stops if autosize type is `none`, autosize resize is `false` or when the bar marks are not interactive.

I guess the ideal solution is that the graph renders the "stabilised" hover render without the user having to mouseover interactive marks. But in the event that this is too difficult/time-consuming/requires a large overhaul, would it be possible to keep interactivity without the rerendering?

jheer:
[...]

PR #2503, opened 2020-04-03
Add signal support for axis orient

haldenl:
ezgif com-video-to-gif

Aimed at supporting axis orient as a signal

I've removed the test scenegraphs changes from the PR to make it easier to review. The changes made by rebuilding them is only `orient` being added to the encoder for axis (tests pass locally when this is done).

I wasn't sure which branch to rebase onto so I have it on `v5.11.0` for now.


packages/vega-parser/src/parsers/guides/axis-grid.js
```diff
@@ -37,26 +39,100 @@ export default function(spec, config, userEncode, dataRef, band) {
...
-  }
-  v2 = v + '2';
+  if (isSignal(orient)) {
+    val = {}
+    for (var key of Object.keys(tickPos)) {
+      val[key] = topOrBottomAxisExpr(orient.signal, tickPos[key],)
+    }
+
+    update['x'] = enter['x'] = {
+      scale: topOrBottomAxisExpr(orient.signal, tickPos.scale, vscale ? vscale : undefined),
```
domoritz, line 49: Can vscale be false or null or what's this check for?

packages/vega-parser/src/parsers/guides/axis-grid.js
```diff
@@ -37,26 +39,100 @@ export default function(spec, config, userEncode, dataRef, band) {
...
+      offset: topOrBottomAxisExpr(orient.signal, tickPos.offset, offset),
+      round: topOrBottomAxisExpr(orient.signal, tickPos.round, undefined),
+      range: topOrBottomAxisExpr(orient.signal, undefined, vscale ? 0 : undefined),
+      mult: topOrBottomAxisExpr(orient.signal, undefined, vscale ? sign : undefined),
+      value: topOrBottomAxisExpr(orient.signal, undefined, vscale ? undefined : 0)
+    };
 
-  update[u] = enter[u] = exit[u] = tickPos;
+    update['x2'] = enter['x2'] = {
+      scale: topOrBottomAxisExpr(orient.signal, tickPos.scale, vscale ? vscale : undefined),
```
domoritz, line 61: You can probably write a function to create these blocks.

packages/vega-typings/types/spec/axis.d.ts
```diff
@@ -47,7 +47,7 @@ export interface Axis extends BaseAxis {
    *
    * __Default value:__ `"bottom"` for x-axes and `"left"` for y-axes.
    */
-  orient: AxisOrient;
+  orient: AxisOrient | SignalRef;
```
domoritz, line 50: 🎉

packages/vega-parser/src/parsers/guides/axis-grid.js
haldenl: looks like it, this is following the original logic for non-signal orient (in this file)

packages/vega-parser/src/parsers/guides/axis-grid.js
haldenl: Good suggestion, done! Thanks

kanitw:
I think the title angle/ positioning is still incorrectly. 
(The GIF above didn't include title.)

It seems like we have to revisit bound calculation as well.  

```
{
  "background": "white",
  "padding": 5,
  "height": 200,
  "style": "cell",
  "data": [
    {
      "name": "source_0",
      "values": [
        {"Month": "05. January", "Temperature": 14.7, "Index": 1},
        {"Month": "04. February", "Temperature": 16.4, "Index": 2},
        {"Month": "08. March", "Temperature": 18.7, "Index": 3},
        {"Month": "01. April", "Temperature": 20.9, "Index": 4},
        {"Month": "09. May", "Temperature": 23.5, "Index": 5},
        {"Month": "07. June", "Temperature": 25.8, "Index": 6},
        {"Month": "06. July", "Temperature": 26, "Index": 7},
        {"Month": "02. August", "Temperature": 26.3, "Index": 8},
        {"Month": "12. September", "Temperature": 26.4, "Index": 9},
        {"Month": "11. October", "Temperature": 23.3, "Index": 10},
        {"Month": "10. November", "Temperature": 18.2, "Index": 11},
        {"Month": "03. December", "Temperature": 14.6, "Index": 12}
      ]
    }
  ],
  "signals": [
    {"name": "x_step", "value": 20},
    {
      "name": "width",
      "update": "bandspace(domain('x').length, 1, 0.5) * x_step"
    }
  ],
  "marks": [
    {
      "name": "marks",
      "type": "line",
      "style": ["line"],
      "from": {"data": "source_0"},
      "encode": {
        "update": {
          "stroke": {"value": "#4c78a8"},
          "x": {"scale": "x", "field": "Month"},
          "y": {"scale": "y", "field": "Temperature"},
          "defined": {
            "signal": "isValid(datum[\"Temperature\"]) && isFinite(+datum[\"Temperature\"])"
          }
        }
      }
    }
  ],
  "scales": [
    {
      "name": "x",
      "type": "point",
      "domain": {"data": "source_0", "field": "Month"},
      "range": {"step": {"signal": "x_step"}},
      "padding": 0.5
    },
    {
      "name": "y",
      "type": "linear",
      "domain": {"data": "source_0", "field": "Temperature"},
      "nice": true,
      "zero": true
    }
  ],
  "axes": [
    {
      "scale": "y",
      "orient": {"signal": "orient"},
      "gridScale": "x",
      "grid": true,
      "domain": false,
      "labels": false,
      "maxExtent": 0,
      "minExtent": 0,
      "ticks": false,
      "zindex": 0
    },
    {
      "scale": "x",
      "orient": "bottom",
      "grid": false,
      "title": "Month",
      "labelAlign": "right",
      "labelAngle": 270,
      "labelBaseline": "middle",
      "labelOverlap": true,
      "zindex": 0
    },
    {
      "scale": "y",
      "orient": {"signal": "orient"},
      "grid": false,
      "title": "Temperature",
      "labelOverlap": true,
      "zindex": 0
    }
  ],
  "config": {
    "signals": [
      {
        "name": "orient",
        "value": "left",
        "bind": {"input": "select", "options": ["left", "right"]}
      }
    ]
  }
}
```

image

image

kanitw:
Similar problem for y-axis:

```
{
  "background": "white",
  "padding": 5,
  "height": 200,
  "style": "cell",
  "data": [
    {
      "name": "source_0",
      "values": [
        {"Month": "05. January", "Temperature": 14.7, "Index": 1},
        {"Month": "04. February", "Temperature": 16.4, "Index": 2},
        {"Month": "08. March", "Temperature": 18.7, "Index": 3},
        {"Month": "01. April", "Temperature": 20.9, "Index": 4},
        {"Month": "09. May", "Temperature": 23.5, "Index": 5},
        {"Month": "07. June", "Temperature": 25.8, "Index": 6},
        {"Month": "06. July", "Temperature": 26, "Index": 7},
        {"Month": "02. August", "Temperature": 26.3, "Index": 8},
        {"Month": "12. September", "Temperature": 26.4, "Index": 9},
        {"Month": "11. October", "Temperature": 23.3, "Index": 10},
        {"Month": "10. November", "Temperature": 18.2, "Index": 11},
        {"Month": "03. December", "Temperature": 14.6, "Index": 12}
      ]
    }
  ],
  "signals": [
    {"name": "x_step", "value": 20},
    {
      "name": "width",
      "update": "bandspace(domain('x').length, 1, 0.5) * x_step"
    }
  ],
  "marks": [
    {
      "name": "marks",
      "type": "line",
      "style": ["line"],
      "from": {"data": "source_0"},
      "encode": {
        "update": {
          "stroke": {"value": "#4c78a8"},
          "x": {"scale": "x", "field": "Month"},
          "y": {"scale": "y", "field": "Temperature"},
          "defined": {
            "signal": "isValid(datum[\"Temperature\"]) && isFinite(+datum[\"Temperature\"])"
          }
        }
      }
    }
  ],
  "scales": [
    {
      "name": "x",
      "type": "point",
      "domain": {"data": "source_0", "field": "Month"},
      "range": {"step": {"signal": "x_step"}},
      "padding": 0.5
    },
    {
      "name": "y",
      "type": "linear",
      "domain": {"data": "source_0", "field": "Temperature"},
      "nice": true,
      "zero": true
    }
  ],
  "axes": [
    {
      "scale": "y",
      "orient": "left",
      "gridScale": "x",
      "grid": true,
      "domain": false,
      "labels": false,
      "maxExtent": 0,
      "minExtent": 0,
      "ticks": false,
      "zindex": 0
    },
    {
      "scale": "x",
      "orient": {"signal": "orient"},
      "grid": false,
      "title": "Month",
      "labelAngle": 270,
      "labelBaseline": "middle",
      "labelOverlap": true,
      "encode": {
        "labels": {
          "update": {
            "align": {"signal": "orient !== \"bottom\" ? \"left\" : \"right\""}
          }
        }
      },
      "zindex": 0
    },
    {
      "scale": "y",
      "orient": "left",
      "grid": false,
      "title": "Temperature",
      "labelOverlap": true,
      "zindex": 0
    }
  ],
  "config": {
    "signals": [
      {
        "name": "orient",
        "value": "top",
        "bind": {"input": "select", "options": ["top", "bottom"]}
      }
    ]
  }
}
```

image

image

haldenl:
Good catches, thanks @kanitw. I've fixed titles and gridlines.

ezgif com-video-to-gif

I also noticed a bug where repeated changes to the orient resulted in the axis mark group `width` property continuing to shrink until it reaches 0. Maybe related to #1350—this PR does not modify the `width` property of mark groups, so I believe this should be addressed in a different PR.

ezgif com-video-to-gif-2

packages/vega-parser/src/parsers/guides/axis-config.js
```diff
@@ -1,12 +1,35 @@
...
+
+
+    xy = {};
+    for (var prop of axisXYConfigKeys) {
+      xy[prop] = xAxisExpr(spec.orient.signal, config.axisX ? config.axisX[prop] : undefined, config.axisY ? config.axisY[prop] : undefined);
+    }
+
+    or = {};
+    for (prop of axisOrientConfigKeys) {
+      or[prop] = axisOrientExpr(spec.orient.signal, config.axisTop ? config.axisTop[prop] : undefined, config.axisBottom ? config.axisBottom[prop] : undefined, config.axisLeft ? config.axisLeft[prop] : undefined, config.axisRight ? config.axisRight[prop] : undefined);
```
kanitw, line 28: instead of checking if axisTop/Bottom/Left/Right exists everytime,

You could declare

```
axisTop = config.axisTop || {}
```

on the top and then you can refer to `config.axisTop` and `config.axisTop[prop]`.

(Same for other props)

packages/vega-parser/src/parsers/guides/axis-config.js
```diff
@@ -1,12 +1,35 @@
...
+    }
+  }
+
+  band = scope.scaleType(spec.scale) === 'band' && config.axisBand;
```
kanitw, line 32: It doesn't seems like band depends on any of the logic above. 
If so, this could stay up where band is declared? 

(no need to cause unnecessary diff.)

packages/vega-parser/src/parsers/guides/axis-config.js
```diff
@@ -1,12 +1,35 @@
...
   var config = scope.config,
       orient = spec.orient,
-      xy = (orient === Top || orient === Bottom) ? config.axisX : config.axisY,
-      or = config['axis' + orient[0].toUpperCase() + orient.slice(1)],
-      band = scope.scaleType(spec.scale) === 'band' && config.axisBand;
+      xy,
+      or,
+      band;
+
+  if (!isSignal(spec.orient)) {
```
kanitw, line 13: (Minor) Instead of using negation, wouldn't it be easier to read if we flip the if/else?

packages/vega-parser/src/parsers/guides/axis-domain.js
```diff
@@ -25,18 +27,27 @@ export default function(spec, config, userEncode, dataRef) {
     strokeOpacity:    _('domainOpacity')
   });
 
-  if (orient === Top || orient === Bottom) {
-    u = 'x';
-    v = 'y';
+  if (!isSignal(spec.orient)) {
```
kanitw, line 30: (Minor) Instead of using negation, wouldn't it be easier to read if we flip the if/else?

packages/vega-parser/src/parsers/guides/axis-domain.js
```diff
@@ -25,18 +27,27 @@ export default function(spec, config, userEncode, dataRef) {
...
+    }
+    u2 = u + '2';
+
+    enter[v] = zero;
+    update[u] = enter[u] = position(spec, 0);
+    update[u2] = enter[u2] = position(spec, 1);
   } else {
-    u = 'y';
-    v = 'x';
+    enter.y = xAxisConditionalEncoding(orient.signal, zero, position(spec, 0));
```
kanitw, line 44: xAxisConditionalEncoding -> xyAxisConditionalEncoding

packages/vega-parser/src/parsers/guides/axis-grid.js
```diff
@@ -38,49 +40,80 @@ export default function(spec, config, userEncode, dataRef, band) {
...
+      update[v] = enter[v] = {scale: vscale, range: 0, mult: sign, offset: offset};
+      update[v2] = enter[v2] = {scale: vscale, range: 1, mult: sign, offset: offset};
+    } else {
+      update[v] = enter[v] = {value: 0, offset: offset};
+      update[v2] = enter[v2] = {signal: s, mult: sign, offset: offset};
+    }
   } else {
-    update[v] = enter[v] = {value: 0, offset: offset};
-    update[v2] = enter[v2] = {signal: s, mult: sign, offset: offset};
+    update.x = enter.x = xAxisConditionalEncoding(orient.signal, tickPos, vscale ? { scale: vscale, range: 0, mult: sign, offset: offset} : { value: 0, offset: offset});
```
kanitw, line 65: This part is really hard to read, and repeating the same logic with ternary operators logic multiple times

- `vscale ? { scale: vscale, range: 0, mult: sign, offset: offset} : { value: 0, offset: offset}`

- `vscale ? {scale: vscale, range: 1, mult: sign, offset: offset} : {signal: xAxisExpr(orient.signal, 'height', 'width', false, false).signal, mult: sign, offset: offset}`

Extract common part and name them so we know what they are?

packages/vega-parser/src/parsers/guides/axis-util.js
```diff
@@ -0,0 +1,42 @@
...
+
+export function ifTopOrLeftAxisExpr(signalExpr, ifTopOrLeft, otherwise) {
+  var ifTopOrLeftStr = stringValue(ifTopOrLeft);
+  var otherwiseStr = stringValue(otherwise);
+  return {
+    signal: `(${signalExpr}) === "${Top}" || (${signalExpr}) === "${Left}" ? (${ifTopOrLeftStr}) : (${otherwiseStr})`
+  }
+}
+ 
+export function xAxisConditionalEncoding(signalExpr, ifXAxis, otherwise, isXAxis = true) {
```
kanitw, line 35: The fact that `isXAxis === true` makes it really hard to understand what you try to do when you call this method in `axis-grid.js`.  It's easier to read if you take 
`xy: 'x' | 'y'` as parameter.

packages/vega-parser/src/parsers/guides/axis-grid.js
```diff
@@ -38,49 +40,80 @@ export default function(spec, config, userEncode, dataRef, band) {
...
-
-    while (entry.mult != null) {
-      if (!isObject(entry.mult)) {
-        entry.mult *= sign;
-        return offset;
-      } else {
-        entry = entry.mult = extend({}, entry.mult);
+    if (!isObject(offset)) {
+      return {
+        signal: `(${sign.signal}) === 1 ? 0 : (${sign.signal}) * (${offset || 0})`
```
kanitw, line 101: if `offset` is zero, we don't really need signal here.

packages/vega-parser/src/parsers/guides/axis-util.js
```diff
@@ -0,0 +1,42 @@
+import { stringValue } from "vega-util";
+import { Top, Bottom, Left } from "./constants";
+
+export function xAxisExpr(signalExpr, ifXAxis, otherwise, ifXAxisIsString = true, otherwiseIsString = true) {
```
kanitw, line 4: - This works for both x and y, so the name should be `xy...`


- It returns a signal ref, so naming it `Expr` is misleading.

packages/vega-parser/src/parsers/guides/axis-util.js
```diff
@@ -0,0 +1,42 @@
...
+  return {
+    signal: `${xAxisBooleanExpr(signalExpr)} ? (${ifXAxisStr}) : (${otherwiseStr})`
+  }
+}
+  
+export function xAxisBooleanExpr(signalExpr, isXAxis = true) {
+  return `${isXAxis ? '' : '!'}((${signalExpr}) === "${Top}" || (${signalExpr}) === "${Bottom}")`
+}
+
+export function axisOrientExpr(signalExpr, top, bottom, left, right) {
```
kanitw, line 16: This returns a signal ref, so naming it `Expr` is misleading.

packages/vega-parser/src/parsers/guides/axis-util.js
```diff
@@ -0,0 +1,42 @@
...
+  var bottomStr = stringValue(bottom);
+  var leftStr = stringValue(left);
+  var rightStr = stringValue(right);
+
+  return {
+    signal: `(${signalExpr}) === "${Top}" ? (${topStr}) : (${signalExpr}) === "${Bottom}" ? (${bottomStr}) : (${signalExpr}) === "${Left}" ? (${leftStr}) : (${rightStr})`
+  }
+}
+
+export function ifTopOrLeftAxisExpr(signalExpr, ifTopOrLeft, otherwise) {
```
kanitw, line 27: This returns a signal ref, so naming it `Expr` is misleading.

packages/vega-parser/src/parsers/guides/axis-util.js
kanitw: Also, I found the flags `ifXAxisIsString = true, otherwiseIsString = true` hard to follow. 

I'd rather get rid of it and just apply stringValue when I call these, to simplify.

kanitw: It also seems like `signalExpr` is not an arbitrary `signalExpr`, but rather `orientSignal`.  Better give a specific name.

packages/vega-parser/src/parsers/guides/axis-util.js
```diff
@@ -0,0 +1,42 @@
...
+
+export function xAxisExpr(signalExpr, ifXAxis, otherwise, ifXAxisIsString = true, otherwiseIsString = true) {
+  var ifXAxisStr = ifXAxisIsString ? stringValue(ifXAxis) : ifXAxis;
+  var otherwiseStr = otherwiseIsString ? stringValue(otherwise) : otherwise;
+  return {
+    signal: `${xAxisBooleanExpr(signalExpr)} ? (${ifXAxisStr}) : (${otherwiseStr})`
+  }
+}
+  
+export function xAxisBooleanExpr(signalExpr, isXAxis = true) {
```
kanitw, line 12: It also seems like `signalExpr` is not an arbitrary `signalExpr`, but rather `orientSignal`.  Better give a specific name.

packages/vega-parser/src/parsers/guides/axis-util.js
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export function axisOrientExpr(signalExpr, top, bottom, left, right) {
+  var topStr = stringValue(top);
+  var bottomStr = stringValue(bottom);
+  var leftStr = stringValue(left);
+  var rightStr = stringValue(right);
+
+  return {
+    signal: `(${signalExpr}) === "${Top}" ? (${topStr}) : (${signalExpr}) === "${Bottom}" ? (${bottomStr}) : (${signalExpr}) === "${Left}" ? (${leftStr}) : (${rightStr})`
```
kanitw, line 23: It also seems like `signalExpr` is not an arbitrary `signalExpr`, but rather `orientSignal`.  Better give a specific name.

packages/vega-parser/src/parsers/guides/axis-util.js
kanitw: It also seems like `signalExpr` is not an arbitrary `signalExpr`, but rather `orientSignal`.  Better give a specific name.

packages/vega-parser/src/parsers/guides/axis-util.js
kanitw: It also seems like `signalExpr` is not an arbitrary `signalExpr`, but rather `orientSignal`.  Better give a specific name.

packages/vega-parser/src/parsers/guides/axis-util.js
```diff
@@ -0,0 +1,42 @@
...
+  }
+}
+ 
+export function xAxisConditionalEncoding(signalExpr, ifXAxis, otherwise, isXAxis = true) {
+  return [
+    {
+      test: xAxisBooleanExpr(signalExpr, isXAxis),
+      ...ifXAxis
+    }
+  ].concat(otherwise ? [otherwise] : []);
```
kanitw, line 41: `[].concat(1) = [1]` so we can also just do

`.concat(otherwise || [])`

packages/vega-parser/src/parsers/guides/axis-util.js
kanitw: The flag `ifXAxisIsString = true, otherwiseIsString = true` makes this method a bit harder to read when called. 

An alternative would be that `ifXAxis` can be either a string or a signalRef.  For the latter, no need to apply stringValue.

packages/vega-parser/src/parsers/guides/axis-grid.js
```diff
@@ -38,49 +40,80 @@ export default function(spec, config, userEncode, dataRef, band) {
...
   return guideMark(RuleMark, AxisGridRole, null, Value, dataRef, encode, userEncode);
 }
 
 function offsetValue(offset, sign)  {
-  if (sign === 1) {
-    // do nothing!
-  } else if (!isObject(offset)) {
-    offset = sign * (offset || 0);
+  var entry;
+  if (!isSignal(sign)) {
```
kanitw, line 78: Instead of breaking by `isSignal(sign)` at the top-level branch, I think you can keep a lot of existing structure and check if `isSignal(sign)` in the existing if/else and while loop structure. 

This will avoid redundant code and make it easier to assess if we're dealing with `sign` signal correctly.

packages/vega-parser/src/parsers/guides/axis-labels.js
```diff
@@ -40,30 +41,50 @@ export default function(spec, config, userEncode, dataRef, size, band) {
...
-    enter: enter = {
+    enter = {
       opacity: zero,
       x: isXAxis ? tickPos : tickSize,
       y: isXAxis ? tickSize : tickPos
-    },
+    }
+  } else {
+    align = labelAlign ||
+      { signal: `(${orient.signal}) === "${Top}" || (${orient.signal}) === "${Bottom}" ? (${flushOn ? flushExpr(scale, flush, '"left"', '"right"', '"center"').signal : "'center'"}) : (${orient.signal}) === "${Right}" ? "left" : "right"`};
```
kanitw, line 70: This is really hard to read. Please add line breaks:

```
{ 
  signal: `(${orient.signal}) === "${Top}" || (${orient.signal}) === "${Bottom}" ?` +  

    `(${flushOn ? flushExpr(scale, flush, '"left"', '"right"', '"center"').signal : "'center'"}) :` +  
    `(${orient.signal}) === "${Right}" ? "left" : "right"`
};
```

Same for `baseline`, `offsetExpr`

kanitw: Also, one way to avoid executing orient's signal twice (if it's an expression) is to use 
`indexof(['top', 'bottom'], ${orient.signal}) >= 0`

packages/vega-parser/src/parsers/guides/axis-title.js
```diff
@@ -33,14 +35,21 @@ export default function(spec, config, userEncode, dataRef) {
...
+      enter.angle = {value: sign * 90};
+      enter.baseline = {value: 'bottom'};
+    }
   } else {
-    update.y = titlePos;
-    enter.angle = {value: sign * 90};
-    enter.baseline = {value: 'bottom'};
+    update.x = xAxisConditionalEncoding(orient.signal, titlePos, null);
+    update.y = xAxisConditionalEncoding(orient.signal, titlePos, null, false);
+    enter.angle = update.angle = xAxisConditionalEncoding(orient.signal, zero, { signal: `warn((${sign.signal}) * 90)` });
```
kanitw, line 51: why `warn`?

packages/vega-parser/src/parsers/guides/axis-config.js
haldenl: done 👍

packages/vega-parser/src/parsers/guides/axis-config.js
haldenl: 👍

packages/vega-parser/src/parsers/guides/axis-config.js
haldenl: I was thinking the not signal case is the more general / simpler so I placed it first, but I also agree that it is more difficult to read. Addressed.

packages/vega-parser/src/parsers/guides/axis-domain.js
haldenl: 👍

packages/vega-parser/src/parsers/guides/axis-domain.js
haldenl: 👍

packages/vega-parser/src/parsers/guides/axis-grid.js
haldenl: yep, done.

packages/vega-parser/src/parsers/guides/axis-util.js
haldenl: refactored. good point

packages/vega-parser/src/parsers/guides/axis-grid.js
haldenl: We need the else case though. Is there a workaround that I'm missing?

packages/vega-parser/src/parsers/guides/axis-config.js
```diff
@@ -1,14 +1,57 @@
...
+        axisTop = config.axisTop || {},
+        axisBottom = config.axisBottom || {},
+        axisLeft = config.axisLeft || {},
+        axisRight = config.axisRight || {},
+        axisXYConfigKeys = new Set(
+          Object.keys(config.axisX || {})
+            .concat(Object.keys(config.axisY || {}))
+        ),
+        axisOrientConfigKeys = new Set(
+          Object.keys(config.axisTop || {})
```
kanitw, line 28: I now wonder if we can use

```
[
  ...Object.keys(config.axisTop || {}),
  ...Object.keys(config.axisBottom || {}),
  ...Object.keys(config.axisLeft || {}),
  ...Object.keys(config.axisRight || {}),
]
```

(easier to read)

packages/vega-parser/src/parsers/guides/axis-util.js
haldenl: refactored to discriminate signalRefs 👍

packages/vega-parser/src/parsers/guides/axis-util.js
haldenl: 👍

packages/vega-parser/src/parsers/guides/axis-util.js
haldenl: done

packages/vega-parser/src/parsers/guides/axis-util.js
haldenl: 👍

packages/vega-parser/src/parsers/guides/axis-util.js
haldenl: 👍

packages/vega-parser/src/parsers/guides/axis-util.js
haldenl: nice, thank you.

packages/vega-parser/src/parsers/guides/axis-domain.js
```diff
@@ -25,18 +27,27 @@ export default function(spec, config, userEncode, dataRef) {
     strokeOpacity:    _('domainOpacity')
   });
 
-  if (orient === Top || orient === Bottom) {
-    u = 'x';
-    v = 'y';
+  if (isSignal(spec.orient)) {
+    enter.y = xyAxisConditionalEncoding('x', orient.signal, zero, position(spec, 0));
```
kanitw, line 31: Why is this using 'x' for `enter.y`?

packages/vega-parser/src/parsers/guides/axis-grid.js
haldenl: refactored

packages/vega-parser/src/parsers/guides/axis-labels.js
haldenl: Added the line breaks.

I find that expression harder to read. Maybe it is okay given this is not exposed?

packages/vega-parser/src/parsers/guides/axis-title.js
haldenl: my bad 😬

packages/vega-parser/src/parsers/guides/axis-grid.js
```diff
@@ -38,48 +44,78 @@ export default function(spec, config, userEncode, dataRef, band) {
...
-        return offset;
-      } else {
-        entry = entry.mult = extend({}, entry.mult);
+  if (!isSignal(sign)) {
+    if (sign === 1) {
+      // do nothing!
+    } else if (!isObject(offset)) {
+      offset = isSignal(sign) ? {
+        signal: `(${sign.signal}) === 1 ? 0 : (${sign.signal}) * (${offset || 0})`
+      } : offset = sign * (offset || 0);
```
kanitw, line 99: why do we need `offset =` after `:`?

packages/vega-parser/src/parsers/guides/axis-labels.js
```diff
@@ -40,30 +41,65 @@ export default function(spec, config, userEncode, dataRef, size, band) {
...
+          + `(${orient.signal}) === "${Right}" ? "left" : "right"`
+      };
 
-  offset = offset && flushOn && flushOffset
-    ? flushExpr(scale, flush, '-(' + flushOffset + ')', flushOffset, 0)
-    : null;
+    baseline = labelBaseline ||
+      {
+        signal: `(${orient.signal}) === "${Top}" ? "bottom"`
+          + ' : '
```
kanitw, line 57: Not sure why we don't append `:` in the previous line?

packages/vega-parser/src/parsers/guides/axis-ticks.js
```diff
@@ -38,14 +40,23 @@ export default function(spec, config, userEncode, dataRef, size, band) {
     round:  _('tickRound')
   };
 
-  if (orient === Top || orient === Bottom) {
-    update.y = enter.y = zero;
-    update.y2 = enter.y2 = tickSize;
-    update.x = enter.x = exit.x = tickPos;
+  if (isSignal(orient)) {
+    update.y = enter.y = xyAxisConditionalEncoding('x', orient.signal, zero, tickPos);
```
kanitw, line 44: It feels weird to have to parameter as 'x' for `enter.y`, same for the rest.

packages/vega-parser/src/parsers/guides/axis-config.js
kanitw: Same for the axisXYConfigKeys

packages/vega-parser/src/parsers/guides/axis-domain.js
kanitw: Why should `enter.y` be different from `update.y` while they are the same in the non-signal condition?

packages/vega-parser/src/parsers/guides/axis-grid.js
```diff
@@ -38,48 +44,78 @@ export default function(spec, config, userEncode, dataRef, band) {
...
-  update[u] = enter[u] = exit[u] = tickPos;
+  gridLineEnd = vscale
+    ? {scale: vscale, range: 1, mult: sign, offset: offset}
+    : {signal: s, mult: sign, offset: offset};
 
-  if (vscale) {
-    update[v] = enter[v] = {scale: vscale, range: 0, mult: sign, offset: offset};
-    update[v2] = enter[v2] = {scale: vscale, range: 1, mult: sign, offset: offset};
+  if (isSignal(orient)) {
+    for (xy of ['x', 'y']) {
```
kanitw, line 61: rename xy to `u`?  (easier to read)

packages/vega-parser/src/parsers/guides/axis-grid.js
kanitw: It's not a big deal, I guess we can leave it as is.

packages/vega-parser/src/parsers/guides/axis-grid.js
kanitw: Why is this `!isSignal(sign)` still here?

packages/vega-parser/src/parsers/guides/axis-labels.js
```diff
@@ -40,30 +41,65 @@ export default function(spec, config, userEncode, dataRef, size, band) {
...
-    baseline = labelBaseline || (flushOn
-      ? flushExpr(scale, flush, '"top"', '"bottom"', '"middle"')
-      : 'middle');
-    offset = !labelBaseline;
-  }
+  if (isSignal(orient)) {
+    align = labelAlign ||
+      {
+        signal: `(${orient.signal}) === "${Top}" || (${orient.signal}) === "${Bottom}"`
+          + ' ? '
```
kanitw, line 48: I think we only need 3 lines -- `?` can go to the previous line.

packages/vega-parser/src/parsers/guides/axis-labels.js
```diff
@@ -40,30 +41,65 @@ export default function(spec, config, userEncode, dataRef, size, band) {
...
+          + `(${orient.signal}) === "${Bottom}" ? "top"`
+          + ' : '
+          + `${flushOn ? flushExpr(scale, flush, '"top"', '"bottom"', '"middle"') : "'middle'"}`
+      };
 
-  encode = {
-    enter: enter = {
+    offsetExpr = flushExpr(scale, flush, '-(' + flushOffset + ')', flushOffset, 0).signal;
+    offset = `${xyAxisBooleanExpr('x', orient.signal)}`
+      + ' ? ' +
```
kanitw, line 65: let's do 3 lines

packages/vega-parser/src/parsers/guides/axis-ticks.js
kanitw: (You can flip the yes and no case?)

packages/vega-parser/src/parsers/guides/axis-title.js
```diff
@@ -33,14 +35,31 @@ export default function(spec, config, userEncode, dataRef) {
...
+        {signal: `(${orient.signal}) === "${Top}" ? "bottom" : "top"`},
+        { value: 'bottom' }
+      );
   } else {
-    update.y = titlePos;
-    enter.angle = {value: sign * 90};
-    enter.baseline = {value: 'bottom'};
+    if (horizontal) {
+      update.x = titlePos;
+      enter.angle = {value: 0};
```
kanitw, line 56: why do we use `zero` up there in the signal and `{value: 0}` here? 

I guess they are the same, but better be consistent?

domoritz: We should use `zero` when possible.

packages/vega-parser/src/parsers/guides/axis-domain.js
haldenl: Rewritten to make this logic more clear (now matches the pattern in the non-signal case).

haldenl: (Matched in other files)

packages/vega-parser/src/parsers/guides/axis-ticks.js
haldenl: It makes more sense to me to read this as: the `y` position of the tick is at `zero` if we are dealing with the `x` axis. In that sense it seems unnecessary to force these values to be identical. 

I rewrote this too to make the logic more immediately similar to the non-signal case. I think it's easier to read these lines if the parameter is identical for all the conditionals (s.t. they can be grouped as `x` axis logic or `y` axis logic).

packages/vega-parser/src/parsers/guides/axis-grid.js
haldenl: much easier!

packages/vega-parser/src/parsers/guides/axis-labels.js
haldenl: sure

packages/vega-parser/src/parsers/guides/axis-labels.js
haldenl: 👍

packages/vega-parser/src/parsers/guides/axis-title.js
haldenl: fixed

packages/vega-parser/src/parsers/guides/axis-labels.js
haldenl: fixed

packages/vega-parser/src/parsers/guides/axis-grid.js
haldenl: my bad, fixed!

packages/vega-parser/src/parsers/guides/axis-labels.js
```diff
@@ -77,8 +107,18 @@ export default function(spec, config, userEncode, dataRef, size, band) {
     }
   };
 
+  if (!isSignal(orient)) {
```
jheer, line 110: It is confusing to have multiple if blocks here, particularly one that checks `isSignal(orient)` followed by one that checks `!isSignal(orient)`. Can we either consolidate into one if block, or break the branches apart into different helper functions?

packages/vega-parser/src/parsers/guides/axis-labels.js
```diff
@@ -91,6 +131,8 @@ export default function(spec, config, userEncode, dataRef, size, band) {
     limit:       _('labelLimit'),
     lineHeight:  _('labelLineHeight')
   });
+    
+
```
jheer, line 135: Extra empty lines?

packages/vega-parser/src/parsers/guides/axis-title.js
```diff
@@ -58,13 +77,45 @@ export default function(spec, config, userEncode, dataRef) {
...
+    } else {
+      if (!has('x', userEncode)) {
+        encode.enter.auto = xyAxisConditionalEncoding('y', orient.signal, { value: true }, null);
+      }
+    }
+
+    if (_('titleY') != null) {
+      delete encode.update['y'][0].signal;
+      encode.update['y'][0] = {
+        ...encode.update['y'][0],
```
jheer, line 96: [Spread for object literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Spread_in_object_literals) is a more recent addition to some browsers. I've been avoiding it in Vega so that we don't run into backwards compatibility issues. (Even though I love them...)

packages/vega-schema/src/axis.js
```diff
@@ -151,6 +153,7 @@ const axis = object({
 
 export default {
   refs: {
+    axisOrient,
```
jheer, line 156: Why is this added as a ref? We use it directly in the axis definition, and I don't see a reference to this schema created and used elsewhere.

packages/vega-parser/src/parsers/guides/axis-util.js
```diff
@@ -0,0 +1,47 @@
...
+export function xyAxisSignalRef(xy, axisOrientExpr, yes, no) {
+  var yesStr = isSignal(yes) ? yes.signal : stringValue(cleanValue(yes));
+  var noStr = isSignal(no) ? no.signal : stringValue(cleanValue(no));
+  return {
+    signal: `${xyAxisBooleanExpr(xy, axisOrientExpr)} ? (${yesStr}) : (${noStr})`
+  };
+}
+  
+export function xyAxisBooleanExpr(xy, axisOrientExpr) {
+  return `${xy === 'x' ? '' : '!'}((${axisOrientExpr}) === "${Top}" || (${axisOrientExpr}) === "${Bottom}")`;
```
jheer, line 14: Referencing expressions multiple times like this just inlines the full expression, leading to redundant evaluation. Obviously the expression language doesn't support defining variables, but another option could be to define an encoding property once (either on an item or parent item) and then reference it in the formula. I don't think it is critical to do this now, but something to consider.

packages/vega-parser/src/parsers/guides/axis-title.js
domoritz: Maybe we should add babel even for our normal builds. I can take a stab at a pull request at some point if you are open to accepting it.

packages/vega-parser/src/parsers/guides/axis-util.js
kanitw: At least for this function, we could use `indexOf(["${Top}", "${Bottom}"], ${axisOrientExpr}) >= 0` to call the axisOrientExpr only once here. 

Of course, we still have redundant calculation across calls of `xyAxisBooleanExpr`, but at least we will reduce the redundancy by half.

packages/vega-parser/src/parsers/axis.js
```diff
@@ -95,6 +95,7 @@ function buildAxisEncode(_, spec) {
   var encode = {enter: {}, update: {}};
 
   addEncoders(encode, {
+    orient:       _('orient'),
```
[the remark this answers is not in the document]
haldenl: removed! and changed references of `datum.orient` to `item.orient`

packages/vega-parser/src/parsers/guides/axis-config.js
```diff
@@ -1,14 +1,57 @@
...
+  } else {
+    var axisX = config.axisX || {},
+        axisY = config.axisY || {},
+        axisTop = config.axisTop || {},
+        axisBottom = config.axisBottom || {},
+        axisLeft = config.axisLeft || {},
+        axisRight = config.axisRight || {},
+        axisXYConfigKeys = new Set(
+          Object.keys(config.axisX || {})
+            .concat(Object.keys(config.axisY || {}))
```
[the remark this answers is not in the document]
haldenl: My bad, fixed 👍

packages/vega-parser/src/parsers/guides/axis-labels.js
haldenl: removed

packages/vega-parser/src/parsers/guides/axis-title.js
haldenl: I've switched these out for `extend()` for now.

packages/vega-schema/src/axis.js
haldenl: Forgot about this—removed!

packages/vega-parser/src/parsers/guides/axis-util.js
haldenl: Switched to `indexof` for now.

haldenl: I've also reduce the redundancy of cleaning values (for `signalRefs` and `undefined`).

packages/vega-parser/src/parsers/guides/axis-labels.js
haldenl: I've consolidated the evaluation of `labelAlign`, `labelBaseline`, and `offset` into a single block by combining the logic for signal and non-signal cases using a resolving function. Happy to make additional changes here.

haldenl:
Thank you @jheer! I made a pass and addressed your comments / suggestions so far.

packages/vega-parser/src/parsers/guides/axis-grid.js
haldenl: removed now

packages/vega-parser/src/parsers/guides/axis-labels.js
haldenl: Changed to indexof, per discussion below.

packages/vega-parser/src/parsers/guides/axis-config.js
haldenl: concat safer for backwards compat? (per discussion below on spread operator)

packages/vega-parser/src/parsers/guides/axis-config.js
```diff
@@ -1,14 +1,57 @@
...
+    }
+
+    or = {};
+    for (prop of axisOrientConfigKeys) {
+      or[prop] = allAxisOrientSignalRef(
+        spec.orient.signal,
+        axisTop[prop],
+        axisBottom[prop],
+        axisLeft[prop],
+        axisRight[prop]
```
haldenl, line 44: @jheer If a prop exists in one axis config but not the others, this will conditionally set that property to `null`. Perhaps for limited use cases this okay (the user would need to match props in the configs if they intend to swap axis orient), but ideally this would instead conditionally set it to the default value. Is there a readily available approach to implementing this?

jheer: Perhaps use the top-level config.axis as backup? Though that might still write null values over lower-level mark or style config entries. You are learning the hard way why I did not make axis orient a signal in the past...

haldenl: It is quite a hairy refactor...

Ah, I hadn't thought about falling back to the top-level config. I also looked into the default style config (in `vega-parser/src/config.js`) and implemented a fallback routine that accesses those values. I believe this solves the issue—it seems like `null` for properties that are not in the axis config or the default styles are handled gracefully by the renderer (correct me if I'm wrong).

jheer:
Replaced by #2610

PR #2504, opened 2020-04-03
Allow only one of prevent/allow prop for defaults

### LINKED ISSUE #2502: Events defaults should require either prevent or allow, not both
[Currently](https://github.com/vega/vega/blob/993c370568fbfa5251d957882f262a60efd0c44f/packages/vega-typings/types/spec/config.d.ts#L73) the types for Events defaults actually requires both `allow` and `prevent`, while [the docs](https://vega.github.io/vega/docs/config/#events) specify that you must only specify _one_ of the two. When I try to do that in TypeScript, I get  

```
Property 'allow' is missing in type '{ prevent: false; }' but required in type 'Record<"prevent" | "allow", boolean | EventType[]>'.ts(2741)
```
or the opposite:
```
Property 'prevent' is missing in type '{ allow: true; }' but required in type 'Record<"prevent" | "allow", boolean | EventType[]>'.ts(2741)
```

willium:
**vega-typings**

- Allow either `prevent` or `allow` in `DefaultsConfig`.

Enforcing 
> The defaults object should have a single property: either "prevent" (to indicate which events should have default behavior suppressed) or "allow" (to indicate only those events whose default behavior should be allowed).
from the [docs](https://vega.github.io/vega/docs/config/#events)

Fixes #2502

### DIFF
--- a/packages/vega-typings/types/spec/config.d.ts
+++ b/packages/vega-typings/types/spec/config.d.ts
@@ -73,9 +73,9 @@ export interface Config
 /**
  *  The defaults object should have a single property: either "prevent" (to indicate which events should have default behavior suppressed) or "allow" (to indicate only those events whose default behavior should be allowed).
  */
-type PreventDefaultConfig = Record<'prevent', boolean | EventType[]>;
-type AllowDefaultConfig = Record<'allow', boolean | EventType[]>;
-export type DefaultsConfig = PreventDefaultConfig | AllowDefaultConfig;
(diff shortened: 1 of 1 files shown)


domoritz:
Can you fix the formatting?

domoritz:
Can you run tests locally?

Screen Shot 2020-04-03 at 17 33 31

willium:
@domoritz seems to pass now

PR #2506, opened 2020-04-06
chore: delete old logos

domoritz:


jheer:
Why delete these without replacement? I'm not sure all the places they might still be referenced...

domoritz:
I tried to look for any references to these images on google and couldn't find anything. Also, do you know where the image came from? @kanitw and I don't remember.

PR #2507, opened 2020-04-06
Add Prettier and make ESLint stricter

domoritz:
This pull request enables prettier and a few stricter eslint rules.

The goal is to make the code more consistent and easier for others to contribute to.

Side note: I through adding prettier and stricter eslint would be easy but I had to go through a lot of cases where I had to manually replace var with const/let. I added a few ignores in cases where I wasn't sure so I'm not breaking anything.


domoritz:
This is probably too much at once. I'll split it.

domoritz:
Actually, this could still make sense as an alternative to #2514. You can decide what you prefer.

jheer:
Thanks for all the tedious work here. The stricter ESLint rules are great, and I like more modular PRs, so I merged the other PRs instead. I totally get the value of adding prettier as well -- particularly when collaboration needs to scale up -- but I also don't always like the code it produces. Maybe I'm old-fashioned but it (by default anyway) erases some of what I find to be nice little human touches of programming (e.g., friendly alignment via extra space characters). But of course there are trade-offs! Happy to further consider using prettier, but after I have time to more deeply explore additional configuration options.

PR #2514, opened 2020-04-07
Stricter ESLint

domoritz:
ESLint now enforces correct semicolons and quotes. I also moved eslint up to the root so that it can be enforced for everything and in a single call.


jheer:
Thanks!

PR #2533, opened 2020-04-18
fix(vega-typings): cornerRadius* support SignalRef

kanitw:
Part of https://github.com/vega/vega-lite/issues/6352

### DIFF
--- a/packages/vega-typings/types/spec/config.d.ts
+++ b/packages/vega-typings/types/spec/config.d.ts
@@ -400,39 +400,39 @@ export interface MarkConfig {
   // ---------- Corner Radius: Bar, Tick, Rect ----------
 
   /**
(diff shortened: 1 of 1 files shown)


packages/vega-typings/types/spec/config.d.ts
```diff
@@ -400,39 +400,39 @@ export interface MarkConfig {
   // ---------- Corner Radius: Bar, Tick, Rect ----------
 
   /**
-   * The radius in pixels of rounded rectangle or arc corners.
+   * The radius in pixels of rounded rectangles or arcs' corners.
```
domoritz, line 403: I think it's

```suggestion
   * The radius in pixels of rounded rectangles' or arcs' corners.
```

domoritz: Actually, probably not.

PR #2534, opened 2020-04-18
feat: support axis `domain/grid/tickCap`

kanitw:
Part of https://github.com/vega/vega-lite/issues/6353

PR #2542, opened 2020-04-21
A branch for Vega pre-release

kanitw:
@jheer 

To accelerate the features that needs review to get into Vega's official release, we're creating this `next` branch so that we can make a pre-release with a tag `next` instead of the default `latest` tag. 

We'll merge our urgent PRs into this branch so that you can review them collectively whenever you have time to make the next official Vega release.


domoritz:
Changelog for ARIA pull request:

**vega-parser**

- Add support for `ariaLabel`, `ariaRole`, `ariaRoleDescription`, and `ariaHidden` to encodings, axes, legends, and titles.
- Automatically set the role of guide groups to `graphics-object`.
- Create default descriptions for axes and legends.
- Set default aria role and aria role description for marks.

**vega-schema**

- Add support for `ariaLabel`, `ariaRole`, `ariaHidden`, and `ariaRoleDescription` to encodings, axes, legends, and titles.

**vega-scenegraph**

- Add support for `ariaLabel`, `ariaRole`, `ariaHidden`, and `ariaRoleDescription` to SVG renderer.

**vega-typings**

- Add `ariaLabel`, `ariaRole`, `ariaHidden`, and `ariaRoleDescription` to encodings, axes, legends, and titles.

Now we can support https://blog.tenon.io/accessible-charts-with-aria and https://www.w3.org/TR/graphics-aria-1.0/.

domoritz:
Changelog for #2534 

**vega-parser**
- Add axis `domain/grid/tickCap`

**vega-typings**
- Add axis `domain/grid/tickCap`

jheer:
I'm afraid this is going to make reviewing much more difficult. Collectively reviewing a pile of PRs that target different needs that have been squashed onto one thread does not sound like a good solution to me, especially when some may need further edits and others may not.

I'm not sure if it is strictly necessary to add this extra level of management. I try to be prompt with reviews and updates. Recently this has taken longer than usual because of the scope and not-yet-finalized design decisions of the aria additions, other work obligations, and a coronavirus home schooling schedule. I don't mind having an "experimental" branch with non-official releases, but I don't want to categorize major, possibly unvetted changes under "next" while also complicating the review process.

kanitw:
> I don't mind having an "experimental" branch with non-official releases, but I don't want to categorize major, possibly unvetted changes

We can rename `next` to `exp` or `experimental` if you prefer that. 

(Between the two, I prefer `exp` since it's shorter.)

> Collectively reviewing a pile of PRs that target different needs that have been squashed onto one thread does not sound like a good solution to me, especially when some may need further edits and others may not.

The other solution we considered is to submit urgent PRs twice -- one to merge into `exp` and another official one for `master`.  The latter can still go through normal review process.

To avoid flooding the pull requests on the main Vega repo, we can also put the experimental branch and PRs in a separate `vega-exp(erimental)` repo.

(ps. I actually original suggested this exp repo approach, but we chose to propose this unified branch as it might be simpler to handle conflicts.  But given this will complicate the review process, I agree that making it clear that it's experimental is better.)

In any case, since there are other packages downstream and we [can't make them depend on a subfolder of monorepo's git url](https://github.com/npm/npm/issues/2974), we still need to make experimental releases to run test with downstream packages (e.g., experimental branch in Vega-Lite).  Please add us to Vega subpackages so we can still make experimental prereleases.

PR #2543, opened 2020-04-21
chore: add `release-next` script

kanitw:


domoritz:
- [ ] Make sure this release script does not automatically update the schema.

jheer:
Closing for now. The added command hopefully won't be needed (at least not often) and is simple enough to be used directly in the future if needed. If we end up using it more than expected we can consider re-adding in the future (though using 'experimental' instead of 'next').

PR #2546, opened 2020-04-21
Add ARIA accessibility features.

jheer:
Restoring this PR for review and revision, after it was merged onto the `next` branch.


jheer:
Note: an alternative approach to ARIA attribute generation is explored in #2548.

jheer:
FYI I noticed a likely bug in vega-typings for the title guide. It looks like the subtitle ARIA properties are not properly named, and the `Title` type is overriding the ARIA properties with identical property names.

jheer:
Superceded by #2548

PR #2548, opened 2020-04-21
Add ARIA attribute generation.

jheer:
_Changelog:_

**monorepo**

- Add ARIA attribute generation documentation.
- Update dev dependencies.

**vega**

- Add `overview-detail-bins` test spec to test suite.
- Update crossfilter test scenes to include `description` property output.

**vega-encode**

- Move internal tick, label utilities to vega-scale.

**vega-parser**

- Add ARIA attribute generation for marks and guides via `aria` and `description` properties.
- Add backing scale names to axis, legend datum objects.
- Revert guide, mark parsing order and zindex config.

**vega-scale**

- Add tick, label guide utilities from vega-encode.
- Add `domainCaption` utility.

**vega-scenegraph**

- Add ARIA attribute generation to SVG renderers:
  - Parent `<g>` tags for mark items include automatic `role` and `aria-roleDescription` attributes.
  - Parent `<g>` tags for axes and legends include automatic `aria-label` captions.
  - Attribute generation for mark and guide definitions can be customized using the `aria` and `description` properties. If `aria` is `false`, the content is hidden from the accessibility tree by setting `aria-hidden` SVG attribute.. The `description` property determines the `aria-label` SVG attribute.
  - Individual mark items do not include ARIA attributes by default, but these can be added uing the mark encoding channels `aria`, `description`, `ariaRole`, and `ariaRoleDescription`.

**vega-schema**

- Add ARIA accessibility properties to mark, guide, and encode schemas.

**vega-typings**

- Add ARIA accessibility properties to mark, guide, and encode typings.

_Rationale_:

This PR explores an alternate ARIA attribute generation approach from #2546. It is intended to address the following issues:

- The prior approach introduced cross-cutting concerns over multiple Vega packages. This PR instead focuses on extending vega-scenegraph to take primary responsibility for attribute generation. This also simplifies some logic by moving it out of signal calculations within the Vega dataflow and into standard JavaScript helper routines.
- The prior approach generated ARIA attributes even when the renderer (e.g., Canvas) can not use them. This PR instead only invokes automatic ARIA attribute generation when using a supporting renderer.
- The prior approach introduces bloat to both the scenegraph and output SVG by automatically adding default ARIA annotations to *all* mark items. This PR instead adds annotations to each enclosing group for a set of sibling mark items. Individual attributes can still be specified via `aria*` encoding channels, but they are not used by default.
- Domain arrays for discrete scales can be arbitrarily long. This PR summarizes discrete domains with more than 10 elements. (We can adjust this threshold if necessary.)
- Even when drawing on multiple scales, legends only have one backing domain. This PR updates legend captions to reflect this.
- The prior approach attempted to provide correct formatting for domain values, but the solution was only partial: only explicitly specified formats were used, internally determined formatting (e.g., calculated precision) was still overlooked. This PR reuses Vega's system for formatting axis and legend labels.

This PR has one notable drawback: the automatically generated `aria-label` captions for mark groups and guides do not yet dynamically update upon changes to the underlying scale domains. That is, after the SVG is rendered, the ARIA attributes for mark groups and guides will not automatically update. However, ARIA attributes defined in encoding channels for individual mark items will dynamically update.


jheer:
Here is an example of annotated SVG output:

image

packages/vega-scenegraph/src/util/aria.js
```diff
@@ -0,0 +1,125 @@
...
+  return peek(item.items).items[0].text;
+}
+
+function channelCaption(props) {
+  props = props.map(p => p + (p === 'fill' || p === 'stroke' ? ' color' : ''));
+  return props.length < 2 ? props[0]
+    : props.slice(0, -1).join(', ') + ' and ' + peek(props);
+}
+
+function isDiscrete(type) {
```
domoritz, line 109: Why did you decide to write this method instead of using `isDiscrete` from vega scales?

domoritz:
By the way, having aria properties on guides would be great so that we can easily hide e.g. grid axes (the axes we generate only for grid lines) without having to generate a custom encode block. Is supporting guide properties more difficult now?

domoritz:
Ah, while updating Vega-Lite, I noticed a bug. In this pull request, I cannot hide axes anymore. 

Screen Shot 2020-04-21 at 16 35 33
Screen Shot 2020-04-21 at 16 36 43

jheer:
Thanks @dom. I will add some more fine grained control over the auto-generated attributes this Thursday, including support to hide guides as needed.

packages/vega-scenegraph/src/util/aria.js
jheer: Fixed now that I've moved the domain captioning utility to vega-scale.

PR #2550, opened 2020-04-22
Use Math.floor to avoid overflow with large domain and small tickMinStep

rwoollen:
If you render the attached spec with a tickMinStep of `0.01` the axis labels disappear.  Setting a tickMinStep of `0.1` works fine.  The issue appears to be the `~~` operator forces a 32-bit conversion.
[vega-overflow.zip](https://github.com/vega/vega/files/4513100/vega-overflow.zip)


jheer:
Thank you for catching the problem and submitting a PR! The fix looks good. However, I'm currently refactoring the tick utilities in a feature branch and so have instead applied this fix there. It will be included in the v5.11 release. Thanks again!

PR #2549, opened 2020-04-21
v5.11.0

jheer:
### Notable Additions

- Preliminary support for [ARIA accessibility attributes](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA) in SVG output. Mark and guide definitions now include automatically-generated roles and labels, which can be customized using `description` and `aria` properties. Individual mark items do not include ARIA attributes by default (which helps prevent bloat of both the scenegraph and output SVG), but these can be added using `description` and `aria` encoding channels.
- Improve generated HTML form elements for signal bindings. Deployments that use custom CSS for styling bound elements may wish to make minor adjustments.
- `dayofyear` time unit support for the `timeunit` transform.
- `dayofyear`, `week`, `utcdayofyear`, and `utcweek` expression functions.
- Axis `domainCap`, `gridCap`, and `tickCap` properties.

### Changelog

Changes from [v5.10.1](https://github.com/vega/vega/releases/tag/v5.10.1):

**monorepo**

- Add ARIA attribute generation documentation.
- Update eslint setup, consolidate configuration.
- Update dev dependencies.

**vega**

- Add `calendar` test specification.
- Add `overview-detail-bins` test spec to test suite.
- Update crossfilter test scenes to include `description` property output.

**vega-encode**

- Move internal tick, label utilities to vega-scale.
- Fix valid tick check calculation. (#2531)

**vega-functions**

- Add `dayofyear`, `week`, `utcdayofyear`, `utcweek` expression functions.

**vega-geo**

- Fix density utility size input checking bug.

**vega-parser**

- Add ARIA attribute generation for marks and guides via `aria` and `description` properties.
- Add backing scale names to axis, legend datum objects.
- Add `zindex` support for guide config.
- Add axis `domainCap`, `gridCap`, and `tickCap` properties.

**vega-scale**

- Add tick, label guide utilities from vega-encode.
- Add `domainCaption` utility.
- Fix overflow with large domain and small `tickMinStep` (#2550, thanks @rwoollen!)

**vega-scenegraph**

- Add ARIA attribute generation to SVG renderers:
  - Parent `<g>` tags for mark items include automatic `role` and `aria-roleDescription` attributes.
  - Parent `<g>` tags for axes and legends include automatic `aria-label` captions.
[...]

Issue #2494, opened 2020-03-30
Add automatic aria labels for marks


domoritz:
Follow up for #2380.

It would be great if we could automatically annotate marks with appropriate aria labels. It would be nice if we could describe the data before scales are applied ("circle representing Europe with horsepower 12 and 4 cylinders").

jheer:
Related support added in #2548. Mark item-level aria attributes can be defined using encoding channels.

domoritz:
Not yet. The intention of this issue is to provide default mark-level aria-labels that describe individual marks without the need for explicit descriptions via encoding channels. I can understand if we don't want to do this at all since Vega charts often have marks that would just clutter the screenreader experience.

Issue #2492, opened 2020-03-30
Automatically add aria-role group so screenreader users can skip over large umbers of marks


domoritz:
Follow up for https://github.com/vega/vega/pull/2380. 

For example in https://vega.github.io/vega/examples/contour-plot/, it would be great to be able to skip over the points.

jheer:
Related support added in #2548. I believe this concern in covered by the more recent ARIA attribute generation scheme?

domoritz:
Yes, it does. I like that the new ARIA scheme summarizes mark groups into one. However, it currently also prevents screen readers from accessing children. 

I think what we may want is to summarize marks when individual marks don't define aria properties but otherwise still set mark-level properties. Let's continue the discussion in https://github.com/vega/vega/pull/2548.

PR #2553, opened 2020-04-23
fix: use a group role to allow screen readers to navigate the hierarchy

domoritz:


packages/vega-scenegraph/src/util/aria.js
```diff
@@ -6,6 +6,7 @@ const ARIA_LABEL = 'aria-label';
 const ARIA_ROLE = 'role';
 const ARIA_ROLEDESCRIPTION = 'aria-roledescription';
 const DEFAULT_ROLE = 'graphics-symbol';
```
domoritz, line 8: Should we rename this to `MARK_ROLE` or `INDIVIDUAL_MARK_ROLE`?

domoritz:
Actually, there is a bigger issue here. We need to set aria-role on individual marks, not on the group that surrounds marks. 

Notice in this SVG (I manually changed `graphics-symbol` to `graphics-object`)

Screen Shot 2020-04-23 at 09 13 41

the description is set on the group mark but that means the individual children don't get the right role description.

packages/vega-scenegraph/src/util/aria.js
jheer: Or maybe rename to `SYMBOL_ROLE`? Also which of these should be used for guides?

jheer:
To clarify: for the screen reader to properly navigate the individual marks, do the mark svg elements needs to have a `role` property set? If so, you could set that using the `ariaRole` encoding channel, yes? (We could possibly automate the choice of `role` if `description` is defined.) Remember that by default we don't add any annotations to marks because that can cause massive bloat to both the scenegraph and output SVG. If we automate the `role` choice by default (assuming that is necessary) that will at least reduce bloat to the scenegraph objects.

domoritz:
It depends a bit on the screenreader (unfortunately it's like in times of IE6) but generally only `<text>` elements are read even without an explicit `aria-label`, `aria-role`, or `aria-roledescription` property. So if we want to not have VoiceOver read out an element, we should not set any property.

domoritz:
In my initial pull request, I decided to add aria roles and aria roledescriptions so that even when the developer does not provide any aria annotations, the screenreader shows something and the person accessing the chart with a screenreader knows that there is something. Vega-Lite will automatically generate a description for marks since it has the high-level semantic information. I filed https://github.com/vega/vega/issues/2494 as a follow up to automatically add annotations for individual marks in Vega.

What I like about your approach is that groups of marks are summarized into a single annotation (solving https://github.com/vega/vega/issues/2492). I think what we need to do is add the aria properties to individual marks when they are explicitly provided by the user.

I'm closing this pull request since it's not really addressing the issues I described in [my review](https://github.com/vega/vega/pull/2549#pullrequestreview-399267475).

packages/vega-scenegraph/src/util/aria.js
domoritz: Guides should use some role that does not have children so that VoiceOver does not read `empty X` after reading the description.

PR #2554, opened 2020-04-23
Make groups navigatable with screenreaders

domoritz:
Without this change, I wasn't able to go into group marks defined in a Vega spec.


jheer:
Fixed in a separate commit.

Issue #2547, opened 2020-04-21
Allow supplying your own node-canvas to render to


pixelspark:
Dear all!

Love your work on Vega. It's been a delight to work with so far. We are embedding Vega in our online tool for making data-driven infographics (see an example here: https://diasli.de/3367/SLafyLSGqP/HHynjfHRN1hcsBp8EoCHy9). 

For this we would like to generate PDF's server side of pages that embed Vega charts. Vega can render to PDF without issue, but node-canvas does not allow compositing multiple canvases into one PDF without loss of quality. What we currently do is render the Vega chart, then get its canvas, then render the chart to the main page using `pageContext.drawImage(vegaCanvas, ...)`. This unfortunately draws the Vega chart as a bitmap to the PDF.

It would be nice if it were possible to tell Vega to draw to a specific canvas (using transforms you can then place the chart on your page). This should allow high-quality (vector) embedding of Vega charts into PDF documents.

Suggested API is to allow calling `toCanvas` with a single `context` parameter (which is a node-canvas 2d context). The chart is then simply drawn at 0x0 to the desired width and height configured for the view:

````
const view = new vega.View(...., {renderer: "none"}).width(desiredWidth).height(desiredHeight);
const canvas = (some canvas we are re-using)
const context = canvas.getContext("2d");
context.save();
context.translate(100,100); // I want the chart at 100,100 in my document
await view.toCanvas(context);
context.restore();
````

Would be happy to implement this myself, but will need some pointers on where to begin!

jheer:
Vega does have existing server-side PDF rendering support through node-canvas. For example, see:
- https://github.com/vega/vega/blob/master/packages/vega-cli/bin/vg2pdf#L11
- Here we pass the following options: `{type: 'pdf', context: {textDrawingMode: 'glyph'}}`

Is it possible for you to invoke that, get the generated canvas/PDF, and use that subsequently? Or do you definitely need to render into an existing canvas?

If the answer is "yes", the code path to follow is:
- View.toCanvas: https://github.com/vega/vega/blob/master/packages/vega-view/src/render-to-canvas.js#L10
- View.renderHeadless: https://github.com/vega/vega/blob/master/packages/vega-view/src/render-headless.js#L17
- View.initializeRenderer: https://github.com/vega/vega/blob/master/packages/vega-view/src/initialize-renderer.js#L6
[...]

PR #2555, opened 2020-04-24
Allow user to specify own canvas context when rendering in headless mode

pixelspark:
Implements https://github.com/vega/vega/issues/2547

This allows a user to specify a pre-existing context to draw into (instead of Vega creating one). This can be used to render multiple charts to a single PDF document without quality loss. The example below renders two charts to the same PDF:

````js
const Vega = require("./packages/vega");
const fs = require('fs');
const { createCanvas, loadImage } = require('canvas');

// Try headless render
const spec = require("./spec.json");

async function main() {
	const view = new Vega.View(Vega.parse(spec), {
		logger: Vega.logger(Vega.Warn, 'error'),
		renderer: "none"
	}).width(640).height(480);

	const canvas = createCanvas(640, 960, "pdf");
	const ctx = canvas.getContext('2d');

	await view.toCanvas(4.0, {
		context: ctx
	});

	ctx.translate(0, 460);
	await view.toCanvas(4.0, {
		context: ctx
	});

	const out = fs.createWriteStream(__dirname + '/test.pdf');
	const stream = canvas.createPDFStream();
	stream.pipe(out);
	out.on('finish', () =>  console.log('The PDF file was created.'));
}

main();
````

### CODE REVIEW

File: packages/vega-scenegraph/src/CanvasRenderer.js [hunk 1]
```diff
@@ -31,9 +36,16 @@ prototype.initialize = function(el, width, height, origin, scaleFactor, options)
...
 prototype.resize = function(width, height, origin, scaleFactor) {
+  if (this._resized) {
+    throw new Error('cannot resize canvas twice');
+  }
+
+  this._resized = true;
   base.resize.call(this, width, height, origin, scaleFactor);
-  resize(this._canvas, this._width, this._height,
-    this._origin, this._scale, this._options && this._options.context);
+  if (this._canvas) {
```
jheer, line 45: It looks like resize will never occur with a passed-in context, as `this._canvas` will be undefined?

packages/vega-scenegraph/src/CanvasRenderer.js
```diff
@@ -31,9 +36,16 @@ prototype.initialize = function(el, width, height, origin, scaleFactor, options)
 };
 
 prototype.resize = function(width, height, origin, scaleFactor) {
+  if (this._resized) {
+    throw new Error('cannot resize canvas twice');
+  }
+
+  this._resized = true;
```
jheer, line 43: I'd prefer to have the logic prevent this from happening, alongside any relevant test cases, rather than add a new member variable and check an assertion at runtime.

packages/vega-scenegraph/src/util/canvas/resize.js
```diff
@@ -4,21 +4,23 @@ function devicePixelRatio() {
...
         ratio = inDOM ? pixelRatio : scaleFactor;
 
-  canvas.width = width * ratio;
-  canvas.height = height * ratio;
+  if (canvas) {
+    canvas.width = width * ratio;
+    canvas.height = height * ratio;
+  }
 
   for (const key in opt) {
```
jheer, line 19: With a passed-in context, I don't think we want to apply all the context-specific operations in the `resize` utility. For example, here we may overwrite desired properties (set externally) of the node-canvas context. That said, the bigger concern is that the transform matrix gets completely overwritten later in the method via `setTransform`!

packages/vega-scenegraph/src/CanvasRenderer.js
pixelspark: Oops, there should indeed not be an 'if' here!

packages/vega-scenegraph/src/CanvasRenderer.js
pixelspark: OK

packages/vega-scenegraph/src/util/canvas/resize.js
pixelspark: Applying setTransform to `context` is fine as long as the user calls `save` (and later `restore`) on it, right?

pixelspark:
OK, so I have now moved the resizing logic (translate + scale to specified scale factor) of the user-supplied context back to CanvasRenderer. The `setTransform` is replaced with calls to `translate` and `scale` as to not overwrite the user's transform.

Not sure how we should prevent calling resize() multiple times as it is not prevented for the regular case either. With regular use it is not called more than once and produces the correct results. I removed the runtime check.

pixelspark:
Just noticed that `toCanvas` already accepted `{context: ...}` as an option to set various context options. I now changed the option name to provide a custom context to `intoContext`. TypeScript typings have been changed accordingly, which ensures that you either specify `intoContext` *or* `context`/`type`, not both at the same time. (I considered re-using `context` and checking whether a context or an options object was supplied using `instanceof`, but that is tricky as this should both work in the browser as well as on the server).

@jheer please let me know if you have any further suggestions!

jheer:
Incorporated into PR #2575

PR #2558, opened 2020-04-24
Create more semantic bindings to improve accessibility

### LINKED ISSUE #2490: Vega bindings are not semantic HTML
For example in https://vega.github.io/vega/examples/pie-chart/

Screen Shot 2020-03-30 at 14 36 15

* The labels are not associated with the form elements (e.g. the checkbox or the slider)
* The label for sliders are showing the values. I believe [labels](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label) should be the description, not the value. 
* We could also improve accessibility here and mark the values as hidden since they are already accessible via the slider.

domoritz:
Fixes #2490

The example spec I tried

```json
{
  "signals": [
    {
      "name": "palette",
      "value": "Viridis",
      "bind": {"input": "select", "options": ["Viridis"]}
    },
    {
      "name": "offset",
      "value": "zero",
      "bind": {"input": "radio", "options": ["zero", "center"]}
    },
    {"name": "reverse", "value": false, "bind": {"input": "checkbox"}},
    {"name": "value", "value": 10, "bind": {"input": "range"}},
    {"name": "name", "value": "Hello", "bind": {"input": "text"}}
  ]
}
```

Before

After

### DIFF
--- a/packages/vega/test/index.html
+++ b/packages/vega/test/index.html
@@ -22,7 +22,9 @@
       .vega-bind { line-height: 18px; margin-bottom: 2px; }
       .vega-bind-name { display: inline-block; width: 100px; }
       .vega-bind input[type="range"] { width: 400px; }
-      .vega-bind-radio label { margin: 0 0.5em 0 2px; }
+      .vega-bind input[type="radio"] { margin-right: 0.5em; }
+      .vega-bind input[type="range"] { margin-right: 0.5em; }
+      .vega-bind-radio label { margin-right: 1em; }
     </style>
   </head>
   <body>
--- a/packages/vega/test/index.html
+++ b/packages/vega/test/index.html
@@ -21,10 +21,9 @@
       canvas, svg { border: 1px dashed #ccc; }
       .vega-bind { line-height: 18px; margin-bottom: 2px; }
       .vega-bind-name { display: inline-block; width: 100px; }
-      .vega-bind input[type="range"] { width: 400px; }
-      .vega-bind input[type="radio"] { margin-right: 0.5em; }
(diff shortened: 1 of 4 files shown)


packages/vega/test/index.html
```diff
@@ -22,7 +22,7 @@
       .vega-bind { line-height: 18px; margin-bottom: 2px; }
       .vega-bind-name { display: inline-block; width: 100px; }
       .vega-bind input[type="range"] { width: 400px; }
-      .vega-bind label { margin: 0 0.5em 0 2px; }
+      .vega-bind-radio label { margin: 0 0.5em 0 2px; }
```
jheer, line 25: Why is the class here limited to radio? This change, in addition to dropping the use of the `label` tag for `range` elements, removes the formatting for the range value label. Was this intentional?

I think these changes may also break downstream deployments that expect specific element types and/or css selectors for their custom formatting. Does this necessitate a breaking (major version) change? (Note that I'm fine with dropping the option element id, I'm referring to the other changes.)

Can we keep the nice changes you've made here without breaking existing stylesheets?

domoritz: Ahh, good catch about the range label. I'll look into that.

The reason why radio is different is that radio is the only input type in vega bindings that has multiple inputs (dropdown, checkbox, range, and text are a single input per label).

I don't think we can not guarantee exactly the same selectors since I'm switching the type of HTML tag in some places. I'm all for reducing changes but this may be a case where we cannot make any changes without changing someone's layout (https://xkcd.com/1172/).

domoritz: `.vega-bind label` for example now selects elements that are nested differently.

PR #2559, opened 2020-04-26
Test supported node versions

domoritz:
https://nodejs.org/en/about/releases/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ node_js:
   - "12"
   - "14"
 before_install:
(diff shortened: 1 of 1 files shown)


domoritz:
Looks like we are having the same issue as in https://github.com/vega/vega/pull/2447.

jheer:
Perhaps related to https://github.com/Automattic/node-canvas/issues/1557?

domoritz:
CLosing since I think we should switch to github actions: https://github.com/vega/vega/pull/2625

PR #2560, opened 2020-04-26
Various code fixes

domoritz:
Details in the commit messages. 

Discovered in https://lgtm.com/projects/g/vega/vega


jheer:
Thanks! Fixed separately on v5.11.0 branch. The density2d change wasn't quite right because there was actually a bug there (it needs to check both dimensions, not just one). Nice catch!

PR #2563, opened 2020-04-27
Fix aria labels

sprmn:
The generated aria labels result in invalid svg syntax. When I output the svg string to a file and try to open it, it shows a syntax error. The reason is the usage of double quotes in the label, which generates for example the following output:

```svg
<g aria-label="X-axis titled "Some title" for a linear scale with values from 0 to 13">
```

The double quotes break the attribute, as shown in the highlighting above. 

This PR replaces the double quotes by single quotes, which should fix the svg syntax.


jheer:
Thank you for the PR! The double-quoted strings work well in the browser, but clearly weren't escaped correctly in the string rendered output. Even if they were correctly escaped, I prefer the single quotes for more readable output.

Issue #2573, opened 2020-04-30
IDs in SVG outputs are not always deterministic


kanitw:
We're generating `id`s in the SVG output (e.g., for clip and gradient).

However, these output ids are not deterministic when we run Vega multiple times.

For example, running  [this spec](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JACyUAVhDYA7EABoQAEzjQATjRyZ289AHE2bMoLgByCAAIImNlADWlnJqhxLbUuss6EcSopUkTBRUUAZ1QTRA4M5bBwhKKAhiEABfJUx1JFkIADM2dQQ0AG1QXJpBTDh1KOUghgRKCABPBCY2QQBebrN9AHk+-TM0gF0lZHV7NFBMZpw4KKR1OBQlYVkF0JAoDoLapfsyZbh5dO3dmq2ADwBGNBulZrvUB5ArgCZ7x8-UAAYlI5IZQ0E6YKLrFY1JQ2Nh4EqgNi5XIQOBgv5KHaCPboADuMBEaQUCKRKLRr0x2MCkyOcBOo1SZxOO2BsjI0ze7PKcEEyn2VQCs3mUSqCBwBSQkTOzU5IJ5UWcNFcArmmxAAEcGNkdMEdKQ0gygA) multiple times produces different `id` for gradient in the SVG.

This is quite problematic for integration testing where we try to detect regression in SVG output from Vega-Lite and dependent projects. 
We sometimes see SVG diffs even though nothing changes in a plot besides the `id`s.

I guess that this might be related to how we generate `_vgsid_` internally.  
If so, is there a way to reset the global counter?

kanitw:
If it helps, these are the Vega-Lite _gallery_ examples that are affected by this problem:

```
area_gradient
bar_count_minimap
bar_diverging_stack_population_pyramid
brush_table
circle_bubble_health_income
circle_natural_disasters
concat_population_pyramid
geo_choropleth
geo_repeat
interactive_area_brush
interactive_bin_extent
interactive_brush
interactive_concat_layer
interactive_overview_detail
interactive_splom
joinaggregate_mean_difference_by_year
rect_binned_heatmap
rect_heatmap
rect_heatmap_weather
selection_translate_scatterplot_drag
stacked_bar_count_corner_radius_mark
stacked_bar_h
[...]

PR #2574, opened 2020-04-30
feat(vega-scenegraph): add resetSVGDefIds()

kanitw:
Fix https://github.com/vega/vega/issues/2573

With this new `resetGradientId()`, we can address https://github.com/vega/vega/issues/2573 by calling both `resetGradientId()` and `resetSVGClipId()`.

### DIFF
--- a/packages/vega-scenegraph/index.js
+++ b/packages/vega-scenegraph/index.js
@@ -59,6 +59,9 @@ export {
   pickVisit as scenePickVisit
 } from './src/util/visit';
 
+// deprecated, remove in next major version
+export {resetSVGClipId} from './src/util/svg/clip';
+
 export function resetSVGDefIds() {
   resetSVGClipId();
   resetSVGGradientId();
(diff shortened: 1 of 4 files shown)


jheer:
Thanks! Can we rename to `resetSVGGradientId` for parity? While gradients are shared across renderers, IIRC the id is only there for SVG use.

While not critical, we could consider a `resetSVGDefIds` helper that invokes both reset methods. If so, this might be added in the top-level index.js.

domoritz:
Do we ever need to reset only one? If not, maybe we can just have only `resetSVGDefIds`?

kanitw:
> Do we ever need to reset only one? If not, maybe we can just have only resetSVGDefIds?

Removing `resetSVGClipId()` will be a breaking change?

domoritz:
It's not documented (right?) and I don't think anyone is using it (a quick search across Github didn't reveal anything).

kanitw:
Ok, I added a few commits to replace the tests to use `resetSVGDefIds()` instead and remove resetSVGGradientId, resetSVGClipID.  

@jheer Feel free to drop some commits as you see fit.

PR #2575, opened 2020-05-01
Add externalContext CanvasRenderer option.

jheer:
**vega-scengraph**

- Add `externalContext` CanvasRenderer option.


pixelspark:
👍

packages/vega-typings/types/runtime/index.d.ts
```diff
@@ -20,6 +20,17 @@ export interface Loader {
   file: (filename: string) => Promise<string>;
 }
 
+export interface ToCanvasOptionsWithoutContext {
+  type?: string;
+  context?: any;
+}
+
+export interface ToCanvasOptionsWithContext {
+  intoContext?: any;
```
pixelspark, line 29: This should be 'externalContext' now right?

jheer: Thanks, good catch! I've pushed an update.

PR #2593, opened 2020-05-05
feat: signal / numbervalue support for axis translate.

haldenl:
translate-signal

NumberValue is also supported by moving `translate` from `datum` to `encode`

Example spec:

```
{
  "description": "A basic bar chart example, with value labels shown upon mouse hover.",
  "width": 400,
  "height": 200,
  "padding": 5,

  "data": [
    {
      "name": "table",
      "values": [
        {"category": "A", "amount": 28},
        {"category": "B", "amount": 55},
        {"category": "C", "amount": 43},
        {"category": "D", "amount": 91},
        {"category": "E", "amount": 81},
        {"category": "F", "amount": 53},
        {"category": "G", "amount": 19},
        {"category": "H", "amount": 87}
      ]
    }
  ],

  "signals": [
    {
      "name": "xAxisTranslate",
      "value": 0.5,
      "bind": {
        "input": "range",
        "min": -5,
        "max": 5
      }
    },
    {
      "name": "yAxisTranslate",
      "value": 0.5,
      "bind": {
        "input": "range",
        "min": -5,
        "max": 5
      }
    }
  ],

  "scales": [
    {
      "name": "xscale",
      "type": "band",
      "domain": {"data": "table", "field": "category"},
      "range": "width",
      "padding": 0.05,
      "round": true
    },
    {
      "name": "yscale",
      "domain": {"data": "table", "field": "amount"},
      "nice": true,
      "range": "height"
    }
  ],

  "axes": [
    { "orient": "bottom", "scale": "xscale", "translate": { "signal": "xAxisTranslate" }},
    { "orient": "left", "scale": "yscale", "translate": { "signal": "yAxisTranslate" }}
  ],

  "marks": [
    {
      "type": "rect",
      "from": {"data":"table"},
      "encode": {
        "enter": {
          "x": {"scale": "xscale", "field": "category"},
          "width": {"scale": "xscale", "band": 1},
          "y": {"scale": "yscale", "field": "amount"},
          "y2": {"scale": "yscale", "value": 0}
        },
        "update": {
        },
        "hover": {
          "fill": {"value": "red"}
        }
      }
    }
  ]
}
```

addresses #2522

### DIFF
--- a/packages/vega-parser/src/parsers/axis.js
+++ b/packages/vega-parser/src/parsers/axis.js
@@ -33,7 +33,6 @@ export default function(spec, scope) {
     grid:   !!_('grid'),
     domain: !!_('domain'),
     title:  spec.title != null,
-    translate: _('translate')
   };
   dataRef = ref(scope.add(Collect({}, [datum])));
 
@@ -111,6 +110,7 @@ function buildAxisEncode(_, spec) {
     minExtent:    _('minExtent'),
     maxExtent:    _('maxExtent'),
     range:        {signal: `abs(span(range("${spec.scale}")))`},
+    translate:    _('translate'),
 
     // accessibility support
     format:       spec.format,
--- a/packages/vega-schema/src/axis.js
+++ b/packages/vega-schema/src/axis.js
@@ -56,7 +56,7 @@ const axis = object({
   offset: numberValue,
   position: numberValue,
   bandPosition: numberValue,
-  translate: numberType,
+  translate: numberValue,
   values: arrayOrSignal,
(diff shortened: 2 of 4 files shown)


haldenl:
I should have used NumberValue for the schema, which includes signal (but number or signal does not allow Value). Thanks for the catch.

Use case: it would be good to be able to author a single spec that is able to react to different displays (e.g. pixel density). In some cases aligning with the pixel grid using a dynamic offset is required to resolve aliasing issues (this can be achieved by slightly adjusting the default offset).

PR #2595, opened 2020-05-06
Use babel for Vega builds

domoritz:
Now we can use modern JavaScript and don't have to worry about backward compatibility. Babel will take care of it. 

Note that we are only transpiling vega itself, not every single package. I don't think we need to transpile every package.


jheer:
Based on the discussion in #2596 it looks like transpiling the vega/vega output only may not be sufficient. For the time being I'd also rather avoid adding more build process overhead for a handful of syntactic conveniences. Happy to revisit in the future if there are features beyond object spread syntax that we believe would significantly improve code quality.

PR #2596, opened 2020-05-07
Remove object spread syntax, not supported in MS Edge

Harris-Miller:
Object spread syntax is not yet natively supported in MS Edge. I see that your ES5 build supports down to IE11, but I'd rather not have to alias `vega: 'vega/build-es5/vega'` in webpack just to get this to work in MS Edge, which the site I'm working on does support. We want ES6 imports and tree shaking from webpack for vega. So if you would kindly use this older syntax, it would be much appreciated.

### DIFF
--- a/packages/vega-parser/src/parsers/guides/guide-util.js
+++ b/packages/vega-parser/src/parsers/guides/guide-util.js
@@ -84,5 +84,5 @@ export function extendOffset(value, offset) {
   return !offset ? value
     : !value ? offset
     : !isObject(value) ? { value, offset }
-    : { ...value, offset: extendOffset(value.offset, offset) };
+    : Object.assign({}, value, { offset: extendOffset(value.offset, offset) });
 }


domoritz:
I would prefer https://github.com/vega/vega/pull/2595 instead. Do you agree?

Harris-Miller:
@domoritz #2595 Sorry to be the bearer of bad news here, but your MR only solves half the problem.

The current npm script `"build": "yarn rollup && node rollup -e && node rollup -m && babel build --out-dir build-es5 --config-file ../../babel.config.js",` what that does is, using `rollup`, bundles all the projects into the `build` folder. It _then_ runs that bundle through babel, which is set to support down to IE11, into a second bundle `build-es5`.

The problem though is _how_ you import Vega. Right now in the main vega package, in `package.json` we have this:
```
"main": "build/vega-node.js",
"module": "index",
```
What this means is that when you use a `require` statement to import vega, you'll get the `main` bundle `build/vega-node.js`, which is appropriate when you'rer in a node environment (for example, running unit tests on your project that imports vega), but if you're using webpack for example, and are using ES6 `import` statements, you'll get the module `index`, which just imports all the other packages, which will also look at their `module `import, which is just the `src` folder for each package.

So while your MR does change the main `build` folder to be run through Babel and do changes like Object spreading, the problem is that will only affect when you're importing the `main` and not the `module`. So any project using webpack won't be affected by this change, except that the newer syntax will fail in browsers that don't yet support it.

Many projects do a dual build where they run through babel their `src` file and export 2 folders, `cjs` and `esm`. The `cjs` folder has the `import` statements transpiled to `require` statements, and the `esm` folder keeps the `import` statements. Your `package.json` then looks like this:
```
"main: "cjs/index.js",
"module": "esm/index.js"
```
Rollup can still happen, and those bundles can still live in another output folder, and be used to be pushed to a CDN or something. But this method is preferred with the most modern js bundlers as they can tree shake with keeping the `import` statements, but still have the code transpiled down to the lowest supported browser/environment that the project maintainers decide on.

This task is a heavier lift than your MR, and requires more time and other considerations by the maintainers of Vega. In the mean time, this MR should be merged in the mean time to fix support for MS Edge, as it's just a syntax change and doesn't affect their build

domoritz:
Ahh, yes, so the babel build in #2595 only transpires the bundle but does not transpile all source files.

Let's dive a bit deeper into this to make sure we follow the right approach. What makes object spreads different from other post-es5 features? [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Browser_compatibility) says that it's actually supported by Edge. Could there be a new feature we use tomorrow that is not supported by some browser and therefore our source files are not compatible? Maybe the cleaner approach is to use whatever modern javascript we want in our source files and require consumers to run it through babel if they want to target a particular use case (we still have transpiled bundles)? The alternative I see is to transpile all files (as you also suggested) with babel and make a decision about what browsers we feel should be supported. Either way, someone might disagree with our assessment so maybe the former approach is easier, allows for more efficient code when needed, and isn't fundamentally different from the latter.

Thanks for helping me think through this!

Harris-Miller:
@jheer thanks for merging this. Was curious about your timeline on patch release?

domoritz:
There may be a soft release (no docs updates yet) this weekend.

Issue #1106, opened 2018-01-11
Implement non-eval based Vega execution path


nyurik:
For security reasons, many modern web sites lock down their Content Security Policy with the script no-eval CSP rule, which prevents the `new Function()` code.  Vega heavily relies on the dynamically generated expressions.  While this may provide some performance benefits, it would be good to have a possibly-slower expression interpreter alternative (configurable).

As an experiment, I have created an alternative sandbox approach, but this might not be suitable to some complex apps:  https://github.com/nyurik/csp-vega-test

jheer:
The (safe) use of a `Function` constructor underlies multiple Vega features, including not only expressions, but also compilation of visual encoding procedures. It's unlikely that we (at UW) will focus efforts on building an alternative interpreter for these functions. However, we're certainly open to reviewing 3rd party contributions.

In Vega 3, `Function` constructor calls arise due to a single line in the [vega-runtime repository](https://github.com/vega/vega-runtime/blob/master/src/expression.js#L9). Future extensions might start there, replacing the use of `Function.apply` with a custom JS parser/evaluator.

_For background context, the Vega parser generates output JS code for expressions and visual encoders, included within an output dataflow graph description. The runtime (typically invoked from vega-view) parses this description into a dataflow instance, including instantiation of functions._

Given the effort and potential performance implications of such changes, I recommend first following a sandbox approach if feasible.

einsweniger:
> The runtime (typically invoked from vega-view) parses this description into a dataflow instance, including instantiation of functions.

Just out of interest: once this is generated, could we dump that generated code into a file during the build process and use that in production? That could be an easier workaround than building a new interpreter.

jheer:
Hi @einsweniger, if you know ahead of time all the Vega specs you want to use in production, then yes,  you could run parse ahead of time. The output of `vega.parse` can be serialized to JSON then handed to the `vega.View` constructor directly. However, this will still result in the `Function` constructor being invoked by the runtime. More work would be needed to translate the parsed output and turn it into standard JS code.

einsweniger:
Thank you!

jornfranke:
I think this would be a very crucial feature in enterprise environments

dlaliberte:
[...]

PR #2598, opened 2020-05-08
Improve adherence to content security policy (CSP)

jheer:
**vega**

- Update to use refactored runtime context.

**vega-expression**

- Fix handling of `this` global variable.

**vega-functions**

- Add and export expression parser, support AST export option.
- Add internal scale functions for use by encoders.
- Fix scale dependencies: non-literal scale references should depend on all scales.

**vega-parser**

- Add options argument to `parse`.
- Add boolean `ast` parse option to enable AST output from expression parser.
- Refactor to use expression parser for synthesized encoder logic.
- Fix circular dependencies.

**vega-runtime**

- Refactor code, add optimized code generation for accessors and comparators.
- Refactor all non-standard code generation to the runtime context.
- Fix circular dependencies.

**vega-transforms**

- Update aggregation ops to no longer use the `Function` constructor.

**vega-util**

- Update `field`, `key`, and `compare` methods to no longer use the `Function` constructor.
- Add optional arguments to inject optimized code generators for accessors and comparators.

**vega-view**

- Update to use refactored runtime context.

See also #1106.


domoritz:
Awesome!

PR #2614, opened 2020-05-16
fix always undefined "this" for trap(view, ...)

terencehonles:
I managed to trigger an unrelated issue, but instead of reporting the issue I saw "Cannot read property 'error' of undefined", and this PR addresses the issue


jheer:
Thanks!

PR #2625, opened 2020-05-17
ci: switch to github actions

domoritz:
Github actions run on the same servers as github and therefore can checkout faster. I also found it much easier to setup deployments. We are using github actions for Vega-Lite, Vega-Embed, Vega-Tooltip, Vega-Themes, and the schema generator already.


domoritz:
Speed comparison:

Screen Shot 2020-05-17 at 13 51 51

Screen Shot 2020-05-17 at 13 51 46

PR #2630, opened 2020-05-18
fix: add missing color schemes

marcprux:
**vega-typings**
* add missing color schemes

The schemes that are new as of vega5 (https://vega.github.io/vega/docs/schemes/) were missing from the typing.


domoritz:
Thank you! Can you format the code with prettier?

domoritz:
We use a particular prettier config. I'll format the code.

domoritz:
Replaced by https://github.com/vega/vega/pull/2631

marcprux:
Sorry about that. Is the correct procedure to run the `format` script in `vega-typings/package.json`?

PR #2633, opened 2020-05-19
Add cursor target customization, use local cursor style by default.

### LINKED ISSUE #2439: Cursor signal and hover cursor property has major performance impact in Chrome
The way the setCursor function handles changing the cursor causes major performance issue in Chrome. This is due to it using `document.style.cursor = cursor`. Since this is being set on the entire document and since `cursor` is a cascading css property, the entire page must recalculate every style.
On complex layouts this causes major performance issues:

image

The purple areas are the impact of "Recalculate Style" on the entire document.

This is related to this chrome issue: https://bugs.chromium.org/p/chromium/issues/detail?id=664066

# Proposed solution

Since the docs `cursor`  state that you can set the cursor to persist even when not hovering the view you need to set it globally. I propose having a separate mark property `viewCursor` or `localCursor` that only changes the cursor for the current view. This would prevent the page style recalculation.

Alternatively change the behaviour of `cursor` to be local by default. But that would change backwards compatibility, even if it is likely what people need most of the time.

jheer:
Previously Vega updated the `cursor` style on the HTML document body. This has the nice effect of controlling the cursor even during interactions (such as drags) that may leave the Vega View component. However, it also can result in large performance penalties in Chrome, which re-evaluates CSS styles in response. This PR changes the default behavior of Vega to set the cursor locally on the Vega View component. If a global cursor is desired, the boolean config property `events.globalCursor` can be set `true` or the View method `globalCursor` can be invoked to change the setting at runtime.

**vega-view**

- Add `globalCursor` method and event configuration.
- Update to make the Vega view container the default cursor target.

Close #2439

### DIFF
--- a/packages/vega-view/src/View.js
+++ b/packages/vega-view/src/View.js
@@ -343,8 +343,9 @@ prototype.removeDataListener = function(name, handler) {
 prototype.globalCursor = function(_) {
   if (arguments.length) {
     if (this._globalCursor !== !!_) {
-      setCursor(this, null); // clear previous cursor target
+      const prev = setCursor(this, null); // clear previous cursor
       this._globalCursor = !!_;
+      if (prev) setCursor(this, prev); // swap cursor
     }
     return this;
   } else {
--- a/packages/vega-view/src/View.js
+++ b/packages/vega-view/src/View.js
@@ -1,6 +1,6 @@
 import {ariaLabel} from './aria';
 import background from './background';
-import cursor from './cursor';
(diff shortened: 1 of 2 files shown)


domoritz:
Do you consider this a breaking change?

jheer:
I don't consider this breaking, as it doesn't "break" any working specs, beyond limiting the default cursor scope to the Vega container. While arguable, the impact of the change is small and I believe the use case is also rare. Definitely a minor version change though, as the event config options are extended.

PR #2635, opened 2020-05-20
fix: fix typings for locale specification and view api

haldenl:
Ensures

1. `vega-typings` exports the `Locale` specification type, additionally splitting it into a `NumberLocale` and a `TimeLocale`.
2. ~`View` API typings for the `locale` function correctly match the `vega-view` implementation and documentation.~
3.  No type collision between `vega-typings/runtime` and `vega-typings/spec` by renaming the `Locale` definition in `runtime` to `LocaleFormatters`.

### DIFF
--- a/packages/vega-typings/types/runtime/index.d.ts
+++ b/packages/vega-typings/types/runtime/index.d.ts
@@ -1,4 +1,12 @@
-import { DataType, EncodeEntryName, Format, SignalValue, Spec } from '../spec';
+import {
+  DataType,
+  EncodeEntryName,
+  Format,
+  NumberLocale,
+  SignalValue,
+  Spec,
+  TimeLocale,
+} from '../spec';
 import { Renderers } from './renderer';
 import { Transform, Changeset } from './dataflow';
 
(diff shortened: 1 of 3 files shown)


packages/vega-typings/types/runtime/index.d.ts
```diff
@@ -52,8 +60,8 @@ export class View {
   loader(): Loader;
   loader(loader: Loader): this;
 
-  locale(): Locale;
-  locale(locale: Locale): this;
+  locale(): LocaleFormatters;
+  locale(time: TimeLocale, number: NumberLocale): this;
```
jheer, line 64: This change is incorrect. The Dataflow (View’s Patent class) takes a combined locale formatters object as input: https://github.com/vega/vega/blob/master/packages/vega-dataflow/src/dataflow/Dataflow.js#L85

haldenl: Ahh gotcha. I got confused by the [Locale API](https://vega.github.io/vega/docs/api/locale/) docs. Reverted.

domoritz: Let's add the correct typings for `vega.locale` as well then.

PR #2636, opened 2020-05-20
docs: add / improve docs for locale

haldenl:
I also noticed a broken / missing locale link in the [`Config` docs](https://vega.github.io/vega/docs/config/) under `View Properties` so I added a page and modified the locale documentation under the API docs to reduce redundancy—feel free to chuck this change if you feel it is unnecessary.


haldenl:
makes sense to me! fixed

PR #2639, opened 2020-05-20
fix: simplify voice over text for titles and subtitles

domoritz:
This pull request changes the description of titles to be just the title. https://developer.apple.com/videos/play/wwdc2019/254/ suggests that descriptions should be as concise as possible so we may not want to repeat that the text is the title or subtitle (similar to buttons in the video).

Marked as draft since I need to think more about whether text should even have an aria label.

### DIFF
--- a/packages/vega-scenegraph/src/util/aria.js
+++ b/packages/vega-scenegraph/src/util/aria.js
@@ -37,11 +37,11 @@ const AriaGuides = {
   'legend': {desc: 'legend', caption: legendCaption},
   'title-text': {
     desc: 'title',
-    caption: item => `Title text '${titleCaption(item)}'`
+    caption: item => titleCaption(item)
   },
   'title-subtitle': {
     desc: 'subtitle',
-    caption: item => `Subtitle text '${titleCaption(item)}'`
+    caption: item => titleCaption(item)
   }
 };
 


jheer:
LGTM. The advantage (or so it seemed to me at the time) of doing this with role / aria-label is that the screen reader identifies the element as title and subtitle text (which still is read out loud at the end thanks to the aria-roleDescription), as opposed to arbitrary text. That seems even more useful in cases where the title currently comes after other chart elements in the SVG.

domoritz:
I agree that the longer descriptions can be useful and is somewhat consistent with how we describe axes ("X-axis for a discrete scale with 8 values: A, B, C, D, E, ending with H") and legends ("Symbol legend for stroke color with 3 values: USA, Japan, Europe") but also different since we have aria role descriptions for title and subtitles separately. We could move the type to the role description (e.g. have VO say "stroke color with 3 values: USA, Japan, Europe, symbol legend" instead of just "Symbol legend for stroke color with 3 values: USA, Japan, Europe, legend") but now it's getting a lot harder to follow what's happening when the descriptions get long. Or maybe we shouldn't customize the role descriptions as much as we are doing right now. 

Let's think about this a bit more and also ask VO users about what they prefer.

PR #2646, opened 2020-05-22
Removed trailing commas which break Terser Webpack Plugin process

### LINKED ISSUE #2645: Terser Webpack Plugin exits with error due to trailing comma in vega-parser package
### Problem Description
Some compilers use the ecma option in the Terser plugin to produce ES6 output. It usually is a set like `ecma: esm ? 6 : 5` and that works fine. However, the trailing commas on following two lines makes it impossible to use anything but ES8.

https://github.com/vega/vega/blob/master/packages/vega-parser/src/parsers/guides/axis-config.js#L73
https://github.com/vega/vega/blob/master/packages/vega-parser/src/parsers/guides/axis-config.js#L84

The Terser option is explained [here](https://github.com/terser/terser#parse-options).

### How to Reproduce
Create a new React project with Nx. Try getting a production build.

### Vega Version
5.12.2

Screen Shot 2020-05-22 at 19 06 05

### Solution
These two trailing commas do not have any particular benefit, so I would suggest removing them. I will open a PR for this.

armanozak:
Fixes #2645

### DIFF
--- a/packages/vega-parser/src/parsers/guides/axis-config.js
+++ b/packages/vega-parser/src/parsers/guides/axis-config.js
@@ -70,7 +70,7 @@ export default function(spec, scope) {
       xy[key] = ifX(
         orient,
         fallback(key, config.axisX, axis, style),
-        fallback(key, config.axisY, axis, style),
+        fallback(key, config.axisY, axis, style)
       );
     }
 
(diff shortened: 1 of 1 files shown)


domoritz:
Can you add a rule to eslint to ensure that we maintain the style?

jheer:
Thanks @armanozak! I will add an eslint rule myself.

armanozak:
Thanks @jheer, and thanks for Vega. 👏

PR #2658, opened 2020-05-27
Add expression interpreter.

jheer:
This PR adds interpreter support for Vega expressions that is [Content Security Policy (CSP)](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP)-compliant. By default, the Vega parser performs code generation for parsed Vega expressions, and the Vega runtime uses the [Function constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/Function) to create JavaScript functions from the generated code. Although the Vega parser includes its own security checks, the runtime generation of functions from source code nevertheless violates security policies designed to prevent cross-site scripting.

This PR provides an optional interpreter that evaluates expressions by traversing an [Abstract Syntax Tree (AST)](https://en.wikipedia.org/wiki/Abstract_syntax_tree) for an expression and performing each operation in turn.  Use of the interpreter enables compliance with CSP, but can incur a performance penalty. In tests of initial parse and dataflow evaluation times, the interpreter is on average ~10% slower. Interactive update performance may incur higher penalties, as it is often more expression-heavy and also amortizes the one-time cost of Function constructor parsing.

**vega**

- Update scene test code.

**vega-functions**

- Fix scale function to not special case `undefined` input. This ensures identical semantics with the internal `_scale` helper function used by code-generated encoders.

**vega-interpreter**

- Add new vega-interpreter package.

**vega-runtime**

- Add pluggable expression evaluators.

**vega-view**

- Add View constructor option `expr` to pass in a custom expression evaluator.


jheer:
**Usage Example**

To use the interpreter, three steps must be taken:

1. Load the Vega interpreter module, as either using a separate HTML script tag (once the vega-interpreter package has been published) or as part of a custom Vega build.
2. Invoke the Vega parser with the `{ast: true}` option to enable inclusion of parsed ASTs in the output.
3. Pass the interpreter as an option to the Vega View constructor. The underlying runtime will be configured to use the alternative expression evaluator.

```js
const spec; // Vega specification to show.

// Parse the Vega specification with AST output enabled
// We pass a null config as the second argument
const runtimeSpec = vega.parse(spec, null, {ast: true});

// Instantiate a Vega View with the interpreter as constructor option
const view = new vega.View(runtimeSpec, {
  expr:      vega.expressionInterpreter, // use interpreter
  renderer:  'canvas',  // renderer (canvas or svg)
  container: '#view',   // parent DOM container
  hover:     true       // enable hover processing
});

view.runAsync();
```

Issue #2665, opened 2020-05-28
Vega logs `Dataflow invoked, but nothing to do.` even if log level is Warn


domoritz:
https://github.com/vega/vega/blob/92c7caccf22ee51cd5b059dedfede073da1d9814/packages/vega-dataflow/src/dataflow/run.js#L46 is not guarded by the level. 

Should the level maybe handled in the logger instead of the code?

jheer:
In general, the level should be handled by the logger. However, the reason the dataflow guards some of the calls here is for performance reasons: to avoid marshaling the debug information if it is not supposed to be shown.

PR #2667, opened 2020-05-28
fix: only log message when log level is high enough

domoritz:
Fixes https://github.com/vega/vega/issues/2665


jheer:
The "Dataflow invoked, but nothing to do." message can be handled by the logger.

The other guard was correct before: The timestamp is needed for both Info and Debug level logging, and the logger then controls if the debug message should be shown or not.

Another guard (not modified by this PR) then modulates whether or not potentially costly debug information is marshaled before invoking the logger debug method.

domoritz:
I see. I filed https://github.com/vega/editor/issues/710 to respect the log level in the custom logger in the editoe.

PR #2733, opened 2020-06-25
chore: add dev container configuration

domoritz:
If you start vega with vscode, it will automatically suggest to open Vega in a dev container running Linux wth node 12. The nice thing is that we can use it to reproduce test failures coming from the OS.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -4,8 +4,8 @@
 #-------------------------------------------------------------------------------------------------------------
 
 # To fully customize the contents of this image, use the following Dockerfile instead:
-# https://github.com/microsoft/vscode-dev-containers/tree/v0.128.0/containers/javascript-node-12/.devcontainer/Dockerfile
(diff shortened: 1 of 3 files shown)


jheer:
I like the idea of being able to test this, and glad that it seems to be opt-in. I'm not sure I want an "automatic suggestion" though. Is there an alternative setup with a more explicit interaction for enabling the container?

domoritz:
I removed the follow up changes I tried to fix the CI issues on master. This pull request now only has the dev container setup.

domoritz:
> I'm not sure I want an "automatic suggestion" though. Is there an alternative setup with a more explicit interaction for enabling the container?

It's not automatically opening it but there is a prompt in VSCode that you can dismiss or silence completely. If you silence it, you can open in the container through the quick open command.

domoritz:
This is ready for review.

domoritz:
We don't really need this anymore since tests pass on different operating systems again.

Issue #1940, opened 2019-08-21
Add Turbo color scheme


domoritz:
See https://ai.googleblog.com/2019/08/turbo-improved-rainbow-colormap-for.html for details. Mike made an observable for it: https://observablehq.com/@mbostock/turbo.

domoritz:
Now available in D3: https://github.com/d3/d3-scale-chromatic/blob/master/README.md#interpolateTurbo

PR #2740, opened 2020-06-29
Add turbo and cividis schemes

mcnuttandrew:
This PR addresses #1940, adding turbo and cividis color schemes. The schemes were generated using the following script: 

```js
const cividis = require("./sequential-multi/cividis").default;
const turbo = require("./sequential-multi/turbo").default;
const color = require("d3-color");

const vals = [];
for (let i = 0; i < 256; i++) {
  vals.push(
    color
      .rgb(turbo(i / 256))
      .formatHex()
      .slice(1)
  );
}
console.log(vals.join(""));
```
Run from inside of d3-scale-chromatic. 256 was selected because it appears that viridis was generated in a similar manner. There's a bunch of noise in this PR (that i'm happy to undo if you feel strongly) induced by autoformating on save within some of the examples. I added the color schemes to the heatmap examples, which also provided a way to verify that the palettes worked as desired


domoritz:
Thank you @mcnuttandrew! Can you send a separate pull request to update https://vega.github.io/vega/docs/schemes/?

Looking at the definitions, other color schemes fewer stops. Do you think we can reduce the size of the definitions without loss of quality?

Screen Shot 2020-06-29 at 12 46 19

mcnuttandrew:
> Can you send a separate pull request to update https://vega.github.io/vega/docs/schemes/?

Yep will do!

> Looking at the definitions, other color schemes fewer stops. Do you think we can reduce the size of the definitions without loss of quality?

I don't have a great intuition for how many steps are necessary for the scales to be sufficiently described, I reduced the resolution from 256 steps to 64, which (baselessly) seems reasonable to me? I'm happy to copy the precedent set by the other schemes if you know better

mcnuttandrew:
Is there a script hanging around for generating the docs/docs/scheme.md file? It seems like it follows a really regular structure. While I'm happy to emulate it, I'd much rather use a script if one exists?

domoritz:
I don't think there is one online but @jheer might have one.

jheer:
Thanks @mcnuttandrew! Happy to add these.

Before merging, it would be great to also:

1) Remove the unrelated re-formatting noise from the updated examples.
2) Add the new schemes to the documentation. The schemes page was originally generated with a script, but I do not have a proper build system around this. So the fastest route it probably just to copy/paste/edit from the existing docs.

mcnuttandrew:
@jheer i've removed the autoformatting noise and I've added these schemes to the docs. (Is there anything required for adding things to the docs beside modifying the markdown files?)

Rather than deal with the error-prone process of manually rewriting the html i wrote a short script to produce the scheme (tested on the viridis output). 

```js
const name = "turbo";
const colorFunc = (frac) => color.rgb(turbo(frac)).formatHex();
// for the continuous svg
const steps = [...new Array(11)]
  .map((_, idx) => colorFunc(idx / 10))
  .map((color, idx) => {
    return `        <stop offset="${idx * 10}%" stop-color="${color}"></stop>`;
  });
// for the discrete swatches
const swatches = [...new Array(9)].map((_, idx) => {
  return `
<div class="swatch">
  <strong>${idx + 1}</strong>
${[...new Array(idx + 1)]
  .map((__, jdx) => colorFunc((jdx + 1) / (idx + 2)))
  .map((color) => {
    return `    <div title="${color}" style="background: ${color};"></div>`;
  })
  .join("\n")}
</div>`;
});

const output = `
<div class="scheme continuous">
<a name="${name}" href="#${name}">#</a> <strong>${name}</strong> <a class="toggle" href='javascript:toggle("#discrete-${name}")'>View Discrete</a>
<svg viewBox="0,0,1,1" preserveAspectRatio="none">
  <defs>
    <linearGradient id="gradient-${name}">
${steps.join("\n")}
    </linearGradient>
  </defs>
  <rect fill="url(#gradient-${name})" x="0" y="0" width="1" height="1"></rect>
</svg>

${swatches.join("\n")}
</div>
</div>;
`;
console.log(output);
```

Let me know if there's anything else i should do!

domoritz:
Thank you @mcnuttandrew!

jheer:
Thanks @mcnuttandrew!!!

PR #2742, opened 2020-06-30
feat: add types for Vega-Expression package

Zwyx:
## This PR improves the definition of the types for Vega-Expression

Fix `constants` in the Vega-Expression documentation

`constants` is an object, not a function; it doesn't take arguments

---

Improve types for Vega-Expression

- Correct the types for `parse` and `codegen`
- Add the types for `constants`, `functions`, and `ASTNode`
- Add documentation helping development:
image

---

Please let me know if you'd like me to change anything.

### DIFF
--- a/packages/vega-expression/index.d.ts
+++ b/packages/vega-expression/index.d.ts
@@ -1,21 +1,51 @@
-// TODO: add missing types
+/** Parse a JavaScript *expression* string and return the resulting abstract syntax tree in the ESTree format */
+export function parse(expression: string): object;
 
-export function parse(expression: string): any;
+/**
+ * Create a new output code generator configured according to the provided options:
+ * - `constants` — a hash of allowed top-level constant values
+ * - `functions` — a function that is given an AST visitor instance as input and returns an object of allowed functions
+ * - `blacklist` — an array of variable names that may not be referenced within the expression scope
+ * - `whitelist` — an array of variable names that may be referenced within the expression scope
+ * - `fieldvar` — the name of the primary data input argument within the generated expression function
+ * - `globalvar` — the name of the variable upon which to lookup global variables
+ */
+export function codegen(options: {
+  /** A hash of allowed top-level constant values */
+  constants?: { [cn: string]: string };
 
-export function codegen(params: {
-  constants?: object;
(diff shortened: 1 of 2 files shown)


Zwyx:
No worries, I'll update the PR and let you know when it's done.

Zwyx:
I shortened the descriptions and added links to the online documentation:
```
@see https://github.com/vega/vega/tree/master/packages/vega-expression#parse
```

Also, although the documentation mentions that `globalvar` is required in the options for `codegen`, it's actually not. Not providing it does not error and just leads to a variable `foo` being returned as `undefined["foo"]`. For an expression containing no variables, there should be no need to provide `globalvar`, so I think we should make it optional like the others. What's you opinion?

domoritz:
Thank you. Let’s not use @see since we don’t use it anywhere else.

Zwyx:
Done. I'm not sure if you wanted me to leave the links or not, let me know if you want them back.

domoritz:
Thanks. Looks good to me.

packages/vega-expression/index.d.ts
```diff
@@ -1,21 +1,51 @@
...
+/**
+ * Create a new output code generator configured according to the provided options:
+ * - `constants` — a hash of allowed top-level constant values
+ * - `functions` — a function that is given an AST visitor instance as input and returns an object of allowed functions
+ * - `blacklist` — an array of variable names that may not be referenced within the expression scope
+ * - `whitelist` — an array of variable names that may be referenced within the expression scope
+ * - `fieldvar` — the name of the primary data input argument within the generated expression function
+ * - `globalvar` — the name of the variable upon which to lookup global variables
+ */
+export function codegen(options: {
```
domoritz, line 13: Can you extract options to a new type (`CodegenOptions`) so we can remove the duplication between the docs below and the function comment?

Zwyx: Done.

FYI,
- the options' explanations won't appear anymore when hoovering `codegen`;
- creating `CodegenOptions` wasn't necessary, we could have just removed the doc from the `codegen` function and leave the options where they were (as it is still now for the returned type of this function).

domoritz:
Thank you @Zwyx!

PR #2743, opened 2020-07-01
ci: run tests on pull requests

domoritz:
With this change, we also run tests on pull requests from forks.


domoritz:
Hmm, now we run every test twice. That seems wrong. Will have to iterate on this. 

Screen Shot 2020-06-30 at 21 51 49


See https://github.com/vega/vega-lite/issues/5076

PR #2758, opened 2020-07-14
fix: label transform do not use undefined bitmap

chanwutk:
Fix of [this issue](https://github.com/vega/vega-label/issues/21) from Vega-Label repository.
Prevent labeling algorithm from using `bitmap1` if it is not defined.


domoritz:
Rebased to master

packages/vega-label/src/util/placeAreaLabel.js
```diff
@@ -1,138 +0,0 @@
-import {textMetrics} from 'vega-scenegraph';
```
domoritz, line 1: Did I delete this file when I rebased or did you remove it @chanwutk?

chanwutk: I will check.

chanwutk: I deleted it. The file is not used anywhere.

PR #2759, opened 2020-07-14
fix: always resolve item for all events before calling handlers

drgould:
For area and line marks the data point reported to all event handlers was always the first point in the series regardless where along the area or line the cursor was positioned. However, for tooltips the values displayed were correct. After looking at the code a bit I found that the `resolveItem` function was only being called in the tooltip handler so this moves the resolution up one level to the base event handlers for both SVG and Canvas so everything has the correct data.

### DIFF
--- a/packages/vega-scenegraph/src/CanvasHandler.js
+++ b/packages/vega-scenegraph/src/CanvasHandler.js
@@ -9,6 +9,7 @@ import {
 import point from './util/point';
 import {domFind} from './util/dom';
 import {inherits} from 'vega-util';
+import resolveItem from './util/resolveItem';
 
 export default function CanvasHandler(loader, tooltip) {
   Handler.call(this, loader, tooltip);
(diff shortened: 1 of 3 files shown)


jheer:
Thanks for the PR! We will have to think about this a bit more before merging though, as it changes the behavior and semantics of Vega interactions. So I think at minimum it would have to wait for a major version change. For example, the resolved items for lines are the points, and the tooltip picker requires that the mouse pointer be within a specific radius of a point. The current behavior, where one can select a line by mousing over any point on that line (and have it resolve to the first data point in the line), would no longer work.

jheer:
Thinking about this more, I realized that the change in this PR might break other parts of Vega as well. For example, properties such as the fill or stroke color of line and area marks are determined by the values of the first item. Let's say we want to change them upon hover: there is only a single geometric element (line or area) so the hover handler should return the first element and update its properties, even if the mouse was over a subsequent element. In short, I think we will need more complicated logic to handle resolution of different data points without breaking existing Vega mechanisms. As a result, I'm closing out this PR for now, but I do very much appreciate the contribution!

Issue #2777, opened 2020-07-30
transform label suggestion


yuicer:
Please:

- [x] Check for duplicate issues. Please file separate requests as separate issues on GitHub.
- [x] Clearly describe the goal of the feature request. Why is it needed?
- [x] Describe an envisioned solution. What do you want the new behavior to be?
- [x] If applicable, share mockup images or links to examples illustrating the desired output.
- [x] If applicable, share JSON spec designs for how the behavior would be made available in Vega.

@chanwutk 

well.... I think the transform-label here show choose the pulse.ALL ? 

Cause I fond if the label mark property [x, y] is a scale effected by another signal like dragger or something.  transform will not run itself again

![image](https://user-images.githubusercontent.com/23371475/88878765-328a4200-d25b-11ea-934a-233e5f48a0c3.png)

chanwutk:
Hi @yuicer,
Thank you! This is a very great suggestion.

yuicer:
@chanwutk  I'll  will submit a mr later

yuicer:
@chanwutk 

well.... there seems to be a new problem with pulse.ALL......

if the label mark has the hover encode. the transform will run again . and the label position [x,y] will be set to a new position based on its last compute position

which leads to the label jump everywhere.......

I closed the PR

chanwutk:
I see. Thank you!

chanwutk:
I see. Could you post an example of a spec where label-transform does not update itself when it is supposed to?

yuicer:
@chanwutk  

sorry

[demo](https://github.com/yuicer/vega-label-transform-demo) here

yuicer:
hover the text mark and ......

chanwutk:
Thank you for the demo,
This seems to be the behaviors that we are looking for:
|Updates|`label-transform` would rerun|
|:-|:-|
| When the source data of the text mark is updated | rerun |
| When properties of the text mark (only the ones that relate to size and position) are updated not from `label-transform` | rerun |
| When properties of the text mark are updated from `label-transform` | not rerun |

What do you think? @yuicer @kanitw @domoritz

yuicer:
For my mind... it seems should to have a status to judge the label should rerun **label-transform** from real chart logic

at least, hover one label mark should not effect other label mark

PR #2778, opened 2020-07-30
feature: transform-label should respond to mark property changes

yuicer:
issue [#2777 ](https://github.com/vega/vega/issues/2777)


yuicer:
well.... there seems to be a new problem with pulse.ALL......

if the label mark has the hover encode. the transform will run again . and the label position [x,y] will be set to a new position based on its last compute position

which leads to the label jump everywhere

chanwutk:
I see. Thank you!

PR #2795, opened 2020-08-10
[WIP] Fix: include gradients when generating scenegraph JSON

Saba9:
Include keys required for gradients (gradient, x1, y1, stops, offset, color) when serializing a scenegraph as JSON.

The "offset" key is used both by gradients and ~~text~~ group mark items which is why so many scenegraphs have been changed. Not sure if we want to exclude "offset" when serializing ~~text~~ group mark items or not.


packages/vega/test/scenegraphs/area.json
```diff
@@ -473,6 +473,7 @@
               ],
               "x": 0.5,
               "y": 200.5,
+              "offset": 0,
```
domoritz, line 476: Can you assume that the default is 0 and keep the scenegraph leaner?

Saba9: Not sure. I've tried to find where offset is being set on group items but haven't found it yet. Alternatively, we could modify the JSON serialization method to exclude offset except when serializing gradients.

PR #2808, opened 2020-08-17
Prevent scaleX/Y SVG transforms on non-path marks.

### LINKED ISSUE #2738: Canvas renderer does not size symbols correctly
I noticed a peculiar bug while trying to solve [this problem](https://stackoverflow.com/questions/62606655/how-to-make-a-chart-with-svg-marker-whose-size-increases-in-only-one-aspect-or-s). The graph that the editor renders is different from what is getting exported. The rendered graph is correct and the exported graph is wrong.

**Vega Spec**

```json
{
  "description": "A simple bar chart with embedded data.",
  "background": "white",
  "padding": 5,
  "height": 400,
  "style": "cell",
  "data": [
    {
      "name": "source_0",
      "values": [
        {"a": "A", "b": 1.5, "c": 0},
        {"a": "B", "b": 0.5, "c": 0},
        {"a": "C", "b": 10, "c": 0},
        {"a": "D", "b": 1, "c": 0}
      ]
    },
    {
      "name": "data_0",
      "source": "source_0",
      "transform": [
        {
          "type": "filter",
          "expr": "isValid(datum[\"b\"]) && isFinite(+datum[\"b\"])"
        }
      ]
    }
  ],
  "signals": [
    {"name": "x_step", "value": 20},
    {
      "name": "width",
      "update": "bandspace(domain('x').length, 1, 0.5) * x_step"
    }
  ],
  "marks": [
    {
      "name": "marks",
      "type": "symbol",
      "style": ["path"],
      "from": {"data": "data_0"},
      "encode": {
        "update": {
          "opacity": {"value": 0.7},
          "fill": {"value": "red"},
          "stroke": {"value": "black"},
          "strokeOpacity": {"value": 1},
          "strokeWidth": {"value": 1},
          "shape": {"value": "M -1 0 L0 -10 L1 0"},
          "ariaRoleDescription": {"value": "point"},
          "description": {
            "signal": "\"a\" + \": \" + (isValid(datum[\"a\"]) ? datum[\"a\"] : \"\"+datum[\"a\"]) + \"; \" + \"b\" + \": \" + (format(datum[\"b\"], \"\"))"
          },
          "x": {"scale": "x", "field": "a"},
          "y": {"scale": "y", "field": "c"},
          "scaleY": {"field": "b", "type": "quantitative"}
        },
        "fill": {"signal": "gradient('red', [0,0], [1,0])"}
      }
    }
  ],
  "scales": [
    {
      "name": "x",
      "type": "point",
      "domain": {"data": "data_0", "field": "a", "sort": true},
      "range": {"step": {"signal": "x_step"}},
      "padding": 0.5
    },
    {
      "name": "y",
      "type": "linear",
      "domain": {"data": "data_0", "field": "b"},
      "nice": true,
      "zero": true
    },
    {
      "name": "size",
      "type": "linear",
      "domain": {"data": "data_0", "field": "b"},
      "range": [0, 361],
      "zero": true
    }
  ],
  "axes": [
    {
      "scale": "y",
      "orient": "left",
      "gridScale": "x",
      "grid": true,
      "domain": false,
      "labels": false,
      "aria": false,
      "maxExtent": 0,
      "minExtent": 0,
      "ticks": false,
      "zindex": 0
    },
    {
      "scale": "x",
      "orient": "bottom",
      "grid": false,
      "title": "a",
      "labelAlign": "right",
      "labelAngle": 270,
      "labelBaseline": "middle",
      "labelOverlap": true,
      "zindex": 0
    },
    {
      "scale": "y",
      "orient": "left",
      "grid": false,
      "title": "b",
      "labelOverlap": true,
      "zindex": 0
    }
  ]
}
```
In the editor it looks like so -
image
However, when I export it OR see it as a preview via VSCode Extension, then I get the following -
 
visualization

jheer:
**vega-scenegraph**

- Fix leaking SVG scaleX/Y transforms on non-path marks.

Fix #2738.

### DIFF
--- a/packages/vega-scenegraph/src/util/svg/transform.js
+++ b/packages/vega-scenegraph/src/util/svg/transform.js
@@ -14,8 +14,13 @@ export function translateItem(item) {
   return translate(item.x || 0, item.y || 0);
 }
 
+export function rotateItem(item) {
+  return translate(item.x || 0, item.y || 0)
+    + (item.angle ? ' ' + rotate(item.angle) : '');
+}
+
 export function transformItem(item) {
   return translate(item.x || 0, item.y || 0)
     + (item.angle ? ' ' + rotate(item.angle) : '')
-    + (item.scaleX || item.scaleY ? ' ' + scale(item.scaleX || 1, item.scaleY || 1) : '');   
+    + (item.scaleX || item.scaleY ? ' ' + scale(item.scaleX || 1, item.scaleY || 1) : '');
 }
--- a/packages/vega-scenegraph/src/marks/markItemPath.js
+++ b/packages/vega-scenegraph/src/marks/markItemPath.js
@@ -3,13 +3,13 @@ import context from '../bound/boundContext';
 import {intersectPath} from '../util/intersect';
 import {drawAll} from '../util/canvas/draw';
 import {pickPath} from '../util/canvas/pick';
-import {transformItem} from '../util/svg/transform';
+import {rotateItem} from '../util/svg/transform';
 import {DegToRad} from '../util/constants';
 
 export default function(type, shape, isect) {
 
   function attr(emit, item) {
-    emit('transform', transformItem(item));
+    emit('transform', rotateItem(item));
     emit('d', shape(null, item));
   }
 


packages/vega-scenegraph/src/util/svg/transform.js
```diff
@@ -14,8 +14,13 @@ export function translateItem(item) {
...
+export function rotateItem(item) {
+  return translate(item.x || 0, item.y || 0)
+    + (item.angle ? ' ' + rotate(item.angle) : '');
+}
+
 export function transformItem(item) {
   return translate(item.x || 0, item.y || 0)
     + (item.angle ? ' ' + rotate(item.angle) : '')
-    + (item.scaleX || item.scaleY ? ' ' + scale(item.scaleX || 1, item.scaleY || 1) : '');   
+    + (item.scaleX || item.scaleY ? ' ' + scale(item.scaleX || 1, item.scaleY || 1) : '');
```
domoritz, line 25: Can `scaleX` or `scaleY` ever be 0 on purpose? Maybe it could happen when you want to animate changes.

jheer: That could be a useful addition if/when animation is added, but should be the subject of a separate issue / PR.

PR #2817, opened 2020-08-18
Refactoring: Use const in packages/vega-dataflow/src/ChangeSet.js

lgrammel:
Coding best practices. Use const whenever variables are not modified.


domoritz:
Please combine with #2819

PR #2819, opened 2020-08-18
Refactoring:Use const in packages/vega-functions/test/scale-gradient-test.js

lgrammel:
Coding best practices. Use const whenever variables are not modified.


domoritz:
Thanks for your contribution. I think we could set up eslint to automatically catch these improvements (Vega-Lite for example does it already). Would you be interested in helping out with that?

domoritz:
I ran eslint in https://github.com/vega/vega/pull/2820.

PR #2820, opened 2020-08-18
style: prefer const

domoritz:
There are still many places where we can't use const because multiple variables are defined with a single `var/let` but we can fix those later.

Maybe we can run a codemod (https://github.com/facebook/codemod) to upgrade all of the code automatically. But at the same time we should probably switch to prettier and babel.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -18,6 +18,7 @@ module.exports = {
     'no-fallthrough': ['error', { commentPattern: 'break omitted' }],
     'semi': 'error',
     'quotes': ['error', 'single', { avoidEscape: true }],
+    'prefer-const': 'warn',  // TODO: make error when we don't define multiple variables at once anymore
     'sort-imports': ['error', {
       ignoreCase: false,
       ignoreDeclarationSort: true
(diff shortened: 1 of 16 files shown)


domoritz:
cc @lgrammel

lgrammel:
Hey! Maybe some background on my pull requests, it was a dry run for an automated tool (Jeff suggested trying it out on Vega). It's similar to codemods. I could actually implement a split of `let` statements that define multiple variables, some of which could be `const` if there's interest?

domoritz:
Awesome. Yeah, I would love to have a more consistent code style between the different Vega projects. 

I recently submitted a pull request to d3 scales to use let/const (https://github.com/d3/d3-scale/pull/212). The other d3 repos still need to be updated (https://github.com/d3/d3/issues/3435) so there may be interest beyond Vega for your tool.

lgrammel:
Cool, thanks for the pointer. I'll see if this works for the d3 repos as well.

lgrammel:
@domoritz I've run 3 codeshifts (var -> let/const, arrow functions, object property shorthands) that would result in roughly ~500 changes. Is this something it would make sense for me to create PRs for? And if so, how would those PRs ideally be sized (e.g. per file/codeshift, for all files at once (one PR per codeshift), some intermediate slicing (e.g. per directory)?

domoritz:
I'd be happy with one pull request for everything since all the changes are of the same kind. But I'll defer to @jheer for deciding on what goes into master.

domoritz:
Also see https://github.com/vega/vega/pull/2827

Issue #2816, opened 2020-08-18
Vega expression should not use implicit any


domoritz:
```
node_modules/vega-expression/index.d.ts:9:16 - error TS7006: Parameter 'astVisitor' implicitly has an 'any' type.

9   functions?: (astVisitor) => { [fn: string]: string | ((args) => string) };
                 ~~~~~~~~~~

node_modules/vega-expression/index.d.ts:9:58 - error TS7006: Parameter 'args' implicitly has an 'any' type.

9   functions?: (astVisitor) => { [fn: string]: string | ((args) => string) };
                                                           ~~~~

node_modules/vega-expression/index.d.ts:25:52 - error TS7006: Parameter 'ast' implicitly has an 'any' type.

25 export function codegen(options: CodegenOptions): (ast) => {
                                                      ~~~

node_modules/vega-expression/index.d.ts:40:27 - error TS7006: Parameter 'codegen' implicitly has an 'any' type.

40 export function functions(codegen): {
                             ~~~~~~~

src/compile/data/expressions.ts:32:7 - error TS2339: Property 'visit' does not exist on type 'object'.

32   ast.visit((node: any) => {
         ~~~~~


Found 5 errors.

```

PR #2821, opened 2020-08-18
fix: update expression typings to be compatible with Vega-Lite

domoritz:
Fixes https://github.com/vega/vega/issues/2816

Needed for https://github.com/vega/vega-lite/pull/6802


domoritz:
Once this is released, I can update Vega-Lite and also make a Vega editor release.

PR #2822, opened 2020-08-18
fix!: replace blacklist/whitelist with forbidden/allowed

domoritz:
To reduce stigma and racial stereotyping in our code base, I am proposing to replace blacklist/whitelist with forbidden/allowed.

This is a breaking change.


jheer:
Thank you! Will merge this when we move to Vega v6.

domoritz:
We could probably tread this change as a breaking change only for the interpreter if we wanted to release it earlier.

jheer:
Good idea!

PR #2826, opened 2020-08-23
feat: compile Vega packages with microbundle

domoritz:
See https://github.com/developit/microbundle

Not quite ready yet but the idea is that we would compile all packages with rollup and automatically transpile modern JavaScript. 

What I don't like yet
* [ ] Lot's of globals declared (just like before, though)
* [ ] Some file names change
* [ ] Extend (for vega) doesn't work yet


domoritz:
Better version https://github.com/vega/vega/pull/2827

Issue #1165, opened 2018-02-19
Building Vega Code Community


nyurik:
This is a meta task to propose and discuss large code changes in order to build and sustain a coding community around Vega project.  The main goal is to make new/accidental code contributors productive with the minimum efforts.  This means that
* Code has to be easy to understand and follow by a newcomer with a reasonable level of coding skills.
* There should be a sufficient documentation for coders to get started and contribute
* Continuous integration should validate any new contributions, and encourage high code coverage

If the discussion for each item becomes too big, lets create a sub issue and link from here.

## Proposals
- [ ] Join all Vega code in the single github repo (e.g. with [Lerna](https://github.com/lerna/lerna)), while keeping sub-projects as separate NPM libs.
- [ ] Setup build system to use rollup + babel -- allows contributors to use the latest JS, while also establishing a clear support for older browsers. (e.g. declare support for Firefox version N, and babel will ensure that no newer features are accidentally used in the compiled result)
- [ ] Configure project badges to indicate build and test passing, related versions, etc
- [ ] Configure continuous integration to
  - [ ] automatically compile and test all pull requests
  - [ ] show coverage results
  - [ ] show performance improvements / degradations
- [ ] Provide older version support by publishing `vega-1`, `vega-1.5`, and `vega-2` npm packages.

domoritz:
I agree with this proposal but would love to see us using typescript. In addition to the above benefits, we get types, .d.ts files, and schema generation.

nyurik:
@domoritz I do agree that TS would bring additional benefits to the above.  Being a strict superset of JavaScript, can TS be introduced at any point? If so, I would propose that we postpone it until all stylistic tasks are implemented first, before adding new code structures.  For example, it would be good to set up proper CI to run all of the existing tests on each PR, and to establish eslint rules, etc.

Another thing that we still have to consider is that our goal is to attract a wide and stable community of devs.  JavaScript is the lowest common denominator here, lowering the barrier of entry, while TS is an extra that some may not be very comfortable with. I think we can migrate to TS, but I don't think it blocks us from doing the above proposals.  What do you think?

saulshanabrook:
I am new to the community but am starting to need to be able to understand parts of the Vega codebase in order to write custom transforms. I would +1 the idea of switching to TypeScript (ex #1706), which can be really helpful to quickly grok a codebase without having to ask the existing maintainers lots of questions. 

[...]

PR #2827, opened 2020-08-23
feat: rollup everything

### LINKED ISSUE #862: Allow ES6 + use Babel for ES5 support
Vega seems to be using a lot of ES6 features, such as object destructuring assignments, and yet it use uglifyjs, not uglifyjs-es6. Also, there is a note that IE10 & 11 are also supported [here](https://vega.github.io/vega/usage/#ie), but IE11 does not support many other features either. According to http://gs.statcounter.com/, IE usage is at 4% and dropping. Is it even worth supporting and jumping through tons of hoops?

domoritz:
* Single rollup config for all modules
  * Consistency across modules
  * Extensible config file
* We can now use modern javascript since code goes through babel. We use a browserlist to ensure compatibility.
* no new major dependencies
* The browser bundles can be smaller iife bundles
* Sourcemaps for all bundles
* Fixed potential bugs:
  * vega-projection-extended was externalizing d3 array in core but we use a different version of d3 array that is not part of d3 (https://github.com/vega/vega/blob/a8e6991411c2c5c4805320f87276e9861e80d69c/packages/vega/rollup.js#L13).

Fixes #862

See #1165


jheer:
This looks pretty nice! Thanks!! However, we will need to update the PR in response to changes from other PRs. This includes a change to the excluded D3 modules for `vega-core`, thanks to the D3 v6 release.

Also, I noticed vega.min.js replacing vega.js in a bunch of places. The use of the unminified bundle was intentional to aid debugging. Is the thought here that we should use the minified code + source map instead?

domoritz:
Awesome. I'll take a look at upgrading the pull request. 

Yes, I thought we could use the Minifie's file with source maps. With this change, we would be testing that the source maps work. I'm happy to go back to the unminified version.

packages/vega-canvas/package.json
```diff
@@ -10,18 +10,17 @@
...
-    "prebuild": "rimraf build && mkdir build",
-    "build": "yarn rollup",
-    "pretest": "yarn prebuild && yarn rollup",
+    "prebuild": "rimraf build",
+    "build": "rollup -c --config-browser",
+    "pretest": "yarn build --config-test",
     "test": "tape 'test/**/*-test.js'",
     "prepublishOnly": "yarn test && yarn build",
-    "postpublish": "git push && git push --tags"
+    "postpublish": "git push"
```
domoritz, line 23: Are you still using the postpublish command @jheer?

domoritz: I'm removing it.

domoritz:
Done updating.

domoritz:
I like this kind of setup that centralizes the build logic in Rollup and started doing the same for other projects like Vega-Embed: https://github.com/vega/vega-embed/pull/544.

PR #2828, opened 2020-08-24
ES6 Modernization: use arrow functions

lgrammel:
Arrow functions are a more concise syntax that can replace the full function syntax in many cases.


domoritz:
Note that they also change behavior. Arrow functions change what 'this' refers to.

lgrammel:
@domoritz good point - the changes were done with a codemod ( https://github.com/lgrammel/js-codemod/blob/master/transforms/arrow-function.js ) that does not transform the `function` elements in cases where `this` causes problems. Would it be helpful to add more information to the pull request summary?

domoritz:
No need to change anything. I just wanted to note it for anyone coming across this pull request in the future.

PR #2854, opened 2020-09-01
ES6 Modernization: convert var to const or let

lgrammel:
**'Let' and 'const' are block-scope variable declarations that can replace 'var' declarations in many cases.**

`const` declares blocked-scoped variables that cannot be re-assigned. `let` declares block-scoped variables that can be changed. They are available since ES6 and are preferred over function-scope, modifiable `var` declarations, because they make it easier to reason about the code.

When there are several variables declared in a statement, this codemod does not change the declaration to prevent formatting breakages and linter issues.

✅ This codemod is semantically safe. It only changes variables that are used within block-scope, and only marks variables as `const` if they are not modified.

### DIFF
--- a/packages/vega-canvas/test/canvas-test.js
+++ b/packages/vega-canvas/test/canvas-test.js
@@ -3,7 +3,7 @@ var tape = require('tape'),
     canvas = require('canvas');
 
 tape('Canvas loader loads node canvas', t => {
-  var c = vega.canvas(10, 20);
+  let c = vega.canvas(10, 20);
   t.ok(c);
   t.ok(c.getContext);
   t.equal(c.width, 10);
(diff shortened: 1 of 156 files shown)


lgrammel:
This pull request has a bunch of lint failures. I've investigated them and actually eslint is wrong in those cases (at least the ones I looked at) - a `const` cannot be used because the assignment happens after the declaration. Please let me know what the best way to proceed is - is it fine to ignore the issues? would it be better to combine the assignment into the declaration first?

domoritz:
> a const cannot be used because the assignment happens after the declaration. Please let me know what the best way to proceed is [...] would it be better to combine the assignment into the declaration first?

I think eslint wants you to move the declaration to the one time the variable is defined, which I think is reasonable.

lgrammel:
Yes, that makes sense. I'll rework the code mod and will resubmit.

PR #2855, opened 2020-09-02
chore: upgrade to Vega datasets v2

domoritz:
Part of #2708

PR #2857, opened 2020-09-02
chore: use symlink instead of copying test data

domoritz:


jheer:
Addressed in revisions to #2855.

PR #2883, opened 2020-09-17
feat(vega-projection): allow "fit" to auto rotate

terencehonles:
While `fit` controls the projection `scale` and `translate` parameters it makes sense to also control the `rotate` parameter unless it has been overridden. This change uses `d3.geoCentroid` to rotate that to the center of the map.


terencehonles:
As I was updating the documentation I realized there's also a `center` for the d3 projection. I'm not sure if using rotation or center is better or if they are synonyms :thinking:

terencehonles:
Based on https://github.com/d3/d3-geo#projection_fitExtent and also testing using `center` instead of `rotate` and my map looking broken I think rotate is the one to use.

PR #2894, opened 2020-09-22
Add label transform.

jheer:
- Add label transform to Vega bundle.
- Update test cases.


domoritz:
cc @chanwutk 🎉

yuicer:
well... one question here

if the label mark property [x, y] is a scale that will be affected by another signal like dragger, datazoom, or something else. the label-transform will not run itself again.

 `if (!(mod || pulse.changed(pulse.ADD_REM) || modp('sort'))) return;`

 And may lead to some unexpected view. 

so how can we make this done?

@jheer

jheer:
@yuicer This is a closed PR. Please file a bug report (ideally with a replicating example) as a new issue. Thanks!

PR #2897, opened 2020-09-23
fix: build umd bundles for browsers instead of iife

### LINKED ISSUE #2896: invalid module error in Observable example
Requiring the module in Observable throws the error: `Te: invalid module`. Can be seen in the example notebook shared in the docs - https://observablehq.com/@vega/hello-vega-embed

domoritz:
Fixes #2896 

This fix is needed to support Vega on Observable without any hacks. See https://observablehq.com/@observablehq/how-to-require-stubborn-modules.


mbostock (commented):
Thanks for the fast fix!

PR #2908, opened 2020-09-29
fix: output browser and node bundles for both cjs and module

### LINKED ISSUE #2907: Unable to pass latest vega-loader & vega-canvas through Google Closure compiler - missing files?
Hi folks

We've discovered a problem with the most recent releases of `vega-loader` and `vega-canvas`:

https://github.com/metasoarous/oz/issues/64#issuecomment-698343242

When trying to build cljsjs output for [Oz](https://github.com/metasoarous/oz), which involves some minification/tree-shaking with Google _Closure_ (that's "s", not "j"; an implementation detail in the _ClojureScript_ ecosystem), it's not able to find the a certain set of files in the most recent releases of `vega-loader` and `vega-canvas`. In particular, the first error that pops up indicates that the `vega-loader.js` file cannot be found:

```
{:tag :shadow.build.npm/invalid-override,
 :require-from #object[java.io.File 0x4d68156d ".../node_modules/vega-dataflow/build/vega-dataflow.js"],
 :require "vega-loader",
 :file #object[java.io.File 0x41ebcb92 ".../node_modules/vega-loader/build/vega-loader.node.js"],
 :override "./build/vega-loader.js",
 :override-file #object[java.io.File 0x5d156b61 ".../node_modules/vega-loader/build/vega-loader.js"]}
```

I realize this doesn't say anything about `vega-canvas`, but I'm assuming once the above was fixed the `vega-canvas` bug popped up. Reverting both packages back one bugfix version has unblocked me for now, but I won't be able to cut releases with newer versions until the underlying issue is resolved.

I'm a bit clueless as to what might have changed in the build process to be causing this, but hopefully this is enough information for you all to track down what might need fixing. If not, please let me know if there's any additional context needed which I might be able to help provide.

Thanks for your time!

domoritz:
Fixes #2907. This pull request adds separate module and umd bundles for browser and node. 

See https://github.com/vega/vega/issues/2907 for a discussion.


jheer:
In what ways have you tested this? The original setup (with multiple entries under the `browser` property) was created to support browser bundling both with CJS and module imports. For example, consider the following:

```json
"browser": {	
    "./build/vega-canvas.node.js": "./build/vega-canvas.js",	
    "./index.js": "./index.browser.js"	
  }
```

The first line redirects CJS imports to use the browser-safe version. If we no longer generate `vega-canvas.js` that might be replaced with `vega-canvas.min.js`? The second line redirects module imports to use the browser-safe version.

Do your changes cover both of these use cases? They were added as different users needed both.

domoritz:
Ahh, I don't support module builds for browsers. I'll update the pull request to make it work.

jheer:
For a simpler alternative see #2910, which simply updates the route for a browser-based module build to point to the appropriate index file. I plan to merge that PR, but we can always refine further via PRs like this one as we see fit.

domoritz:
This pull request now adds all the builds one may want. I looked at the compiled bundles and they look reasonable to me but a second pair of eyes won't hurt either.

domoritz:
@metasoarous and @thheller if you have thoughts on this pull request, let me know.

domoritz:
@jheer can you see whether you want to merge this change?

domoritz:
ping @jheer

domoritz:
@jheer I think this fix should still be merged as `index.browser.js` does not go through babel.

PR #2910, opened 2020-10-01
Fix browser build routes.

### LINKED ISSUE #2907: Unable to pass latest vega-loader & vega-canvas through Google Closure compiler - missing files?

jheer:
**vega-canvas**
- Fix browser index route for ES modules

**vega-loader**
- Fix browser index route for ES modules

Intended to help resolve #2907

### DIFF
--- a/packages/vega-loader/package.json
+++ b/packages/vega-loader/package.json
@@ -39,6 +39,6 @@
   },
   "browser": {
     "./build/vega-loader.node.js": "./build/vega-loader.min.js",
-    "./index.js": "./index.browser.js"
+    "./build/vega-loader.module.js": "./index.browser.js"
   }
 }
--- a/packages/vega-canvas/package.json
+++ b/packages/vega-canvas/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vega-canvas",
-  "version": "1.2.5",
+  "version": "1.2.6",
   "description": "Canvas and Image utilities.",
   "keywords": [
     "vega",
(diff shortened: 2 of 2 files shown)


packages/vega-loader/package.json
```diff
@@ -39,6 +39,6 @@
   },
   "browser": {
     "./build/vega-loader.node.js": "./build/vega-loader.min.js",
-    "./index.js": "./index.browser.js"
+    "./build/vega-loader.module.js": "./index.browser.js"
```
domoritz, line 42: Hmm, I don't think this is the full fix yet since `index.browser.js` might use features that need to be transpiled. It's a good hotfix, though.

PR #2916, opened 2020-10-08
feat: add `pluck` expr function to retrieve values from an array.

arvind:


jheer:
Looks good (thanks!), but I have a few questions:

- Are the multiple input types (object or array) necessary? Users can already access a single object's properties in the expression language, so I would have an expected an array-specific method. Is there a particular use case that motivates the current approach?

- I'm not sure if the name `field` is the most clear. We use that term in a related but nevertheless slightly different way throughout Vega. To my mind it refers to a property of a single object, not a "verb" for extraction from a set of objects. Following other popular utility libraries, maybe something like `pluck` would be more familiar and array-specific?

arvind:
> Are the multiple input types (object or array) necessary? Users can already access a single object's properties in the expression language, so I would have an expected an array-specific method. Is there a particular use case that motivates the current approach?

Nope, no reason. I had the ternary in there because it seemed like a convenient/"easy win" to expose dot-separated access paths. But I'm happy to drop it and always assume an array. 

> I'm not sure if the name field is the most clear. We use that term in a related but nevertheless slightly different way throughout Vega. To my mind it refers to a property of a single object, not a "verb" for extraction from a set of objects. Following other popular utility libraries, maybe something like pluck would be more familiar and array-specific?

Agreed, I wasn't happy with `field` either. `pluck` sounds better. Happy to make the switch.

arvind:
Renamed to `pluck` but kept the array/object input types around because it may simplify Vega-Lite's handling of nested field references (which are currently copied to a new top-level property name with a formula transform).

jheer:
Thanks @arvind! I will keep this PR open for now, then merge it into a v5.18.0 branch when that is ready, to avoid pre-releasing the documentation.

domoritz:
Should this change go into v5.19 (v5.18.0 is out already)?

PR #2918, opened 2020-10-10
ci: remove caching

domoritz:


domoritz:
Caching can be really slow. Here it took 20 minutes. 😱 

Screen Shot 2020-10-09 at 17 02 18

Issue #2920, opened 2020-10-12
Memory Leak when replacing data - Subflows/Subcontexts with references to data not GCed


krisdages:
Our application has a faceted chart for which we reload and replace the data on a timer using the View api.
The symptom of the issue is that the RAM usage of the chart grows indefinitely every time we update the data and re-render,
  eventually crashing the browser tab.
We are using the default behavior, and `clean()` does get executed in a `runAfter` callback.

The memory profiler reveals that the previous data instances are kept around because references are held in Subflows / Subcontexts that are not garbage collected.  These appear to be retained in a few places:
* The `pulse` objects and `source` properties of detached `Operator`s
* The bound `this` context of `EventStream._filter` and stream's `targets`.  `EventStream.detach` is a noop so these aren't cleared.
  (are these streams present because we have signal events on the group marks?)

Using Vega v5.17.0
Seems related to #688

I created a patch to clear these references on clean/detach that:
* Removes `pulse` and `source` on `Operator` when detaching
* Changed `EventStream.detach` from a noop to a fn that resets `_filter` to `truthy` and clears `targets`
It works great for our project, but I don't know the framework intimately and I think it may be a little too naive of an approach for other use cases, especially the change to EventStream.
I'll create a PR, but I expect some improvements might be needed.

Thanks for taking a look!

[...]

PR #2921, opened 2020-10-12
Re #2920: Fix memory leaks when replacing data

krisdages:
https://github.com/vega/vega/issues/2920

* When detaching Operators, clear the references to `source` and `pulse`
* When detaching EventStreams, reset the `_filter` function and clear `targets`
* When cleaning Facet, remove subflows with count 0 from the active _targets array.

These may be a little too much of a "blunt object" approach for the general use case, just wanted to share what fixed the issue for me.


packages/vega-dataflow/src/EventStream.js
```diff
@@ -122,7 +122,10 @@ EventStream.prototype = {
   },
 
   detach() {
-    // no-op for handling detach requests
     // ensures compatibility with operators (#2753)
+    //Remove any references to other streams
```
domoritz, line 126:
```suggestion
    // Remove any references to other streams
```

packages/vega-runtime/src/context.js
```diff
@@ -65,7 +65,10 @@ Context.prototype = Subcontext.prototype = {
     // wipe out targets first for better efficiency
     const keys = Object.keys(ctx.nodes);
     for (const key of keys) ctx.nodes[key]._targets = null;
-    for (const key of keys) ctx.nodes[key].detach();
+    for (const key of keys) {
+      ctx.nodes[key].detach();
+      delete ctx.nodes[key];
```
jheer, line 70: Is the `delete` operation really necessary? (We null out `ctx.nodes` right after this.) Or is there a remaining pointer to that object elsewhere? If so, we might want address that directly...

If it does make an improvement to delete, we might instead assign `undefined`, as this is faster than actually removing the property.

krisdages: Sorry, I can't remember... I think that there may have been references to it held by forked subcontexts, but I could be wrong.

krisdages: I tried removing this change and rechecking the issue in my dev environment and it doesn't look to be necessary.

domoritz:
@krisdages do you have time to wrap up this pull request so we can merge it?

packages/vega-transforms/src/Facet.js
```diff
@@ -60,6 +60,11 @@ inherits(Facet, Transform, {
         delete flows[key];
       }
     }
+    //Remove targets that are no longer active from the active targets array.
```
domoritz, line 63:
```suggestion
    // Remove targets that are no longer active from the active targets array.
```

krisdages:
Sorry, things have been busy, hadn't checked github in a while...

I updated the branch with the requested change and merged the latest master in, wasn't sure if you'd prefer a rebase instead...
I got an error trying to run `yarn test` locally with both master and my branch after the merge:

$ TZ=America/Los_Angeles tape 'test/**/*-test.js'
/.../vega/node_modules/ajv/dist/core.js:129
                throw new Error(`no schema with key or ref "${schemaKeyRef}"`);
                ^

Error: no schema with key or ref "http://json-schema.org/draft-06/schema#"
    at Ajv.validate (/.../vega/node_modules/ajv/dist/core.js:129:23)
    at Ajv.validateSchema (/.../vega/node_modules/ajv/dist/core.js:240:28)
    at Ajv._addSchema (/.../vega/node_modules/ajv/dist/core.js:434:18)
    at Ajv.compile (/.../vega/node_modules/ajv/dist/core.js:140:26)
    at Object.<anonymous> (/.../vega/packages/vega/test/schema-test.js:17:28)
 
Any idea what might be causing that? The CI server doesn't seem to have had a problem with it.

jheer:
Thanks @krisdages! Fortunately I'm not seeing that problem locally.

PR #2937, opened 2020-10-20
update link to protovis

mulka:
... because it isn't the right site any more


domoritz:
Thank you. Let's replace the link with https://mbostock.github.io/protovis/.

PR #2941, opened 2020-10-22
Remove ableist language

domoritz:
Removing ableist language such as "dumb" and "sanity check". See https://developers.google.com/style/inclusive-documentation.


jheer:
Thanks!

PR #2944, opened 2020-10-24
Mention why there will be fewer bins than specified in `maxbins`.

joelostblom:
I find this a bit unclear currently, and also often get asked what determines the actual number of bins. This PR is my understanding of https://observablehq.com/@d3/d3-bin, particularly this section:

> Unexpected? The result is not guaranteed to yield the exact number of buckets we asked for. That number is only an indication, passed to the threshold generator. As you change that value with the slider, you might notice the reason why this happens: the domain gets sliced at “nicely-rounded” values — using d3.ticks — and these rounded values do not fall exactly on the bin’s domain.


jheer:
Thanks!

PR #2988, opened 2020-11-30
Update transforms.md

oaustegard:
Add a link back to the Extensibility API documentation to highlight that custom transformations are also possible


jheer:
Thanks!

PR #2996, opened 2020-12-09
Option to run force simulations on a Web Worker

wmurphyrd:
@jheer we talked about this on Slack when I first began the project and there was potential interest in merging to main. We've been using this in production for several months, but I've just now had a chance to upgrade and confirm it works with the current Vega version. 

This adds a worker option to the force transform which will cause simulation calculations to be performed in a Web Worker thread. This improves the user experience when running sims on large datasets because it avoids threadlocking the browser.

I've added a `simulationWorker` that mimics the existing api of `simulation` (to minimize changes to existing code), but coordinates messaging with the worker thread to offload the work there. 

Due to browser security policies surrounding Web Workers, [they must be instantiated with a URL to a script for the worker source that obeys a strict same-origin policy](https://developer.mozilla.org/en-US/docs/Web/API/Worker/Worker), the use cases for the worker mode are limited compared to Vega in general. The built worker thread file, `vega-force.worker.js` cannot be sourced from CDN, so users will have to host and serve the file. However, the transform continues to work on the main thread without changes if the worker mode is not set in the transform settings.

Here's a live demo: https://glitch.com/edit/#!/vega-force-worker?path=index.html%3A1%3A1


wmurphyrd:
Re the worker script URL requirement, [there are workarounds](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers#Embedded_workers), but they mostly involve unsafe inline code evaluation and may not have universal browser support. If you'd prefer this route, such that the `worker` transform option could be a simple Boolean and it would work from CDN, I'd be happy to implement, but I wanted to check first since it's non-standard.

PR #3008, opened 2020-12-15
add param to vlSelectionResolve to distinguish VL5 "point" selections

arvind:
This was the cleanest change I could think of introducing that wouldn't trigger a major version bump for Vega. But I'm open to other ideas.


domoritz:
Needed for https://github.com/vega/vega-lite/pull/7116

PR #3019, opened 2020-12-21
Fix XSS vulnerability in expression parser.

### LINKED ISSUE #3018: XSS in transform filter
This was found during the hxp ctf.
### Credit
@cgvwzq and his [writeup](https://gist.github.com/cgvwzq/0392d64fe1aed82f65c1648b71df1a37)

- Describe how to reproduce the bug / the goal of the feature request:
  Paste the below JSON in the [Vega Editor](https://vega.github.io/editor/#/custom/vega-lite).
  [Working demo.](https://vega.github.io/editor/#/url/vega-lite/N4IgJghgLhIFygG4QDYFcCmBneBtYAvgLoEA0IUAThAHZYBmA9pQLZ6j0CWKUGl8IABQAGAPSiAjKICUAWgDkAHWU1SqPlEETpO6ePkhi5FhEoBrAQCNThoA)
  You will see a '1' alert dialog. 
  To my understanding you should not be able to run arbitrary JS using vega-lite json, should you?
- Provide an example spec in JSON, wrapped by triple backticks like this:

```json
{
  "data": {
    "values": [{}]
  },
  "transform": [
    {"filter": "(0//1/)-'\\\n,alert(1))))//'"}
  ],
  "mark": "bar"
}
```

jheer:
**vega-expression**

- Fix XSS vulnerability in expression parser and code generator.

Fix #3018.

### DIFF
--- a/packages/vega-expression/src/codegen.js
+++ b/packages/vega-expression/src/codegen.js
@@ -83,7 +83,7 @@ export default function(opt) {
         '[' + n.elements.map(visit).join(',') + ']',
 
     BinaryExpression: n =>
-        '(' + visit(n.left) + n.operator + visit(n.right) + ')',
+        '(' + visit(n.left) + ' ' + n.operator + ' ' + visit(n.right) + ')',
 
     UnaryExpression: n =>
         '(' + n.operator + visit(n.argument) + ')',
--- a/packages/vega-expression/src/parser.js
+++ b/packages/vega-expression/src/parser.js
@@ -436,6 +436,10 @@ function scanPunctuator() {
     };
   }
 
+  if (ch2 === '//') {
+    throwError({}, MessageUnexpectedToken, ILLEGAL);
+  }
+
   // 1-character punctuators: < > = ! + - * % & | ^ /
 
   if ('<>=!+-*%&|^/'.indexOf(ch1) >= 0) {
--- a/packages/vega-expression/test/parser-test.js
+++ b/packages/vega-expression/test/parser-test.js
@@ -60,19 +60,14 @@ tape('Parser should allow literal regular expressions', t => {
     regex: { pattern: 'a', flags: ''}
   });
   // Empty regex
-  t.deepEqual(parse('//')(), {
-    type: 'Literal',
-    value: {},
-    raw: '/(?:)/',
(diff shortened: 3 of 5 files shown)


domoritz (dismissed):

cgvwzq:
Hi, not sure if a patch at that level will be enough:

```{
  "data": [
    {
      "name": "data",
      "values": [{}],
      "transform": [
        {
          "type": "filter",
          "test": "(0/ /1/)-'\\\n,alert(1))))//'"
        }
      ]
    }
  ]
}```

In this case it builds a wrong AST and the code generation removes the whitespace, inserting a comment.

**Comment by jheer** [comment]
Thanks @cgvwzq, I've updated the PR to safeguard the code generation as well.

**Review by domoritz — APPROVED** [review]

**Comment by cgvwzq** [comment]
Glad to help :)

However, I'm not sure this is the right way to fix the overall threat, as this patch only prevents the specific payload I used. My experience with JS sandboxing based on parsing/rewriting is that it is extremely hard to get right, and if someone looks into it more security issues will certainly emerge. Just looking at some other features, signals and events seem another source of similar problems.

I do not want to turn this into a cat-and-mouse game (as fun as it could be :P), so I would rather suggest to try to rethink the overall eval business. If you need it, I would at least try to make sure that all generated code runs in a sandboxed iframe in isolation (you can rely on postMessage to communicate all data (and events?).

**Comment by nename0** [comment]
Is there a specify reason that you have to use `eval`?

If you already have an AST couldn't you also just evaluate that AST?
Something along those lines:
```js
{
  BinaryExpression: n => {
    switch(n.operator) {
      case '*': return visit(n.left) * visit(n.right);
      case '+': return visit(n.left) + visit(n.right);
      case '==': return visit(n.left) == visit(n.right);
      ...
     }
  },

  UnaryExpression: n =>{
    switch(n.operator) {
      case '!': return !visit(n.argument);
      case '-': return -visit(n.argument);
      ...
    } 
  },
  ...
}
```
That would at least prevent any mismatch between interpretation by your parser and the evaluation.

domoritz:
@jheer actually added an interpreter mode to Vega but it's slightly slower so it's not the default: https://github.com/vega/vega/tree/master/packages/vega-interpreter.

PR #3026, opened 2020-12-21
fix: use correct JSON schema keyword

### LINKED ISSUE #3025: Schema uses undefined keyword numItems
vega/packages/vega-schema/src/scale.js uses numItems, which is not a valid JSON schema keyword.

domoritz:
fixes #3025


jheer:
Thanks!

PR #3028, opened 2020-12-22
ci: add codeql analysis

domoritz:


domoritz:
Looks like it's the same as LGTM.

PR #3036, opened 2021-01-03
docs: add link to vega-lite-api in the dropdown

kanitw:
fix https://github.com/vega/vega-lite/issues/7020


domoritz (commented):
Don't we also need to update the CSS as in https://github.com/vega/vega-lite/pull/7143/files#diff-45f82b68551d044bd2bc72c7812d0c331efb9c23f2799fb6fc298bfa32737c92?

kanitw:
No, it looks ok here because the text Vega is way shorter.

domoritz (approved):

PR #3043, opened 2021-01-13
Update index.md

afdta:
I don't have a ton of experience with bundlers (I am using rollup) and I would have benefited from this quick explanation.

### DIFF
--- a/docs/usage/index.md
+++ b/docs/usage/index.md
@@ -53,7 +53,7 @@ To use Vega on a web page you first need to load the Vega JavaScript files. The 
 </head>
 ```
 
-**Using Vega with a bundler.** If using Vega with a package manager (e.g. [npm](https://www.npmjs.com/package/vega)) and a bundler like [rollup.js](https://rollupjs.org/guide/en/#with-npm-packages), you can import everything into a namespace as follows:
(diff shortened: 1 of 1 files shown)


domoritz:
Thank you for the pull request. However, I think the readme isn't the right place for this note. Instead, it should be at https://vega.github.io/vega/usage/.

afdta:
> Thank you for the pull request. However, I think the readme isn't the right place for this note. Instead, it should be at https://vega.github.io/vega/usage/.

I guess I'm confused, I thought this was the markdown for the usage page?

domoritz:
Ignore me. I should go to bed. You are right.

docs/usage/index.md
```diff
@@ -53,6 +53,12 @@ To use Vega on a web page you first need to load the Vega JavaScript files. The
 </head>
 ```
 
+**Using Vega with a bundler.** If using Vega with a package manager (e.g. [npm](https://www.npmjs.com/package/vega)) and a bundler like [rollup.js](https://rollupjs.org/guide/en/#with-npm-packages), you can import everything into a namespace as follows:
```
domoritz, line 56:
```suggestion
**Using Vega with a bundler.** If you use Vega with a bundler like [rollup.js](https://rollupjs.org/guide/en/#with-npm-packages), you can import Vega [as a module](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import).
```

domoritz (approved):

jheer:
Thanks @afdta!

PR #3049, opened 2021-01-20
fix xmlns namespace

mbostock:
For context, please see:

https://talk.observablehq.com/t/svg-export-from-a-vega-lite-api-chart-error-on-opening-the-file/4398/7

I didn’t touch SVGStringRenderer.js or image.js, which also reference this metadata without regard to namespaces; it might be good to fix those too, but if they’re just strictly for serialization, it might not matter.

Fixes vega/vega-lite-api#213.

### DIFF
--- a/packages/vega-scenegraph/src/SVGRenderer.js
+++ b/packages/vega-scenegraph/src/SVGRenderer.js
@@ -44,7 +44,9 @@ inherits(SVGRenderer, Renderer, {
 
     if (el) {
       this._svg = domChild(el, 0, 'svg', ns);
-      setAttributes(this._svg, metadata);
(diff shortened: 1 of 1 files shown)


domoritz:
This should fix https://github.com/vega/vega-lite-api/issues/213.

jheer:
Thanks @mbostock! Updated to use the correct xmlns namespace and merged.

jheer:
As for the other uses of the namespace metadata:

- SVGStringRenderer is for text serialization only, so it does not matter.
- The image mark passes namespace info to the renderer via the emit call, which routes to appropriate setters as needed.

mbostock:
Thanks for fixing my fix and for merging! 🙏

PR #3063, opened 2021-02-01
feat!: switch from node canvas to skia canvas

domoritz:
Try https://github.com/samizdatco/skia-canvas.

There are some issues with images and so the tests fail.


domoritz:
https://github.com/samizdatco/skia-canvas doesn't seem well maintained anymore. Maybe we try again later.

PR #3095, opened 2021-02-24
feat: export vega-expression and vega-event-selector

### LINKED ISSUE #2917: Export vega-expression and vega-event-selector
Both are used by Vega-Lite and exporting them would allow Vega-Lite to ship with a smaller bundle.

domoritz:
fixes #2917

### DIFF
--- a/packages/vega-typings/types/runtime/index.d.ts
+++ b/packages/vega-typings/types/runtime/index.d.ts
@@ -257,3 +257,12 @@ export * from 'vega-util';
 export * from './dataflow';
 export * from './renderer';
 export * from './scene';
+
+export {
+  codegen as expressionCodegen,
+  parse as expressionParse
+} from 'vega-expression';
+
+export {
+  selector as parseSelector
+} from 'vega-event-selector';
--- a/packages/vega-typings/types/runtime/index.d.ts
+++ b/packages/vega-typings/types/runtime/index.d.ts
@@ -260,4 +260,4 @@ export * from './scene';
 
 export { codegen as codegenExpression, parse as parseExpression } from 'vega-expression';
 
-export { selector } from 'vega-event-selector';
+export { selector as parseSelector } from 'vega-event-selector';
--- a/packages/vega-typings/types/runtime/index.d.ts
+++ b/packages/vega-typings/types/runtime/index.d.ts
@@ -260,4 +260,4 @@ export * from './scene';
 
 export { codegen as codegenExpression, parse as parseExpression } from 'vega-expression';
 
-export { selector as parseSelector } from 'vega-event-selector';
+export { selector } from 'vega-event-selector';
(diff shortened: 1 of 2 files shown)


domoritz:
Needed for https://github.com/vega/vega-lite/pull/7263

jheer (approved):
LGTM

domoritz:
One thing I wonder is whether it is better to export with a different name or export the exact same name a package exports. The advantage of the latter is that one can do a rename in rollup as we do in https://github.com/vega/vega-lite/blob/e19d9ac23396dc9b3579279f4f66a37173ffa098/rollup.config.js#L34. 

Do use the renamed exports (as in this pull request), I had to change Vega-Lite do depend on Vega directly (as opposed to specific packages). I don't think it's a big issue for Vega-Lite since Vega is already a peer dependency but I wanted to see what you think before we go down this route.

jheer:
Sometimes it might be necessary to rename an export -- obviously if there is a name collision, but also if the name becomes to ambiguous in the enlarged context. So I'm not averse to that. That said, if the name works well enough on its own, my bias would be to prefer keeping names unchanged. In this case, the expression exports need to be renamed (e.g., we can't have two _parse_ methods), but I'm not sure the selector export needs to be renamed.

packages/vega-typings/types/runtime/index.d.ts
```diff
@@ -257,3 +257,7 @@ export * from 'vega-util';
 export * from './dataflow';
 export * from './renderer';
 export * from './scene';
+
+export { codegen as expressionCodegen, parse as expressionParse } from 'vega-expression';
+
+export { selector as parseSelector } from 'vega-event-selector';
```
arvind, line 263: Would be good to keep the naming scheme parallel. So parseSelector/Expression or selector/expressionParse (I prefer the former).

jheer: I similarly prefer `parseExpression` over `expressionParse`.

domoritz: Btw, we use `runtimeContext` above. Do you also prefer `codegenExpression` (it sounds less natural than `expressionCodegen` as we go from expression AST to code)?

jheer: I'm fine with `codegenExpression` for parity with `parseExpression`

domoritz: Done

packages/vega-typings/types/runtime/index.d.ts
```diff
@@ -260,4 +260,4 @@ export * from './scene';
 
 export { codegen as codegenExpression, parse as parseExpression } from 'vega-expression';
 
-export { selector as parseSelector } from 'vega-event-selector';
```
arvind: I think my prev comment was unclear — I had wanted us to be consistent (`parseSelector` and `parseExpression` or `selectorParse` and `expressionParse`). I think we should keep the rename here as `parseSelector`.

domoritz: Done

arvind (approved):

PR #3096, opened 2021-02-25
fix: handle error in label-transform when using empty data

chanwutk:
Fix #3064

PR #3097, opened 2021-02-25
chore: use @babel/eslint-parser

domoritz:
Use babel eslint parser since it always supports the latest js features. 

Alternatively, we can use https://github.com/vega/vega/pull/3098 if we only want to support e.g. nullish coalescing.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,4 +1,5 @@
 module.exports = {
+  parser: '@babel/eslint-parser',
   extends: [
     'eslint:recommended'
   ],
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,7 +9,6 @@ module.exports = {
     node: true
   },
   parserOptions: {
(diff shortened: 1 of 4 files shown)


jheer:
This looks fine, but can you include a short note about the reason for switching to this parser?

jheer:
If this is related to #3096 and support for `??`, I'm a bit skeptical (more regarding #3096 than this particular PR). I believe Node 12 does not support the `??` operator but is still used in production in various companies, etc. We will need to test and make sure the build + babel outputs are correct and working for older but still relatively modern runtimes.

domoritz:
I added a note to the pull request. Yes, this is to support any JavaScript features babel supports. Since all packages run through Babel, I think we should be able to use `??` or any other modern js features. We should still confirm that it's working correctly, though.

domoritz:
I believe that https://github.com/vega/vega/pull/2908 is necessary to ensure everything we publish runs through babel.

PR #3098, opened 2021-02-25
chore: use ecmaVersion 2020 in eslint

domoritz:
Needed to support nullish coalescing: https://eslint.org/blog/2020/06/eslint-v7.2.0-released.


domoritz:
I think https://github.com/vega/vega/pull/3097 is the right approach so I'm closing this.

Issue #3075, opened 2021-02-10
Vega triggers override mistake when overriding O.p.hasOwnProperty by assignment


erights:
At docs/vega-core.js line 32463 and at doc/vega.js line 45238 appears the line

```js
  exports.hasOwnProperty = has;
```

This `exports` object inherits from `Object.prototype` which has a method named `hasOwnProperty`. Unfortunately, because of the [EcmaScript Override Mistake](https://web.archive.org/web/20141230041441/http://wiki.ecmascript.org/doku.php?id=strawman:fixing_override_mistake) (also [here](https://github.com/tc39/ecma262/pull/1307) and [here](https://github.com/tc39/ecma262/pull/1320)), in running in an environment (like SES, XS, or TC53 JavaScript for Embedded Systems) where `Object.prototype` is frozen, the Override Mistake will cause that assignment to fail.

From the line numbers above, it is clear these two files are generated files. However I do not understand what the actual source lines are that are translated into these generated lines, so I'm not sure what the best fix is. Two immediate possibilities: 
   * Have `exports` not inherit from `Object.prototype`. Depending on the role of `exports` in the translation, I would not be surprised if that is not feasible.
   * Use `defineProperty` rather than assignment, such as
      ```js
       Object.defineProperty(exports, 'hasOwnProperty', {
           value: has, 
           enumerable: true, 
           writable: true, 
           configurable: true 
       });
       ```

See https://github.com/Agoric/agoric-sdk/issues/2373 for the consequences of this bug in the Agoric use of SES.

erights:
See https://github.com/Agoric/agoric-sdk/pull/2575

domoritz:
I suspect the problem is in https://github.com/vega/vega/blob/master/packages/vega-util/src/hasOwnProperty.js. Do you have an idea how we can resolve the issue you are seeing?

erights:
No, that's not the problem. There is nothing about that file that would trigger the override mistake. 

My first post above https://github.com/vega/vega/issues/3075#issue-805227532 explains the problem well, but complains about the wrong file. The problem is in vega-util/build/vega-util.js . Aside from that, please ask about anything unclear in the explanation in that first post.

Vega actually has three problems along these lines, for `constructor`, 'hasOwnProperty` and `toString`, all in the vega-util.js file. A patch which fixes all these is at
https://github.com/Agoric/agoric-sdk/blob/master/patches/vega-util%2B1.16.0.patch

erights:
[...]

PR #3109, opened 2021-03-05
Do not override primordial property by assignment

erights:
By "primordial object" I mean all standard EcmaScript objects that exist before code starts running, like especially `Object.prototype`. [SES](https://github.com/tc39/proposal-ses) and several other systems freeze all JS primordials on initialization. As explained at https://github.com/vega/vega/issues/3075 , because of the override mistake, on an object that inherits from a primordial object, an attempt to override one of these properties, such as `hasOwnProperty`, by assignment will fail to do so. (By "override" we mean to create a new own property of the same name on the inheriting object.)

This PR replaces the occurrences we are aware of vega making this mistake directly in its source code. 

However, as explained at https://github.com/vega/vega/issues/3075 the cases that matter are automatically generated somehow by vega itself. The change needs to produce code like that show by the patches additions of https://github.com/Agoric/agoric-sdk/blob/master/patches/vega-util%2B1.16.0.patch which enable vega to work under SES as used by the agoric-sdk monorepo.

### DIFF
--- a/docs/docs/signals.md
+++ b/docs/docs/signals.md
@@ -131,7 +131,7 @@ This signal definition invokes a custom encoding set upon `mousedown` and `mouse
 
 ## <a name="bind"></a>Input Element Binding
 
-The _bind_ property binds a signal to an input element defined outside of the visualization. Vega will generate new HTML form elements and set up a two-way binding: changes to the input element will update the signal, and vice versa. Vega includes dedicate support for `checkbox` (single boolean value), `radio` (group of radio buttons), `select` (drop-down menu), and `range` (slider) input types.
+The _bind_ property binds a signal to an input element defined outside of the visualization. Vega will generate new HTML form elements and set up a two-way binding: changes to the input element will update the signal, and vice versa. Vega includes dedicated support for `checkbox` (single boolean value), `radio` (group of radio buttons), `select` (drop-down menu), and `range` (slider) input types.
 
 | Property            | Type                           | Description  |
 | :------------------ | :----------------------------: | :------------|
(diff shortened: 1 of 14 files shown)


domoritz:
Thank you for the pull request. Can you remove the compiled files in docs as we compile them during releases?

erights:
Hi @domoritz I did, so we can proceed with this PR. However, this omits the change that matters most, to the line

```js
exports.hasOwnProperty = has;
```
This line does not appear anywhere in your sources. I see now that vega does its own code generation, and so must be generating this line. However, I could not figure out where this happens or how to fix it. So I have removed from the PR description that is fixes the
https://github.com/vega/vega/issues/3075
issue.

domoritz:
I assigned @jheer as he wrote most of the code.

jheer:
It looks like the remaining issues stem from vega-util/index.js where we name the exports such as hasOwnProperty. One route could be to rename those exports, though that involves a breaking change and thus a major version bump. Alternatively, perhaps there is a way to signal to package bundlers (such as rollup, which Vega uses) to use defineProperty for those exports?

erights:
Is the `export.<stuff> = <stuff>;` coming from rollup? That would explain why I couldn't find it in your codegen ;)

jheer:
Yes, that is the output of rollup.

erights:
Thanks. In that case, this PR as is addresses the entire vega specific issue. The remaining bug is in rollup. I'll pursue it over there.

Is this PR by itself good? Thanks.

erights:
This PR would, therefore, close out #3075 . I will file a separate bug with rollup.

erights:
Thanks!

PR #3117, opened 2021-03-14
Add signal binding to external input element.

jheer:
When binding signals to input elements, Vega typically generates its own HTML input elements within a given container. This PR add supports for binding _directly_ to an existing input element by specifying the `element` property of a binding but _not_ the `input` type. This `element` entry must expose a `value` property and support the [`EventTarget` interface](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget) (or be a CSS selector for such an element).

This approach should support binding to an Observable cell containing an input element, for example using signal syntax such as: `bind: { element: viewof inputCell }`.

**vega**
- Add web test for external element binding.

**vega-schema**
- Add signal binding to external input element.

**vega-typings**
- Add signal binding to external input element.

**vega-view**
- Add signal binding to external input element.

### DIFF
--- a/packages/vega-schema/src/bind.js
+++ b/packages/vega-schema/src/bind.js
@@ -39,7 +39,12 @@ const bind = oneOf(
     element: elementRef,
     debounce: numberType,
     name: stringType
-  }, true)
+  }, true),
+  object({
+    _element_: elementRef,
+    event: stringType,
+    debounce: numberType
+  })
 );
 
 export default {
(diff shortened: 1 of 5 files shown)


domoritz:
Can you update https://github.com/vega/vega/blob/master/packages/vega-typings/types/spec/bind.d.ts as well?

jheer:
@domoritz I updated the typings. Documentation updates still TODO.

PR #3124, opened 2021-03-16
v5.20.0

jheer:
Changes from [v5.19.1](https://github.com/vega/vega/releases/tag/v5.19.1):

**monorepo**
- Update CI configuration. (thanks @domoritz!)
- Update build configuration. (thanks @domoritz!)
- Update dependencies.

**vega**
- Add web test for external element binding.

**vega-scenegraph**
- Add early exit for image rendering with zero width or height.

**vega-schema**
- Add signal binding to external input element.

**vega-typings**
- Add signal binding to external input element.

**vega-util**
- Update inherit utility to use defineProperty, avoids upsetting configurations that ban assignment overrides of Object prototype properties. (#3109, thanks @erights!)

**vega-view**
- Add signal binding to external input element.

PR #3127, opened 2021-03-16
feat: add typing to label-transform properties

chanwutk:


domoritz (approved):
Looks good but tests are failing

chanwutk:
It seems like the failing tests are the same ones from #3124. Should I try to fix it here, or should I wait for fixes in #3124 and rebase?

domoritz:
Ahh, right. No, it should be fixed in https://github.com/vega/vega/pull/3124

domoritz:
These changes will make it easier for us to import the types in https://github.com/vega/vega-lite/pull/7222.

PR #3154, opened 2021-04-06
feat: consistently name the exports from Vega and vega-expression and vega-event-selector

domoritz:
This change allow consumers to externalize vega-expression and vega-event-selector when they already have Vega

### DIFF
--- a/packages/vega-event-selector/index.js
+++ b/packages/vega-event-selector/index.js
@@ -1,1 +1,4 @@
-export {default as selector} from './src/event-selector';
+export {
+    default as selector,
+    default as parseSelector
+} from './src/event-selector';
--- a/packages/vega-expression/index.d.ts
+++ b/packages/vega-expression/index.d.ts
@@ -1,5 +1,9 @@
+import { Node } from "estree";
+
 /** Parse a JavaScript *expression* string and return the resulting abstract syntax tree in the ESTree format */
-export function parse(expression: string): any;
+export function parse(expression: string):  Node.Expression | Node.SequenceExpression;
+
+export const parseExpression = parse;
 
 interface CodegenOptions {
   /** A hash of allowed top-level constant values */
@@ -38,6 +42,8 @@ export function codegen(
   /** A hash of all properties referenced outside a provided allowed list */
   globals: string[];
 };
+
+export const codegenExpression = codegen;
 
 /** An object defining default constant values for the Vega expression language */
 export const constants: { [cn: string]: string };
--- a/packages/vega-expression/index.js
+++ b/packages/vega-expression/index.js
@@ -14,7 +14,13 @@ export {
   default as ASTNode
 } from './src/ast';
 
-export { default as parse} from './src/parser';
-export { default as codegen } from './src/codegen';
+export {
+  default as parse,
+  default as parseExpression
+} from './src/parser';
+export {
+  default as codegen,
+  default as codegenExpression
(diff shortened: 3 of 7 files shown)


packages/vega-expression/index.d.ts
```diff
@@ -1,7 +1,11 @@
+import { SequenceExpression, Expression } from "estree";
+
 /** Parse a JavaScript *expression* string and return the resulting abstract syntax tree in the ESTree format */
-export function parse(expression: string): any;
+export function parse(expression: string): Expression | SequenceExpression;
+
+export {parse as parseExpression};
 
-interface CodegenOptions {
+export interface CodegenOptions {
```
domoritz, line 8: Not sure why this wasn't exported already.

packages/vega-typings/types/runtime/index.d.ts
```diff
@@ -1,19 +1,17 @@
+import { LoggerInterface } from 'vega-util';
 import {
   Color,
   Config,
   DataType,
   EncodeEntryName,
   Format,
```
domoritz, line 7: I'm removing unused types.

packages/vega-expression/index.d.ts
```diff
@@ -1,7 +1,11 @@
+import { SequenceExpression, Expression } from "estree";
+
 /** Parse a JavaScript *expression* string and return the resulting abstract syntax tree in the ESTree format */
-export function parse(expression: string): any;
+export function parse(expression: string): Expression | SequenceExpression;
```
domoritz, line 4: I made the types more specific so that one can typecheck ASTs.

packages/vega-event-selector/index.js
```diff
@@ -1 +1,4 @@
-export {default as selector} from './src/event-selector';
+export {
+    default as selector,
+    default as parseSelector
```
kanitw, line 3: Add comment documenting why we need to export default twice?

packages/vega-expression/index.js
```diff
@@ -14,7 +14,13 @@ export {
   default as ASTNode
 } from './src/ast';
 
-export { default as parse} from './src/parser';
-export { default as codegen } from './src/codegen';
+export {
+  default as parse,
+  default as parseExpression
```
kanitw, line 19: Add comment documenting why we need to export default twice?

packages/vega-expression/index.js
```diff
@@ -14,7 +14,13 @@ export {
...
+} from './src/parser';
+export {
+  default as codegen,
+  default as codegenExpression
```
kanitw, line 23: Add comment documenting why we need to export default twice?

kanitw (changes requested):
It seems that we're exporting redundantly.  Include commments to explain why?

packages/vega-expression/index.js
domoritz: I think it's because Vega exports it this way as well. It's so that we can import from Vega or this specific package without changing the import.

domoritz: `default as parse` because that's what it used to be so we don't want to break backward compatibility

`default as parseExpression` because the vega package uses this name: https://github.com/vega/vega/blob/1f14aa2acfa34825e3809a7076553504b0ef90db/packages/vega/index.js#L93. In order to swap out a specific package for the vega bundle, we need to use the same names.

I added a comment.

PR #3157, opened 2021-04-10
feat: Reference spec source in DataFlow nodes

danmarshall:
This PR adds a new property `src` to DataFlow nodes. This references the specification property which spawned each node.
It is helpful to be able to map the scenegraph item to the specification.

Scenario: I have a custom Vega renderer which traverses the scenegraph and renders in WebGL. I am using Vega [axes](https://vega.github.io/vega/docs/axes/). I want one of these axes rendered in a 3D plane instead of 2D.

Example: In my spec for the axes, I add a new attribute `plane`:
```json
"axes": [
        { "orient": "bottom", "scale": "x", "title": "X Axis"},
        { "orient": "left", "scale": "y", "title": "Y Axis" },
        { "orient": "left", "scale": "z", "title": "Z Axis", "plane": "z" },
    ],
```
During the scenegraph render pass, I can get to the mark node generated by the DataFlow, and via the `src` property I can get to this axis object from the spec and get the `plane` attribute.

This may also enable other scenarios, such as animation attributes etc.

### DIFF
--- a/packages/vega-parser/src/DataScope.js
+++ b/packages/vega-parser/src/DataScope.js
@@ -63,7 +63,7 @@ function addSortField(scope, p, sort) {
   }
 }
 
-function cache(scope, ds, name, optype, field, counts, index) {
+function cache(scope, ds, name, optype, field, counts, index, src) {
   const cache = ds[name] || (ds[name] = {}),
         sort = sortKey(counts);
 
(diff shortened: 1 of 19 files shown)


jheer:
Thanks for the PR, and apologies for the delayed review. I'm a bit skeptical of annotating the runtime with specification source: this violates a "separation of concerns" across levels of abstraction and will also add bloat to the runtime.

That said, supporting extensibility (e.g., to new renderers) is a nice goal. Vega already supports this to some degree: any arbitrary encoding property can be defined and will be passed through as a property written to scenegraph items. This applies to axes and legends as well, but in those cases requires using a [custom encoding block](https://vega.github.io/vega/docs/axes/#custom-axis-encodings). For example, you could write custom properties to the `axis` group mark and have your custom renderer respond to those.

Issue #2736, opened 2020-06-26
Standardize Vega dataflow format


domoritz:
I am proposing to standardize the Vega dataflow format. It's very useful as a low-level representation and it would be great to be able to rely on it. To standardize it, we need to

1) Version the vega dataflow format and define what it means to break backwards compatibility.
2) Define the format. Maybe this could be through the implementation or typescript typing. 

A major change in the dataflow format may still only be a minor change in Vega.

hydrosquall:
I like this idea, especially since debugger tools (like the Dataflow visualizer from @saulshanabrook ) would depend on the format. 

It will also help us to enforce that the dataflow stays JSON serializable (no functions attached), which could facilitate future work on multi-thread processing, where the dataflow and scene graph are processed separately (webworkers etc)

domoritz:
I think it makes sense for the format to be implementation driven for the time being. We can formalize it when needed and try to avoid breaking changes (while not promising that we don't make breaking changes).

PR #3237, opened 2021-06-06
Add typings for Vega runtime dataflow

saulshanabrook:
This PR begins to add typings for the vega runtime dataflow, which would partially address #2736.

I am working on this in conjunction with adding a runtime dataflow viewer to the vega editor (vega/editor#1023).

Not ready for review yet, but I wanted to open this since the other PR is already open.

### DIFF
--- a/packages/vega-typings/types/runtime/index.d.ts
+++ b/packages/vega-typings/types/runtime/index.d.ts
@@ -15,8 +15,8 @@ import { Changeset, Transform } from './dataflow';
 import { Scene } from './scene';
 import { LoggerInterface } from 'vega-util';
 
-// TODO
-export type Runtime = any;
+export { Runtime } from './runtime';
+import { Runtime } from './runtime';
 
 export const version: string;
 
--- a/packages/vega-typings/types/runtime/runtime.d.ts
+++ b/packages/vega-typings/types/runtime/runtime.d.ts
@@ -1,0 +1,65 @@
+import { EventStream } from "..";
+
+export type Runtime = {
+    description: string,
+    operators: Entry[],
+    streams: Stream[],
+    updates: any,
+    bindings: any,
+    eventConfig: any,
+    locale?: any
+};
+
+// From Scope.js:Scope:add -> transforms.js
+type Entry = { id: id } & (
+    {
+        type: 'operator',
+        // Adding root property in Scope.finish()
+        // Adding root operator node in parseView
+        root?: true
+    } |
+    { type: 'aggregate', value?: any, params?: any, parent?: any } |
+    { type: 'axisticks', value?: any, params?: any, parent?: any } |
+    { type: 'bound', value?: any, params?: any, parent?: any } |
+    { type: 'collect', value?: any, params?: any, parent?: any } |
+    { type: 'compare', value?: any, params?: any, parent?: any } |
+    { type: 'datajoin', value?: any, params?: any, parent?: any } |
+    { type: 'encode', value?: any, params?: any, parent?: any } |
+    { type: 'expression', value?: any, params?: any, parent?: any } |
+    { type: 'extent', value?: any, params?: any, parent?: any } |
+    { type: 'facet', value?: any, params?: any, parent?: any } |
+    { type: 'field', value?: any, params?: any, parent?: any } |
+    { type: 'key', value?: any, params?: any, parent?: any } |
+    { type: 'legendentries', value?: any, params?: any, parent?: any } |
+    { type: 'load', value?: any, params?: any, parent?: any } |
+    { type: 'mark', value?: any, params?: any, parent?: any } |
+    { type: 'multiextent', value?: any, params?: any, parent?: any } |
+    { type: 'multivalues', value?: any, params?: any, parent?: any } |
+    { type: 'overlap', value?: any, params?: any, parent?: any } |
+    { type: 'params', value?: any, params?: any, parent?: any } |
+    { type: 'prefacet', value?: any, params?: any, parent?: any } |
+    { type: 'projection', value?: any, params?: any, parent?: any } |
+    { type: 'proxy', value?: any, params?: any, parent?: any } |
+    { type: 'relay', value?: any, params?: any, parent?: any } |
+    { type: 'render', value?: any, params?: any, parent?: any } |
+    { type: 'scale', value?: any, params?: any, parent?: any } |
+    { type: 'sieve', value?: any, params?: any, parent?: any } |
+    { type: 'sortitems', value?: any, params?: any, parent?: any } |
+    { type: 'viewlayout', value?: any, params?: any, parent?: any } |
+    { type: 'values', value?: any, params?: any, parent?: any }
+)
+
+// from `Scope.js:Scope:id`
+// String if sub id with `:` seperate parent from child id numbers
+type id = string | number
+// from `parsers/stream.js:eventStream` function
+type Stream = {
+    id: id
+} &
+    ({
+        source: "timer", type: number
+    } | {
+        source: Omit<EventStream['source'], 'scope'>,
+        type: EventStream['type'],
+    })
+// TODO: Add all calls of `addStream`


saulshanabrook:
I would like to run `vega.parse` on all the vega examples, and set breakpoints and tracing in the `vega-parser` to better understand how it's used. Do you know if any of the existing test suites do this, so I could reuse the functionality?

Otherwise, I can try writing a node script to do this one off task.

packages/vega-typings/types/runtime/index.d.ts
```diff
@@ -15,8 +15,8 @@ import { Changeset, Transform } from './dataflow';
 import { Scene } from './scene';
 import { LoggerInterface } from 'vega-util';
 
-// TODO
-export type Runtime = any;
+export { Runtime } from './runtime';
+import { Runtime } from './runtime';
```
domoritz, line 19:
```suggestion
```

packages/vega-typings/types/runtime/runtime.d.ts
```diff
@@ -0,0 +1,251 @@
...
+export type RenderEntry = EntryType<'render', { [k: string]: unknown }>;
+export type ScaleEntry = EntryType<'scale', { [k: string]: unknown }>;
+export type SieveEntry = EntryType<'sieve', { [k: string]: unknown }>;
+export type SortItemsEntry = EntryType<'sortitems', { [k: string]: unknown }>;
+export type ViewLayoutEntry = EntryType<'viewlayout', { [k: string]: unknown }>;
+export type ValuesEntry = EntryType<'values', { [k: string]: unknown }>;
+
+export type EntryType<NAME extends string, BODY extends { [k: string]: unknown }> = {
+  id: id;
+  type: NAME;
```
domoritz, line 181:
```suggestion
  type: T;
```

packages/vega-typings/types/runtime/runtime.d.ts
```diff
@@ -0,0 +1,251 @@
...
+  // Scope.add
+  refs?: null;
+  // Scope.finish.annotate
+  data?: { [name: string]: ('input' | 'output' | 'values' | `index:${string}`)[] };
+} & BODY;
```
domoritz, line 186:
```suggestion
} & B;
```

packages/vega-typings/vega-to-dataflow.js
```diff
@@ -0,0 +1,13 @@
...
+
+const vega = require('vega');
+const fs = require('fs');
+
+const specPath = process.argv[2];
+const outputPath = process.argv[3];
+const spec = require(specPath);
+const dataflow = vega.parse(spec);
+const dataflowStr = JSON.stringify(dataflow);
+fs.appendFileSync(outputPath, dataflowStr + ';');
```
domoritz, line 13:
```suggestion
fs.appendFileSync(outputPath, dataflowStr + ';');

```

packages/vega-typings/types/runtime/index.d.ts
saulshanabrook: The `Runtime` type is used in the `parse` function in this file, so I found I had to import it as well as export.

saulshanabrook:
There are some types, like the `Update['update']` field, which are either some value or an object with keys `$expr` and `$params`. At first, I typed it like `unknown | {$expr: ...}` but this ends up being just `unknown`. What I want to say is that _if_ the value has  either of the keys `$expr` and `$params` then those keys have certain types. If not, then it can be anything. (see runtime code that handles this condition: https://github.com/vega/vega/blob/2d1462476d39889c31d23e5970296929ff2405bf/packages/vega-runtime/src/update.js#L20-L25)

Any ideas on the best ways to type this? I think related to microsoft/TypeScript#41524 and microsoft/TypeScript#41524.

packages/vega-typings/types/runtime/runtime.d.ts
```diff
@@ -0,0 +1,251 @@
...
+
+// // parsers.transform.js:parseParameter
+export type TransformParam =
+  // parseIndexParameter
+  | Ref
+  // parseSubParameters
+  | Ref[]
+  // parameterValue
+  | Parse
+  | unknown;
```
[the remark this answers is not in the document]
saulshanabrook: Replaced the unknown with a different types that allows other objects and primitives without just expanding to `unknown`

packages/vega-typings/types/runtime/runtime.d.ts
saulshanabrook: Removed this generic

saulshanabrook:
This is ready for another review. I have added tests that run this against all the example dataflows. I am also trying it out in the new Vega dataflow viewer.

packages/vega-typings/types/runtime/runtime.d.ts
```diff
@@ -0,0 +1,370 @@
+import { Config, Format } from '../spec';
+import { EventType, WindowEventType, Transforms, Binding as SpecBinding } from '..';
+// All references to source code are from the vega-parser package
+
+export interface Runtime {
+  description: string;
+  operators: Operator[];
+  // Event streams that Vega is listening to. Often DOM events.
```
domoritz, line 8: Let's use /**...*/ comments so that we get previews in vscode.

domoritz (changes requested):

packages/vega-typings/types/runtime/runtime.d.ts
saulshanabrook: Sounds good, will update

saulshanabrook:
I realize the tests are failing, because now the `vega-typings` test depend on having `vega` compiled first, so that they can use `vega-parser` with all the custom transforms registered to generate the dataflows for all the examples.

So I would like the `vega` build to happen before the `vega-typings` test, but I can't add a dependency, because then it would be cyclical... So I am thinking of just running the `vega-typings` tests after all the other tests have built and finished?

package.json
```diff
@@ -12,7 +12,8 @@
     "license": "lerna exec -- cp ../../LICENSE .",
     "release": "yarn license && lerna publish from-package",
     "serve": "serve packages/vega/ -l 8080",
-    "test": "lerna run test && yarn lint",
+    "test": "yarn run test && yarn run lint",
```
jheer, line 15: This looks circular: `yarn test` calls `yarn run test`? Should this be `yarn test:no-lint && yarn lint`?

saulshanabrook, line 15:
```suggestion
    "test": "yarn run test:no-lint && yarn run lint",
```

saulshanabrook: Yeah, you are right, thanks.

package.json
```diff
@@ -12,7 +12,8 @@
     "license": "lerna exec -- cp ../../LICENSE .",
     "release": "yarn license && lerna publish from-package",
     "serve": "serve packages/vega/ -l 8080",
-    "test": "lerna run test && yarn lint",
+    "test": "yarn run test:no-lint && yarn run lint",
```
jheer, line 15: I believe this should be `yarn test:no-lint && yarn lint`. No `run` required.

saulshanabrook: You are right, I didn't realize that the run was optional.

packages/vega-typings/types/runtime/index.d.ts
domoritz: I see. What I usually do is import and then do something like `export cost foo`. But I see that's not our style here so ignore my comment.

packages/vega-typings/tslint.json
```diff
@@ -6,6 +6,7 @@
     "max-line-length": false,
     "semicolon": false,
     "whitespace": false,
-    "file-name-casing": false
+    "file-name-casing": false,
+    "no-irregular-whitespace": false
```
domoritz, line 10: Why?

domoritz (approved):

packages/vega-typings/tslint.json
saulshanabrook: In `packages/vega-typings/tests/dataflow/bar-time.ts` on line 2310 it has:

```
number: { decimal: ',', thousands: '.', grouping: [3], currency: ['', ' €'] },
```

The second currency raises an error with dtslint:

```
Error: /Users/saul/p/vega/packages/vega-typings/tests/dataflow/bar-time.ts:2310:76
ERROR: 2310:76  no-irregular-whitespace  Irregular whitespace not allowed

    at /Users/saul/p/vega/node_modules/dtslint/bin/index.js:207:19
    at Generator.next (<anonymous>)
    at fulfilled (/Users/saul/p/vega/node_modules/dtslint/bin/index.js:6:58)
error Command failed with exit code 1.
```

domoritz: Hmm, interesting. Oh well, we need to replace it with eslint eventually anyway.

domoritz:
@jheer Could you make a final pass so we can merge, release, and unblock https://github.com/vega/editor/pull/1023?

PR #3248, opened 2021-06-15
Update title.js

MartinA731:
PR to fix issue #[6894 ](https://github.com/vega/vega-lite/issues/6894) in vega-lite 

The `config.title.align, config.title.angle, config.title.limit,` and `config.title.baseline` are being ignored on rendering in vega-lite. Meanwhile, `config.title.dx, config.title.dy, config.title.font`, and other title configurations are working fine. 

This, however, appears to be a Vega issue instead of a Vega-lite issue. When I set dx, limit, and angle in config.style["group-title"], only dx is changed:

```
{
  "width" : 500,
  "height" : 500,
  "title": {
    "text": "Hello"
          },
  "config": {
    "style" : {
      "group-title" :
      {
        "dx" : 100,
        "limit" : 1,
        "angle" : 100
      }
    }
  }
}
```

### DIFF
--- a/packages/vega-parser/src/parsers/title.js
+++ b/packages/vega-parser/src/parsers/title.js
@@ -92,10 +92,10 @@ function buildTitle(spec, _, userEncode, dataRef) {
 
   addEncoders(encode, {
     text:       text,
-    align:      {signal: 'item.mark.group.align'},
-    angle:      {signal: 'item.mark.group.angle'},
-    limit:      {signal: 'item.mark.group.limit'},
-    baseline:   'top',
+    align:      _('align'),
+    angle:      _('angle'),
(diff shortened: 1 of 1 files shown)


jheer:
I'm not sure this PR is a safe change to make. While perhaps a bit unsightly, the calls to `item.mark.group.{property}` are there to ensure shared configuration for both title and subtitle, as written to their enclosing group element.

So the place to look further is the `groupEncode` method in the same file (`title.js` in vega-parser). The `limit` property should already be pulling from the config. The `align` and `angle` properties have specialized logic to make the response to the choice of orientation (`orient`). Additional logic would be needed there if we want to have a config value override the use of orientation-sensitive values.

PR #3252, opened 2021-06-27
 Allow label-transform to place labels out of a chart's boundary with padding: `'Infinity'` or `Infinity`

chanwutk:
Add an option to `label-transform`.
Allow `padding: 'inf'`.

### DIFF
--- a/packages/vega-label/src/Label.js
+++ b/packages/vega-label/src/Label.js
@@ -33,7 +33,7 @@ const Anchors = [
  *   'bottom', 'top-right', 'right', 'bottom-right', 'middle'.
  * @param {Array<number>} [params.offset] - Label offsets (in pixels) from the base mark bounding box.
  *   This parameter  is parallel to the list of anchor points.
- * @param {number | 'inf'} [params.padding=0] - The amount (in pixels) that a label may exceed the layout size.
+ * @param {number | 'Infinity'} [params.padding=0] - The amount (in pixels) that a label may exceed the layout size.
  * @param {string} [params.lineAnchor='end'] - For group line mark labels only, indicates the anchor
  *   position for labels. One of 'start' or 'end'.
  * @param {string} [params.markIndex=0] - For group mark labels only, an index indicating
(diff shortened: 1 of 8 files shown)


domoritz:
Hmm, can someone not already use `Infinity`?

chanwutk:
@domoritz Do you mean we should let users pass in the value `Infinity` instead?
I thought that JSON does not support `Infinity` like JavaScript.

domoritz:
Ahh, right. We should still support it for people who don't use Vega via json, I think.

chanwutk:
So, should we make the behavior for `padding: Infinity` to be the same as `padding: 'inf'`?

domoritz:
> So, should we make the behavior for padding: Infinity to be the same as padding: 'inf'?

I would expect that to be the case.

chanwutk:
Should the string value be `'inf'` or `'infinity'`?

jheer:
The nice part about `'inf'` is that it is fairly readable, though I do think `'Infinity'` is clearer and more consistent with JS numbers. However, that also changes the type signature of the parameter... One alternative might include using `null` to indicate no bounds, but I think that might be confusing for a parameter named "padding" -- I'd associate that with having no padding (or at least a default value). Another option might be to use negative numbers to indicate Infinite padding -- that is not as clear as text but seems reasonable as a way of saying "no limits" and keeps the types tidier (unless we already support negative numbers as a way of forcing an inset margin). Any thoughts on this, @chanwutk or @domoritz? I'm OK moving ahead with the text string `'Infinity'` unless someone strongly prefers another option.

I also agree with Dominik that numeric `Infinity` should be supported for specs that are defined in JS directly (not JSON text).

chanwutk:
Thank you for the comment.
We already support negative padding value (this will tighten the space allowed for labels to be placed).
I am also OK with the text string `'Infinity'`.

jheer:
OK, so let's go with `'Infinity'` and then I think we can merge.

jheer:
I see that that the typings have been updated, but the metadata for Vega's JSON schema has not been updated:
https://github.com/vega/vega/blob/cc7733fd86b42470202867bb9bede8c316dd73f6/packages/vega-label/src/Label.js#L62

chanwutk:
I see.
How about if we omit padding, the padding is default to 0.
But, if we set padding to `null`, then the padding can be Infinity?

jheer:
Yes, null would work. We just need to include `null: true` as part of the parameter metadata.

Issue #3255, opened 2021-07-03
Encoding.sort doesn't work with field containing a dot


christophe-petitjean:
If fields contain a dot, the `encoding.sort` doesn't work even escaped (but `encoding.field` work with the same escaping character).

Here an example : 
```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v5.1.0.json",
  "data": {
    "values": [
      {"Continent": "Asia", "Population.": 1367486},
      {"Continent": "Europe", "Population.": 80855},
      {"Continent": "Africa", "Population.": 181563},
      {"Continent": "Oceania", "Population.": 22752},
      {"Continent": "North America", "Population.": 321369},
      {"Continent": "Antarctica", "Population.": 5},
      {"Continent": "South America", "Population.": 204260}
    ]
  },
  "mark": "bar",
  "encoding": {
    "x": {
      "field": "Continent",
      "type": "nominal",
      "sort": {"field": "Population\\.", "order": "descending"}
    },
    "y": {"field": "Population\\.", "type": "quantitative"}
  }
}
```

<img width="235" alt="image" src="https://user-images.githubusercontent.com/354149/124004846-b7845b00-d9d8-11eb-9756-c80878bdb935.png">

[...]

PR #3259, opened 2021-07-13
Update util.js

MartinA731:
encoding.sort does not seem to work when the field has a dot ([Issue 3225](https://github.com/vega/vega/issues/3255)). It also does not seem to work when there are brackets in the field, for example:

```
{
"$schema": "https://vega.github.io/schema/vega-lite/v5.1.0.json",
"data": {
"values": [
{"Continent": "Asia", "Population[": 1367486},
{"Continent": "Europe", "Population[": 80855},
{"Continent": "Africa", "Population[": 181563},
{"Continent": "Oceania", "Population[": 22752},
{"Continent": "North America", "Population[": 321369},
{"Continent": "Antarctica", "Population[": 5},
{"Continent": "South America", "Population[": 204260}
]
},
"mark": "bar",
"encoding": {
"x": {
"field": "Continent",
"type": "nominal",
"sort": {"field": "Population\\[", "order": "descending"}
},
"y": {"field": "Population\\[", "type": "quantitative"}
}
}
```

### DIFF
--- a/packages/vega-parser/src/util.js
+++ b/packages/vega-parser/src/util.js
@@ -61,8 +61,8 @@ export function sortKey(sort) {
 export function aggrField(op, field) {
   const removeUnnecessary = (str) => {
     if(str) {
-      s = ''
-      for(i = 0; i < str.length; i++) {
+      let s = ''
+      for(let i = 0; i < str.length; i++) {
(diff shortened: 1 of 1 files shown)


jheer:
Thanks for this. I will try to look at this closer soon. I think we may want to reuse the existing facilities to parsing path access expressions rather than introduce a new method here. We also need test cases for this issue.

PR #3262, opened 2021-07-15
[logger] Allow a custom log handler as optional parameter

challet:
Add an extension capacity for using the native logger as a base for custom ones.

### DIFF
--- a/packages/vega-util/index.d.ts
+++ b/packages/vega-util/index.d.ts
@@ -161,5 +161,5 @@ export const Info: number;
 export const Debug: number;
 
 export function logger(_?: number, method?: string, handler?: typeof log): LoggerInterface;
-export function log(...args: readonly any[]): void;
+export function log(method: 'error' | 'warn' | 'log', level: 'error' | 'warn'| 'info' | 'debug', input: readonly any[]): void;
 export function error(msg: string): Error;
--- a/packages/vega-util/index.d.ts
+++ b/packages/vega-util/index.d.ts
@@ -160,6 +160,6 @@ export const Warn: number;
 export const Info: number;
 export const Debug: number;
 
-export function logger(_?: number, method?: string): LoggerInterface;
+export function logger(_?: number, method?: string, handler?: typeof log): LoggerInterface;
 export function log(...args: readonly any[]): void;
 export function error(msg: string): Error;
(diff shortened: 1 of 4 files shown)


domoritz:
Can you also update the types in Vega-typings?

challet:
@domoritz sure thing, but I'm not really sure about what I'm doing here. Is there a command that I can use to check the said types ?

packages/vega-util/index.d.ts
```diff
@@ -160,6 +160,6 @@ export const Warn: number;
 export const Info: number;
 export const Debug: number;
 
-export function logger(_?: number, method?: string): LoggerInterface;
+export function logger(_?: number, method?: string, handler?: typeof log): LoggerInterface;
 export function log(...args: readonly any[]): void;
```
challet, line 164: Looks like the type of the `log` function used by the logger

domoritz:
If you add a test case, you can run yarn test in the Vega typings package.

packages/vega-util/index.d.ts
domoritz: You can add a usage in https://github.com/vega/vega/tree/master/packages/vega-typings/tests/runtime and that will test the typings.

challet: Thank you for pointing this out. It's done, along with expliciting these arguments types.

packages/vega-util/index.d.ts
```diff
@@ -160,6 +160,6 @@ export const Warn: number;
 export const Info: number;
 export const Debug: number;
 
-export function logger(_?: number, method?: string): LoggerInterface;
-export function log(...args: readonly any[]): void;
+export function logger(_?: number, method?: string, handler?: typeof log): LoggerInterface;
+export function log(method: 'error' | 'warn' | 'log', level: 'error' | 'warn'| 'info' | 'debug', input: readonly any[]): void;
```
domoritz, line 164: Should the level be all caps like in https://github.com/vega/vega/pull/3262/files#diff-377e9062f2f35efd117c4c0b89857464b8f54af1c5a8e1e648cc0683b07ca895R24?

challet: That is absolutely right, sorry for the confusion

jheer:
Thanks!

PR #3280, opened 2021-08-14
build: add script for generating offline docsets

hydrosquall:
## Motivation

- Vega equivalent of https://github.com/vega/vega-lite/pull/7642
- Enable offline browsing of Vega docs, even when browsing directly from a filesystem and not a webserver

## Changes

- Added a script to package.json along with a custom copy of the jekyll build config 
- Add switches to toggle SEO tags + the base URL used for fetching data (Due to browser security constraints, you can't use HTTPRequest to fetch files from the file system without a live server). 

## Testing

- Works together with the output from https://github.com/hydrosquall/vega-docset-generator


domoritz:
What's left here or shall we close this?

domoritz:
I think a general refactor of the docs would be good to make it cleaner and more friendly for LLMs. You can leave the pull request open since it's on your queue.

PR #3300, opened 2021-09-23
style: use let and const in vega-util

domoritz:
Follow up from #3228. This pull request splits multi-variable declarations and switches to let/const in vega-util.


domoritz:
Didn't get a review so outdated.

PR #3312, opened 2021-09-29
fix(vega-typings): narrow type signature for vega-loader.http

hydrosquall:
## Motivation

- Providing a custom data fetcher, and noticed the response type could be narrowed.
- ~Checked `vega-loader`, I think it's already behaving as if this type is a `Response` rather than a `string`.~ (You can't use fetch directly, you have to wrap it so that it returns a string. See comment below for a usage example. )

## References

- RequestInit: https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_dom_d_.requestinit.html
- fetch: https://github.com/microsoft/TypeScript/blob/e96c10fe0ced0aa27ff40f0108df8e7153b348c7/lib/lib.dom.d.ts#L17373
- https://github.com/vega/vega/blob/774165e29850b66ec8b79ba52a7955f1ab936ea6/packages/vega-loader/src/loader.js#L175-L180

### DIFF
--- a/packages/vega-typings/types/runtime/index.d.ts
+++ b/packages/vega-typings/types/runtime/index.d.ts
@@ -28,7 +28,7 @@ export function parse(spec: Spec, config?: Config, opt?: { ast?: boolean }): Run
 export interface Loader {
   load: (uri: string, options?: any) => Promise<string>;
   sanitize: (uri: string, options: any) => Promise<{ href: string }>;
-  http: (uri: string, options: any) => Promise<string>;
+  http: (uri: string, options: RequestInit) => Promise<string>;
   file: (filename: string) => Promise<string>;
 }
 


hydrosquall:
I misspoke - it turns out you can't use fetch directly specifically because the previous . Something to this effect works as a default fetching function.

```ts
return fetch(url, options).then(response => {
            if (response.ok) {
                return response.text();
            }
            return response.text().then(text => {
                throw new Error(text);
            });
        });
```

domoritz:
In https://github.com/vega/vega/blob/774165e29850b66ec8b79ba52a7955f1ab936ea6/packages/vega-loader/src/loader.js#L174, we read `response`, which I don't see in `RequestInit`. ARe you sure the type is right?

hydrosquall:
@domoritz, ah, good catch. I was going off of this comment from the `README.md`, but you're right that `RequestInit` properties are actually nowhere to be found.

From logging out the actual output at runtime,

Image 2021-09-30 at 2 24 50 PM

this is where the options object is actually populated. 

https://github.com/vega/vega/blob/02d79dddb2850a24abfe6a4ebaf8823548709bb5/packages/vega-dataflow/src/dataflow/load.js#L42-L4

The response type would be one of these, or `text`.

https://github.com/vega/vega/blob/02d79dddb2850a24abfe6a4ebaf8823548709bb5/packages/vega-loader/src/formats/index.js#L7-L11

Do you think the fix here would be to update the README for vega-loader (text copied below), update the implementation, or both?

> If provided, the options argument may include any valid fetch [RequestInit](https://fetch.spec.whatwg.org/#requestinit) properties. The provided options will be combined with any default options passed to the loader constructor under the http property.

domoritz:
For questions about the code and readme, @jheer knows best. The typings should follow the implementation.

hydrosquall:
This might be replaced by https://github.com/vega/vega/pull/4000/ .

hydrosquall:
This is replaced by https://github.com/vega/vega/pull/4000

PR #3320, opened 2021-10-07
Create boston_weather

95abhijeet:


jheer:
Any new datasets should be added to the vega-datasets repo, then we can import them into vega/vega dynamically.

PR #3388, opened 2022-01-14
Vega Expressions for Lasso Selection

dvmoritzschoefl:
lassoexample

This PR includes the following vega expressions:

- **lassoPath** (generates a svg path from a lasso signal, can be directly used on a path mark)
- **lassoAppend** (handles the logic to append a point to the lasso signal)
- **intersectLasso** (performs the intersection tests for the lasso and returns the appropriate tuples)

There is an example specification added that allows to test the lasso (called lasso.vg.json). If this PR gets approved, I will create another PR for vega-lite that includes the generator for lasso selections.


arvind (changes requested):
Hi @dvmoritzschoefl, this PR looks like a great start. I've left some line/block-level suggestions but three high-level thoughts:

1. I think it'd be useful to have these lasso functions remain as general purpose as possible rather than tightly coupled with Vega-Lite selections. So, I recommend removing any references to selections here. Instead, have this function return an array of scene graph items within the lasso and, in the expression string generated byVega-Lite, pass this array to the `vlSelectionTuples` function introduced in #3009 and improved in #3397 to extract `SELECTION_ID` and populate the selection stores.

2. Relatedly, I'm not sure I'm totally following why we have this function inverting pixels -> data coordinates. Can we not stay entirely in pixel space for the entirety of the hit testing process (i.e., an `intersectLasso` function instead)? I'm imagining the same steps you have currently (start with a lasso path, calculate the bounding box, `intersect` it with the scenegraph, and then validate which `pointInPolygon`), but you'd never make a roundtrip to data space at all. Instead, and following (1) above, we'd only go into data space once we have all the scene graph items that fall within the lasso and just to extract the `SELECTION_ID`. Does this make sense? I could very well believe I'm missing something though. 

3. A more minor point: in Vega and Vega-Lite, we default to a 2-space indent. Do you mind updating this file to follow that format? Thanks!

dvmoritzschoefl:
> Hi @dvmoritzschoefl, this PR looks like a great start. I've left some line/block-level suggestions but three high-level thoughts:
> 
> 1. I think it'd be useful to have these lasso functions remain as general purpose as possible rather than tightly coupled with Vega-Lite selections. So, I recommend removing any references to selections here. Instead, have this function return an array of scene graph items within the lasso and, in the expression string generated byVega-Lite, pass this array to the `vlSelectionTuples` function introduced in #3009 and improved in #3397 to extract `SELECTION_ID` and populate the selection stores.
> 2. Relatedly, I'm not sure I'm totally following why we have this function inverting pixels -> data coordinates. Can we not stay entirely in pixel space for the entirety of the hit testing process (i.e., an `intersectLasso` function instead)? I'm imagining the same steps you have currently (start with a lasso path, calculate the bounding box, `intersect` it with the scenegraph, and then validate which `pointInPolygon`), but you'd never make a roundtrip to data space at all. Instead, and following (1) above, we'd only go into data space once we have all the scene graph items that fall within the lasso and just to extract the `SELECTION_ID`. Does this make sense? I could very well believe I'm missing something though.
> 3. A more minor point: in Vega and Vega-Lite, we default to a 2-space indent. Do you mind updating this file to follow that format? Thanks!

1. I used the functions you mentioned, the function is now called 'intersectLasso' instead and does not use data space
2. solved
3. updated to 2-space indent

attached is a working lasso spec
[lasso.vg.zip](https://github.com/vega/vega/files/7925282/lasso.vg.zip)

packages/vega-functions/src/functions/lasso.js
```diff
@@ -0,0 +1,116 @@
...
+ * https://wrf.ecse.rpi.edu/Research/Short_Notes/pnpoly.html
+ * 
+ * This method will not need the same start/end point since it wraps around the edges of the array
+ * 
+ * @param {*} test a point to test against
+ * @param {*} polygon a polygon in the form [[x,y], [x,y], ...]
+ * @returns true if the point lies inside the polygon, false otherwise
+ */
+function pointInPolygon(test, polygon) {
+  const [testx, testy] = test;
```
arvind, line 101:
```suggestion
function pointInPolygon(testx, testy, polygon) {
```
Seemed like an unnecessary array being constructed in `intersectLasso` to merely be destructured here?

packages/vega-functions/src/functions/lasso.js
```diff
@@ -0,0 +1,116 @@
...
+  var intersections = 0;
```
arvind, line 102:
```suggestion
  let intersections = 0;
```

packages/vega-functions/src/functions/lasso.js
```diff
@@ -0,0 +1,116 @@
...
+
+  for (var i = 0, j = polygon.length - 1; i < polygon.length; j = i++) {
```
arvind, line 104:
```suggestion
  let (var i = 0, j = polygon.length - 1; i < polygon.length; j = i++) {
```

packages/vega-functions/src/functions/lasso.js
```diff
@@ -0,0 +1,116 @@
...
+  for (const [i, [x, y]] of lasso.entries()) {
+    svg += i === 0
+      ? `M ${x},${y} `
+      : i === lasso.length - 1
+        ? ' Z'
+        : `L ${x},${y} `;
+  }
+
+  return svg;
+}
```
arvind, line 44:
```suggestion
export function lassoPath(lasso) {
  return lasso.entries().reduce((svg, [i, [x, y]]) => {
    return svg += i === 0
      ? `M ${x},${y} `
      : i === lasso.length - 1
        ? ' Z'
        : `L ${x},${y} `
  }, '');
}
```

packages/vega-functions/src/functions/lasso.js
```diff
@@ -90,7 +90,7 @@ export function intersectLasso(markname, pixelLasso) {
 function pointInPolygon(testx, testy, polygon) {
   let intersections = 0;
 
-  for (var i = 0, j = polygon.length - 1; i < polygon.length; j = i++) {
+  let (var i = 0, j = polygon.length - 1; i < polygon.length; j = i++) {
```
arvind, line 93: Sorry, heads up, I typo'd my suggestion.

packages/vega-functions/src/functions/lasso.js
```diff
@@ -0,0 +1,105 @@
...
+
+
+/**
+ * Generates a svg path command which draws a lasso
+ * 
+ * @param {*} lasso the lasso in pixel space in the form [[x,y], [x,y], ...]
+ * @returns the svg path command that draws the lasso
+ */
+export function lassoPath(lasso) {
+  return (lasso ?? []).reduce((svg, [x, y], i) => {
```
dvmoritzschoefl, line 34: using entries().reduce gave an error for me, this version works
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce

dvmoritzschoefl:
> Hi @dvmoritzschoefl, thanks for making the changes! Looks like we're pretty close. I left a handful of comments, just to make the code a bit more functional rather than imperative. Two other high-level comments:
> 
> * `packages/vega/test/specs-valid.json` is currently deleted but shouldn't be.
> * `docs/*.js` will be built as part of the final version release, so you don't need to include them in this PR.

Yes sry I made a mistake checking these files in, should be ok by now. I changed your suggestions accordingly and made sure the lasso still works.

arvind (approved):
Thanks @dvmoritzschoefl! The updates look good to me, and apologies for the malformed suggestions — the dangers of writing JavaScript code on-the-fly in the GitHub editor textbox 🤦 

I've cc'ed @jheer to take a final look in case he notices anything I missed, and then we can merge 🙂

dvmoritzschoefl:
@arvind very nice, I will adjust the vega-lite PR to those changes tomorrow and ping you when I am finished

jheer (approved):

PR #3391, opened 2022-01-19
chore: upgrade to @definitelytyped/dtslint

domoritz:
https://github.com/microsoft/dtslint mentions that the package moved. 

I also enabled skipping lib checks since we don't need to check library types.

Otherwise we get this error with ts 4.6 right now

```
Error: Errors in typescript@4.6 for external dependencies:
9374
Error: ../../node_modules/@types/node/globals.d.ts(72,13): error TS2403: Subsequent variable declarations must have the same type.  Variable 'AbortSignal' must be of type '{ new (): AbortSignal; prototype: AbortSignal; abort(reason?: any): AbortSignal; }', but here has type '{ new (): AbortSignal; prototype: AbortSignal; }'.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "@rollup/plugin-babel": "^5.3.0",
     "@rollup/plugin-json": "^4.1.0",
     "@rollup/plugin-node-resolve": "^13.1.1",
-    "@types/node": "^17.0.2",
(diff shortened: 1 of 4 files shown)


arvind (approved):
Thanks for tracking this down and fixing it @domoritz!

PR #3392, opened 2022-01-19
Use willReadFrequently on 2D canvas contexts, where appropriate

junov:
willReadFrequently is a performance hint that tells browsers to
expect calls to getImageData on that context.  This allows browsers
to optimize performance for readback use cases.

See: https://html.spec.whatwg.org/multipage/canvas.html#concept-canvas-will-read-frequently

This feature will soon be released in Chromium-based Web Browsers
(i.e. Chrome, Edge, Opera, etc.)

This pull request aims to improve performance and prevent possible
performance regressions when the new API goes live in various Web
browsers.


domoritz:
@junov what's your reason for closing this?

PR #3426, opened 2022-03-09
Adding Packed Bubble Chart Example

PBI-David:
@domoritz adding a packed bubble chart example. I think I have added the required files in the right places but this is my first time contributing and using GitHub so apologies in advance if I haven't got it completely right.


PBI-David:
Just spotted my mouse in the png so will recreate.

PR #3427, opened 2022-03-09
Packed Bubble Chart Example

PBI-David:
@domoritz adding a packed bubble chart example. I think I have added the required files in the right places but this is my first time contributing and using GitHub so apologies in advance if I haven't got it completely right.


jheer:
Thanks @PBI-David! Retargetting PR to `v5.22.0` branch for staging prior to next release.

PBI-David:
@jheer I just checked the new release and FYI when clicking the link for "View in Online Vega Editor" [here](https://vega.github.io/vega/examples/packed-bubble-chart/), there seems to be a problem loading the spec. Apologies if I missed something.

domoritz:
Yes, instead of sending the spec, for examples we just refer to the example by id. This means we need to update the Vega package in the editor.

domoritz:
Done. See https://vega.github.io/editor/#/examples/vega/packed-bubble-chart.

PR #3455, opened 2022-03-25
feat: add stroke, fill, and opacity to various SceneItem types

lsh:
In [vega-scenegraph](https://github.com/vega/vega/tree/main/packages/vega-scenegraph), several fields are expected on the runtime types which are not present in the TypeScript representation.

Specifically:
Adds opacity and stroke related fields to `SceneSymbol` and `SceneRect`, along with the general `opacity` field to `SceneItem`.

`cornerRadius` is also added to `SceneRect`.

Before merging it is worth discussing:
- Should the fill fields and stroke fields be abstracted to `FilledSceneItem` and `StrokedSceneItem` types, or should the fields be inlined as they are currently?
- Which of these fields should be optional?

### DIFF
--- a/packages/vega-typings/types/runtime/scene.d.ts
+++ b/packages/vega-typings/types/runtime/scene.d.ts
@@ -50,6 +50,7 @@ export interface SceneItem {
   };
   x: number;
   y: number;
+  opacity?: number;
 }
 
 export type SceneAxis = SceneItem & {
@@ -58,12 +59,16 @@ export type SceneAxis = SceneItem & {
 
 export type SceneRect = SceneItem & {
   fill: string;
+  fillOpacity?: number;
+  stroke?: string;
+  strokeWidth?: number;
+  strokeOpacity?: number;
   height: number;
   width: number;
+  cornerRadius?: number;
 };
 
 export type SceneLine = SceneItem & {
-  opacity: number;
   stroke: string;
   strokeWidth: number;
   x2: number;
@@ -84,9 +89,12 @@ export type SceneGroup = SceneItem & {
 
 export type SceneSymbol = SceneItem & {
   fill: string;
+  fillOpacity: number;
   shape: string;
   size: number;
+  stroke: string;
   strokeWidth: number;
+  strokeOpacity: number;
 };
 
 export type SceneTextBaseline = 'top' | 'middle' | 'bottom';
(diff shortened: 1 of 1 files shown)


packages/vega-typings/types/runtime/scene.d.ts
```diff
@@ -84,9 +89,12 @@ export type SceneGroup = SceneItem & {
 
 export type SceneSymbol = SceneItem & {
   fill: string;
+  fillOpacity: number;
```
domoritz, line 92: A lot of these properties are optional in SceneRect. Are they optional here as well?

domoritz:
Looks reasonable to me. I would not add more classes for now to reduce the complexity of the types. 

@jheer could you take a brief look to let us know whether the properties should be optional or required?

packages/vega-typings/types/runtime/scene.d.ts
lsh: I wasn't intending to make any assumptions about which fields should actually be optional yet. That being said, I'd assume `SceneRect` and `SceneSymbol` would match, regardless of optional or not.

domoritz: It's probably best to err on the side of optional since it's less strict.

lsh: In this case, I would argue it would make sense then for `strokeWidth` to also become optional, since it would be strange for `stroke` and `strokeOpacity` to be optional, but not the width. It would also match with the changes proposed to `SceneRect`.

packages/vega-typings/types/runtime/scene.d.ts
```diff
@@ -58,12 +59,16 @@ export type SceneAxis = SceneItem & {
...
+  stroke?: string;
+  strokeWidth?: number;
+  strokeOpacity?: number;
   height: number;
   width: number;
+  cornerRadius?: number;
 };
 
 export type SceneLine = SceneItem & {
-  opacity: number;
```
domoritz, line 66: Does a line not have opacity?

jheer:
In case it helps, the scenegraph has an associated JSON schema, which you can see here: https://cdn.jsdelivr.net/npm/vega-scenegraph@4.10.1/build/vega-scenegraph-schema.json

The JSON schema is specific to JSON serialization (and so the Vega runtime is actually looser, e.g., in terms of which properties can be added to mark items), but nevertheless documents the properties that the current renderers might access. In any case, none of the item properties are strictly required, so I'd lean towards making all item properties optional in the typings.

domoritz:
Published as vega-typings@0.22.3.

PR #3480, opened 2022-04-28
Add `view` style for Vega-Lite

arvind:
Vega-Lite styles group marks as `cell` only if they define cartesian plots (vega/vega-lite#7665). For other types of group marks (namely, projections), we need another style in order to continue setting `transparent` fills on group marks to capture events.


domoritz (approved):

domoritz:
Adding the blocking Vega-Lite label.

PR #3496, opened 2022-05-18
Change `amount` to `date` in example JSON for timeunit transforms

mistidoi:
Fixes two simples mistakes in timeunit transform documentation.


jheer:
Thanks!

PR #3520, opened 2022-06-21
feat: use @types/geojson for better GeoJSON types

lsh:
Right now we have:
```typescript
export type GeoJsonFeature = any; // TODO
export type GeoJsonFeatureCollection = any; // TODO
```
Since DefinitelyTyped already has a repository for GeoJSON types, it would give more accurate type inference to add it as a package.


packages/vega-typings/package.json
```diff
@@ -19,8 +19,12 @@
     "test": "yarn format && dtslint . --localTs ../../node_modules/typescript/lib"
   },
   "dependencies": {
+    "@types/geojson": "^7946.0.8",
     "vega-event-selector": "^3.0.0",
     "vega-expression": "^5.0.0",
     "vega-util": "^1.15.2"
+  },
+  "devDependencies": {
+    "@types/geojson": "^7946.0.8"
```
domoritz, line 28: We don't need to list this again.

domoritz (approved):

kanitw (approved):

domoritz:
Unfortunately, this is breaking the schema generator. @lsh do you have cycles to help fix it? See https://github.com/vega/ts-json-schema-generator/commit/2512d3a50d642059a2205f7fc6035153775b54d9

Issue #3562, opened 2022-08-26
vega-typings for scenegraph


jonathanzong:
Current:

In `SceneGroup`, the `items` array is typed as `SceneItem[]`.

```
items: SceneItem[];
```

https://github.com/vega/vega/blob/825bfaba6ccfe3306183df22b8c955a07bb30714/packages/vega-typings/types/runtime/scene.d.ts#L79

Expected:

`items` array should be typed as `Scene[]`.

```
items: Scene[];
```

Rationale:

1. It's implied by this test

https://github.com/vega/vega/blob/825bfaba6ccfe3306183df22b8c955a07bb30714/packages/vega-typings/tests/runtime/renderer.ts#L20-L32

2. making this change makes some scenegraph parsing in [mitvis/olli](https://github.com/mitvis/olli) typecheck correctly

example:

https://vega.github.io/editor/#/examples/vega-lite/trellis_barley

```
const a = VEGA_DEBUG.view.scenegraph().root.items[0]; // a is a SceneGroup
a.items.find((el) => el.role === "scope").items // if a.items were a SceneItem[], the last .items wouldn't be there
```

I'm too scared of the runtime code / don't know how to run the tests to make this change myself, but I think it should be easy for someone to verify?

PR #3563, opened 2022-08-26
Update SceneGroup typing in scene.d.ts

jonathanzong:
Explanation for why I think this is right: https://github.com/vega/vega/issues/3562

It would be great if someone could verify that this is correct / doesn't break anything : )

### DIFF
--- a/packages/vega-typings/types/runtime/scene.d.ts
+++ b/packages/vega-typings/types/runtime/scene.d.ts
@@ -81,7 +81,7 @@ export interface SceneContext {
 
 export type SceneGroup = SceneItem & {
   context: SceneContext;
-  items: SceneItem[];
+  items: Scene[];
   height: number;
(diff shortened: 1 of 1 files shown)


jonathanzong:
tagging @arvind for review

domoritz:
@danmarshall any opinions here?

danmarshall:
I'm not certain, but perhaps the [sceneVisit](https://github.com/vega/vega/blob/main/packages/vega-typings/types/runtime/scene.d.ts#L132) call might give us a hint:
```ts
 itemCallback: (item: Scene | SceneGroup | SceneItem) => void
```
Perhaps we should rename the current SceneItem to SceneMark, since the uses are for SceneLine, SceneRect, SceneText etc.
Then the new SceneItem would be `Scene | SceneGroup | SceneMark`.

PR #3570, opened 2022-09-10
Interpreter: Prevent evaluation of unsafe methods

AMoo-Miki:
While in a CSP-compliant environment, unsafe evals are blocked by the browser. However, in an unsafe environment, blocking these methods from being interpreted would prevent their evaluation.


joshuarrrr:
@jheer Is there any other context or information we can provide to help expedite the review of this PR?

jheer:
Thanks!

PR #3577, opened 2022-09-16
feat: add 95th and 99th percentile to aggregation

artiya4u:
The 95th and 99th percentile are crucial for some applications e.g. network latency visualization.
I proposed new aggregation operations `p95` and `p99` to calculate the 95th, and 99th percentile respectively.
By using the well-defined `quantiles()` function, used by `q1` and `q3` operations.
I modify it to add the `p95`  and `p99` without making a new function.


domoritz:
Related issue: https://github.com/vega/vega-lite/issues/4439

Can I propose to add a parameter to set the quantile and to have an ntile aggregate instead? 

Note that Vega already has a quartile transform: https://vega.github.io/vega/docs/transforms/quantile/.

artiya4u:
@domoritz 
> vega/vega-lite#4439

I came from the issue, I just think my proposed `p95`, `p99` operations are just faster to implement and straightforward to use for an average user. It cover most of my use cases(network latency, benchmark result). I would love my solution to share with everyone who has the same problem. The Ntiles with parameter solution might be too hard for me since no one wants to work on it for a year.

domoritz:
Fair points. Please remove the compiled files and also update the ts types so @jheer can do a review.

artiya4u:
OK, I think it is ready to review.

packages/vega/test/specs-valid/window.vg.json
```diff
@@ -16,7 +16,7 @@
       "name": "aggregate", "value": "sum",
       "bind": {
         "input": "select",
-        "options": ["sum", "count", "average", "median", "stdev"]
+        "options": ["sum", "count", "average", "median", "stdev", "p95", "p99"]
```
domoritz, line 19: should this have q1 and q3 as well?

domoritz (approved):
Thank you. Looks good. I still think it would be nice to have a general quantile method.

domoritz:
aggregate_params addressed this and is on main already.

PR #3607, opened 2022-10-31
feat: allow slider input element to show max/min boundary value

caok2709:
1. allow range input element to show max/min as a boundary. 
2. When you move slider close to max/right value. It should looks like: 450/1200 -> 500/1200 -> 600/1200
3. When you move slider close to min/left value. It should looks like: -450/-1200 -> -500/-1200 -> -600/-1200


caok2709:
@hadley @cjiang @Fil @gvangool  
Hello. 
I found this PR is useful in our project.
When you change the slider value, it will keep the max or min value as a boundary to indicate the number gap between current value to max/min value.
Let me know if you can merge and publish it as a new version. Thx

PR #3614, opened 2022-11-17
Fix Slack community link on main docs page

mofojed:
The old link (http://bit.ly/vega-slack) just opens a new page to https://communityinviter.com/, instead of to the Vega Slack community.
I think the correct link is (https://bit.ly/join-vega-slack-2022), which I got from the Vega Editor help: https://vega.github.io/editor/#/edited.


jheer:
Thank you @mofojed!

PR #3615, opened 2022-11-17
Fix the "Edit this Page" link

mofojed:
- The link pointed to the `master` branch, which does not exist in the vega repo so resulted in a 404
- Update the link to point to `main` branch instead
- There are other references in the repo (such as in docs/docs/api/time.md) to the `master` branch, but GitHub seems to redirect those correctly so I did not update those


jheer:
Thanks again @mofojed!

Issue #1310, opened 2018-06-05
Band support for the time-based scale


nyurik:
A very common graph is a bar chart with the **temporal** X axis. Yet, Vega has relatively little support for that usecase.  Graph designers must hardcode the bar width, like the [wheat & wages](https://vega.github.io/vega/examples/wheat-and-wages/) example, and know ahead of time the "bar unit" of dates, e.g. monthly vs daily. This goes against the whole idea of the data driver graph'.

## Goals
* time scale should calculate the smallest gap size between time elements
* time scale should provide band width to the rect mark, similar to the band scale (consistent, simpler for novice users)
* the band size should account for any padding/rounding that time scale adds
* it should be possible for the user to add in-between bar padding, same as in band scale

Note that there was a [similar request in 2013](https://github.com/vega/vega/issues/62), but there was no example for it, so there might be a workaround.

CC: @domoritz @dnmiller @gerwitz @mkscrg

kanitw:
Time scale is a continuous scale, so I don't think extending it to serve both continuous and discrete use case is a good design

You can actually already do this with a band scale (with some special formatting for the axis labels).

In fact, you can already do this in Vega-Lite if your data has no missing time values. 
(If you don't then you have to impute your data, then it will work.  Note that VL doesn't have impute yet, but it will soon. You can already impute data in Vega though.) 

Here is an example Vega-Lite spec:

```
{
  "$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "data": {"url": "data/seattle-weather.csv"},
  "width": 200,
  "mark": "bar",
  "encoding": {
    "x": {
      "timeUnit": "yearmonth",
      "field": "date",
      "type": "ordinal"
    },
    "y": {
      "aggregate": "mean",
      "field": "precipitation",
      "type": "quantitative"
    }
  }
}
```

Feel free to use the editor to see the full Vega spec.

nyurik:
@kanitw your example highlights the issue I was talking about -- you have to "know" your data range when designing the graph -- by specifying `yearmonth`. You cannot design a truly data-driven graph, where the data decides the time units - months or minutes, and the graph can change on the fly depending on the data given.

The bands here are not discrete, but are also continuous. Bars do not have to align - in fact bars could appear at any place anywhere on the scale. It is imputing that makes them descrete.  Using minimum gap as the bar width will make both the continuous and "pre-descretized" (data generated on the server as buckets) data work the same way.

nyurik:
[...]

PR #3645, opened 2023-01-12
feat: add "auto" unit to timeunit transform

paulrosenzweig:
This pull request adds a new infer unit option to the TimeUnit transform. When set to infer units, the transform will automatically choose the appropriate time unit based on the input data.

I wanted to try and solve the problem discussed in #1310: plotting discretized timeseries data without knowing the grain.

Here's [a demo Observable notebook](https://observablehq.com/d/5d869cd32fb125a4). It patches the timeunit transform and uses DuckDB to produce differently grouped time data.

## Problem/Solution
Currently, you can discretize data into months using this transform. 
```json
{"type": "timeunit", "field": "date", "units": ["year", "month"]}
```

But, if your data is _already_ grouped by month, there's no way to infer that from the data.
This PR adds an "auto" unit option that would work with this example data.

```json
[{"date": "2023-01-01"}, {"date": "2023-02-01"}, {"date": "2023-03-01"}]
```

```json
{"type": "timeunit", "field": "date", "inferUnits": true}
```

## Alternate approaches

In [an issue comment](https://github.com/vega/vega/issues/1310#issuecomment-396734615), @jheer suggested creating a new discretizing time scale that acts like a band scale. I believe that discussion happened before the TimeUnit transform was added. Enhancing TimeUnit seemed more natural fit than adding a scale, but I'd love to hear any other thoughts! I meant this PR mostly as a discussion starter with some running code. 

## Questions
- Is the API right?
~I'm uncertain about the `"units": ["auto"]` parameter. Would a separate param be better? e.g. `"inferUnits": true` or similar?~
I updated this to `inferUnits: true`. Very open to feedback on the naming here though!
- Should I pull in the dependency?
I created a small npm package that does the work to determine the likely unit/step. It felt like a separate concern that's not specific to vega, but it's very small. Should I pull that into vega-time as part of this PR?
The main reason to leave it separate IMO is to add tests that would be annoying to have in Vega. I want to run the tests with Node booted into different timezones, but that's annoying requirement to put on a bigger project.

### DIFF
--- a/packages/vega-transforms/src/TimeUnit.js
+++ b/packages/vega-transforms/src/TimeUnit.js
@@ -1,7 +1,7 @@
 import {Transform} from 'vega-dataflow';
 import {
   TIME_UNITS, timeBin, timeFloor, timeInterval, timeUnits,
-  utcFloor, utcInterval
+  utcFloor, utcInterval, detectTimeUnits
 } from 'vega-time';
 import {accessorFields, extent, inherits, peek} from 'vega-util';
 
@@ -27,6 +27,7 @@ TimeUnit.Definition = {
     { 'name': 'step', 'type': 'number', 'default': 1 },
     { 'name': 'maxbins', 'type': 'number', 'default': 40 },
     { 'name': 'extent', 'type': 'date', 'array': true},
+    { 'name': 'inferUnits', 'type': 'boolean', 'default': false },
     { 'name': 'timezone', 'type': 'enum', 'default': 'local', 'values': ['local', 'utc'] },
     { 'name': 'as', 'type': 'string', 'array': true, 'length': 2, 'default': OUTPUT }
   ]
@@ -35,18 +36,18 @@ TimeUnit.Definition = {
 inherits(TimeUnit, Transform, {
   transform(_, pulse) {
     const field = _.field,
-          band = _.interval !== false,
-          utc = _.timezone === 'utc',
-          floor = this._floor(_, pulse),
-          offset = (utc ? utcInterval : timeInterval)(floor.unit).offset,
-          as = _.as || OUTPUT,
-          u0 = as[0],
-          u1 = as[1],
-          step = floor.step;
+      band = _.interval !== false,
+      utc = _.timezone === 'utc',
+      floor = this._floor(_, pulse),
+      offset = (utc ? utcInterval : timeInterval)(floor.unit).offset,
+      as = _.as || OUTPUT,
+      u0 = as[0],
+      u1 = as[1],
+      step = floor.step;
 
     let min = floor.start || Infinity,
-        max = floor.stop || -Infinity,
-        flag = pulse.ADD;
+      max = floor.stop || -Infinity,
+      flag = pulse.ADD;
 
     if (
       _.modified() ||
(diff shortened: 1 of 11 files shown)


domoritz (commented):
Thanks for your patience with this and thank you for the pull request.

I love the idea of an auto time unit that's based on the grain of the data. @hamilton has written a related DuckDB query in https://www.hamiltonulmer.com/smallest-timegrain-heuristic/. Having the transform instead of a new scale type is nice because we can easily do aggregation after the time unit (which is commonly what people do after time unit transforms). 

@PBI-David @nyuri could you see whether this addresses your use cases? 

For API, I feel that having another option "auto" in an array is odd since that allows for `units: ["auto", "hour"]`. I'd prefer either an array or "auto". 

A separate package with tests makes sense here. 

@kanitw I think otherwise this looks like a great addition and I'm in favor of accepting it.

PBI-David:
@domoritz I appreciate the new functionality but I don't see how it helps my use case unless I'm missing something obvious. I needed to create a Gantt chart and for this, I needed to treat a continuous temporal scale as if it was ordinal/discrete, including the ability to zoom in and out to change the grain (so both continuous **and** discrete at the same time). I managed to get it working using existing functionality but it was a lot of work and the ability to treat a continuous temporal scale as a band scale would have helped immensely. 


[Editor](https://evyb.short.gy/nCRa3r)

Gantt 2 0 Thumbnail

For me, it comes down to this. Supply two dates or datetimes and have the chart be rendered correctly with the ability to have centered labels in the middle of the band which is very similar to what @kanitw was proposing here: https://github.com/vega/vega-lite/issues/4868

In the image below, I have had to change the end date to +1 day to get the label in the correct position (other attempts left you with not enough ticks or gaps at the start or end of the chart, even using tickExtra) and then manually change the labels and ticks to get them to show what I want. 
image

domoritz:
That makes sense and thanks for chiming in here. Regardless it still seems to make sense to make this change even if we introduce another scale type or improvements to scales/axes in the future, right?

PBI-David:
> That makes sense and thanks for chiming in here. Regardless it still seems to make sense to make this change even if we introduce another scale type or improvements to scales/axes in the future, right?

I agree that this should be decoupled from the separate request of implementing another scale and the improvements to axes. Whether it makes sense to add, I can't say as I don't feel like I understand well enough the use cases it would solve. It _seems_ to make sense on the surface but I suspect that @nyurik  is much better placed to comment.

paulrosenzweig:
Thanks for taking a look!

> For API, I feel that having another option "auto" in an array is odd since that allows for units: ["auto", "hour"]. I'd prefer either an array or "auto".

Agreed. Do you prefer `"units": "auto"` or a separate parameter like `"inferUnits": true`?

I can make that change and update this branch to have the latest from `main`.

paulrosenzweig:
I went with `"inferUnits": true`. I didn't see a way to express `array | "auto"` in the Vega transform param definitions.

Let me know if any other changes would be helpful!

rollup.config.mjs
domoritz, line 1: Please undo these changes

packages/vega-transforms/src/TimeUnit.js
```diff
@@ -18,35 +18,47 @@ export default function TimeUnit(params) {
...
-    { 'name': 'field', 'type': 'field', 'required': true },
-    { 'name': 'interval', 'type': 'boolean', 'default': true },
-    { 'name': 'units', 'type': 'enum', 'values': TIME_UNITS, 'array': true },
-    { 'name': 'step', 'type': 'number', 'default': 1 },
-    { 'name': 'maxbins', 'type': 'number', 'default': 40 },
-    { 'name': 'extent', 'type': 'date', 'array': true},
-    { 'name': 'timezone', 'type': 'enum', 'default': 'local', 'values': ['local', 'utc'] },
-    { 'name': 'as', 'type': 'string', 'array': true, 'length': 2, 'default': OUTPUT }
-  ]
+  "type": "TimeUnit",
```
domoritz, line 21: let's reduce changes to the formatting (like quotes here)

domoritz (approved):
Thanks for your patience on this. Let's get it in! 

I like the API and it looks good overall. I think I changed my mind about where the code should live and think it would be good to move the time-grain-detector here so we have the code and tests local. But don't feel strongly about it. 

Could you update the pull request to use the latest main and reduce the formatting changes (see a few comments here).

paulrosenzweig:
I brought this up to current main and removed the unnecessary diff. Looks like it needs a maintainer to approve running the Node 18 workflow.

> I think I changed my mind about where the code should live and think it would be good to move the time-grain-detector here so we have the code and tests local. But don't feel strongly about it.

Agreed! I'll pull that in when I get a chance. There's really only one test that loops over different examples, so it shouldn't be too hard to port.

domoritz:
I approved the tests. Let's get the library moved over and I'll merge. This is great!

domoritz:
@paulrosenzweig would you have a moment to wrap this up so we can merge?

PR #3659, opened 2023-02-10
Fix: Make `__count__` private from schema

lsh:
Currently `__count__` appears in the aggregate ops in the schema. This should be hidden for internal use in `Pivot` operations (confirmed by @jheer over Slack).


lsh:
Should `__count__` also be removed from `vega-typings` aggregate ops as part of this? cc @domoritz @jheer

domoritz:
The types should be in sync with the schema imo.

domoritz:
Please remove the compiled Vega files. We make those updates in a release.

domoritz (approved):

domoritz:
Thank you. Looks good to me. @jheer can you merge this?

PR #3675, opened 2023-03-03
feat: add field predicate types to selectionTest in vega-selections

jonathanzong:
This change introduces the ability for `vlSelectionTest` to check for more types of [field predicates](https://vega.github.io/vega-lite/docs/predicate.html).

This additive change will have no immediate effect, since Vega-Lite doesn't currently produce selection tuples that use these predicate types. However, future proposed changes for Animated Vega-Lite will require us to support these field predicates in addition to the existing point and interval predicates.

cc @arvind


jonathanzong:
i believe this supersedes https://github.com/vega/vega/pull/1839 which is also very stale anyway

packages/vega-selections/src/selectionTest.js
```diff
@@ -41,6 +46,16 @@ function testPoint(datum, entry) {
         if (!inrange(dval, values[i], false, false)) return false;
       } else if (f.type === TYPE_RANGE_LE) {
         if (!inrange(dval, values[i], false, true)) return false;
+      } else if (f.type === TYPE_PRED_LT) {
```
domoritz, line 49: Should this become a switch now?

jonathanzong: i personally don't find switch to be clearer in this instance because all of the returns are conditional, so you'd have to add break statements to every one and remember to keep adding them if we change it in the future. IMO, not shorter, possibly less maintainable, not a significant change in readability

domoritz: Makes sense. I don't have a strong opinion. You could pull the logic here out into a function and then use a return in each branch.

jonathanzong: the if statements only return false because it's an early exit from the loop (none of the comparisons can be false for the test to return true), so we can't use an unconditional return inside the switch without changing the loop too

presumably it is written this way for efficiency(tm) over readability? but it did take me trying to rewrite it to realize this so maybe it should just be made more readable?

what do you think of e.g.
```
fields.every(f => {
  switch (f.type) {
    case TYPE_ENUM:
      return ...;
    case ...
  }
})
```

arvind: Your guess is correct @jonathanzong! Because this code path is typically called against every visualized data point on every interaction (eg potentially every mouse drag), we need it to be really performant. When this was first written, most browsers were still faster at evaluating an old school for loop. It was also one of the reasons we minimized functions calls (because they can—or at least could—incur performance overhead).

So, while I’m sympathetic to readability concerns, if we really wanted to make a change here (including extracting this to another function call), perhaps we could run a micro benchmark to validate that it would cause a performance hit? 

(My guess is browsers have advanced significantly since these code paths were first written, so I wouldn’t be surprised if some of these reasons no longer hold. On the other hand, back in the day, I do remember being surprised  by what syntaxes did and didn’t incur performance penalties!)

jonathanzong:
bump @arvind

domoritz (approved):
Looks good to me but I will defer to @arvind for the final review.

domoritz:
@arvind can you wrap up the review and merge if it looks good?

jonathanzong:
I've found an issue with this, which is that the [ops in `selectionResolve`](https://github.com/vega/vega/blob/11a33fbc0de9b9b45c91658708bd96ee9b00b8d6/packages/vega-selections/src/selectionResolve.js#L89) do a lookup that's hardcoded to the name of the `type`, so introducing new `type`s will mess with this. I can fix this later, but let's un-approve this until then

jonathanzong:
I found a potential problem here: [ops in `selectionResolve`](https://github.com/vega/vega/blob/11a33fbc0de9b9b45c91658708bd96ee9b00b8d6/packages/vega-selections/src/selectionResolve.js#L89) do a lookup with the first letter of the type name, which is assumed to be E or R. Since we are introducing new types here the lookup seem like they will fail if you try to use them.

I'm pretty confused but can look into this later but let's not merge this yet until then.

domoritz:
I converted this pull request to a draft. Please mark it as ready.

jonathanzong:
this is ready for review now

arvind (changes requested):
Hi @jonathanzong, so very sorry for lagging behind on approving this (to put it mildly). The additional predicate types and expansion of the if-block look good to me. But, I'm not sure whether the `E` prefix is what we want. 

To clarify, the `selectionResolve` code path gets called when a selection is defined within a faceted or repeated view (i.e., there is an instance of the selection for each facet/repeated unit and we need a way of resolving them to produce an overall predicate that is applied to the other views). 

The type here is not about the predicate _comparator_ but rather the _terms_ of the predicate (what the Vega-Lite codebase names [`TupleStoreType`](https://github.com/vega/vega-lite/blob/main/src/compile/selection/project.ts#L23)). Essentially, are the predicates enumerated values (typically generated by point selections or brushes over cartographic projections) or express a two tuple range (typically generated by brushes or point selections over binned scales). 

So, I'm not sure that using the `E` (for enum) prefix across all the new predicate types is what we want? I think it makes sense for `VALID` and `ONE_OF` since I expect those probably operate a list of discrete values. But, for the inequalities, I would imagine they should be `R`? 

Relatedly, have we implemented tests somewhere in the stack to validate these functions? (Thus far, I think these functions have been tested in the Vega-Lite codebase.) I just want to make sure since I'd be nervous if parts of this PR are designed speculatively.

jonathanzong:
I believe that `E` is the appropriate prefix for even the inequalities. `R` is used for interval selections where the predicate takes in a ranged term, but the inequalities are defined by a single value. I've demonstrated this in the new unit tests I've added in [`predicate-test.js`](https://github.com/vega/vega/blob/0474cda2c6d3f563905165fbec89c290704788b5/packages/vega-selections/test/predicate-test.js). 

Example:

"a is between 55 and 160"
```[{unit: '', fields:[{field: 'a', channel: 'x', type: 'R'}], values: [[55,160]]}]```

vs

"a is greater than 100"
```[{unit: '', fields:[{field: 'a', channel: 'x', type: 'E-GT'}], values: [100]}]```

I think this is consistent with the point/interval metaphor because you could imagine clicking a point and intending to select e.g. "everything to the right of this point on the x axis".

So far, I've been testing these in Umwelt and Animated Vega-Lite at runtime. I don't think I'm at a point where I could add runtime tests to Vega-Lite for this, since there isn't currently anything in Vega-Lite that generates these predicates. But hopefully the unit tests help us confirm this functionality works as intended and helps us move confidently once we get to the point where we're enabling more types of predicates.

arvind:
Aah this makes sense to me. Thanks for explaining it and for adding `predicate-test.js`—having those around makes me feel much more confident (and is something I should've done when I first added the predicate tests!).

arvind (approved):

PR #3676, opened 2023-03-03
feat(vega-force): Enable expr support for nbody 'strength' param

handsfelloff:
Improve n-body force flexibility by enabling support for per-node _expr_ control of the 'strength' parameter.


jheer:
Thanks for the PR! A few notes:

- The `theta` parameter is a global simulation setting and must be a single number. Making this parameter a per-node expression is not valid, so we need to revert those changes.
- The `strength` parameter can be function-valued, so the change should be fine. However, the TypeScript typings in the vega-typings package will also need to be updated (https://github.com/vega/vega/blob/main/packages/vega-typings/types/spec/transform.d.ts#L331)

handsfelloff:
Thanks @jheer for the swift feedback!  I have reverted the theta expr and added the interface strength ExprRef typing

jheer:
Thanks!

Issue #3341, opened 2021-11-21
Add exponential moving average as an aggregation function.


Xitian9:
One type of aggregation which is not currently supported is the [exponential moving average](https://en.wikipedia.org/wiki/Moving_average#Exponential_moving_average). Unlike the simple moving average (i.e. mean over a window), the exponential moving average weights recent values more heavily, with an exponentially decaying weighting dying off towards infinity. This forms a smooth alternative to a simple moving average, and is commonly used when you want to more heavily weight recent values, but don't want a discontinuous drop-off when numbers drop out of an averaging window.

For any value `0 < r < 1`, we can construct a vector of weights:
```
[(1-r), (1-r)*r, (1-r)*r^2, (1-r)*r^3, ...]
```
The inner product of the data with this weight vector is then taken, with more recent (i.e. later in the vector) values more highly weighted.

I would propose a new aggregation type, which could be specified as follows:
```
aggregate: { "exponential": 0.5 }
```
Alternatives to `exponential` include `ema`, `expma`, `expmovingaverage`, `exponentialmovingaverage`, or similar.

A possible variation, which we may or may not want to support, is the notion of a biased/unbiased exponential moving average. The weight vector above sums to 1 only when there are infinitely many elements in the vector, and otherwise will undercount the average. This can be corrected for by multiplying the weight vector by `1 / (1 - r^n)`, where `n` is the length of the data vector. The magnitude of the bias is relatively small for reasonably large values of `n` and most values of `r`. The biased exponential moving average is equivalent to the unbiased exponential moving average if you assume that all values before the start of the vector are zero.

A possible interface would be:
```
aggregate: { "exponential": 0.5, "unbiased": true }
```
where unbiased defaults to true.

Xitian9:
Is there any interest in implementing this? I've tried poking around myself, but as I'm not a Javascript programmer I ran into some roadblocks. The main one I see:

1. Can we provide a parameterised aggregation function, i.e. does the existing interface allow us to provide a value for r.

Let me know how I can help.

Xitian9:
Recording this here as a previous proposal for adding aggregates which take parameters. The comments are probably relevant for this. https://github.com/vega/vega-lite/issues/4439.

hydrosquall:
I think we can close this as fixed in #3686

PR #3686, opened 2023-03-12
feat: Add aggregate parameters to vega-transform, and exponential moving average

Xitian9:
This adds exponential moving averages, fixing #3341. In order to do this, it enables supplying parameters to aggregation functions. This addresses the block on https://github.com/vega/vega-lite/issues/4439.

As I'm not a javascript programmer, feedback will be gratefully taken on board.

### DIFF
--- a/packages/vega-transforms/src/Aggregate.js
+++ b/packages/vega-transforms/src/Aggregate.js
@@ -11,6 +11,7 @@ import {accessorFields, accessorName, array, error, inherits} from 'vega-util';
  * @param {Array<function(object): *>} [params.groupby] - An array of accessors to groupby.
  * @param {Array<function(object): *>} [params.fields] - An array of accessors to aggregate.
  * @param {Array<string>} [params.ops] - An array of strings indicating aggregation operations.
+ * @param {Array<object>} [params.aggregate_params=[null]] - An optional array of parameters for aggregation operations.
  * @param {Array<string>} [params.as] - An array of output field names for aggregated values.
  * @param {boolean} [params.cross=false] - A flag indicating that the full
  *   cross-product of groupby values should be generated, including empty cells.
@@ -45,6 +46,7 @@ Aggregate.Definition = {
   'params': [
     { 'name': 'groupby', 'type': 'field', 'array': true },
     { 'name': 'ops', 'type': 'enum', 'array': true, 'values': ValidAggregateOps },
+    { 'name': 'aggregate_params', 'type': 'field', 'null': true, 'array': true, 'default': [null] },
     { 'name': 'fields', 'type': 'field', 'null': true, 'array': true },
     { 'name': 'as', 'type': 'string', 'null': true, 'array': true },
     { 'name': 'drop', 'type': 'boolean', 'default': true },
@@ -161,10 +163,11 @@ inherits(Aggregate, Transform, {
 
     const fields = _.fields || [null],
           ops = _.ops || ['count'],
+          aggregate_params = _.aggregate_params || [null],
           as = _.as || [],
           n = fields.length,
           map = {};
-    let field, op, m, mname, outname, i;
+    let field, op, aggregate_param, m, mname, outname, i;
 
     if (n !== ops.length) {
       error('Unmatched number of fields and aggregate ops.');
@@ -173,6 +176,7 @@ inherits(Aggregate, Transform, {
     for (i=0; i<n; ++i) {
       field = fields[i];
       op = ops[i];
+      aggregate_param = aggregate_params[i] || null;
 
       if (field == null && op !== 'count') {
         error('Null aggregate field specified.');
@@ -195,7 +199,7 @@ inherits(Aggregate, Transform, {
       }
 
       if (op !== 'count') this._countOnly = false;
-      m.push(createMeasure(op, outname));
+      m.push(createMeasure(op, aggregate_param, outname));
     }
 
     this._measures = this._measures.map(m => compileMeasures(m, m.field));
--- a/packages/vega-transforms/src/Window.js
+++ b/packages/vega-transforms/src/Window.js
@@ -17,6 +17,7 @@ import {bisector} from 'd3-array';
  * @param {Array<function(object): *>} [params.fields] - An array of accessors
  *   for data fields to use as inputs to window operations.
  * @param {Array<*>} [params.params] - An array of parameter values for window operations.
+ * @param {Array<object>} [params.aggregate_params] - An optional array of parameter values for aggregation operations.
  * @param {Array<string>} [params.as] - An array of output field names for window operations.
  * @param {Array<number>} [params.frame] - Window frame definition as two-element array.
  * @param {boolean} [params.ignorePeers=false] - If true, base window frame boundaries on row
@@ -37,6 +38,7 @@ Window.Definition = {
     { 'name': 'groupby', 'type': 'field', 'array': true },
     { 'name': 'ops', 'type': 'enum', 'array': true, 'values': ValidWindowOps.concat(ValidAggregateOps) },
     { 'name': 'params', 'type': 'number', 'null': true, 'array': true },
+    { 'name': 'aggregate_params', 'type': 'field', 'null': true, 'array': true, 'default': [null] },
     { 'name': 'fields', 'type': 'field', 'null': true, 'array': true },
     { 'name': 'as', 'type': 'string', 'null': true, 'array': true },
     { 'name': 'frame', 'type': 'number', 'null': true, 'array': true, 'length': 2, 'default': [null, 0] },
--- a/packages/vega-transforms/src/util/AggregateOps.js
+++ b/packages/vega-transforms/src/util/AggregateOps.js
@@ -127,6 +127,16 @@ export const AggregateOps = {
     add:  (m, v, t) => { if (v > m.max) m.argmax = t; },
     rem:  (m, v) => { if (v >= m.max) m.argmax = undefined; },
     req:  ['max', 'values'], idx: 3
+  },
+  exponential: {
+    init: (m, r) => { m.exp = 0; m.exp_r = r; },
+    value: m => m.valid ? (m.exp * (1 - m.exp_r) / (1 - m.exp_r ** m.valid)) : undefined,
+    add:  (m, v) => m.exp = m.exp_r * m.exp + v,
+    rem:  (m, v) => m.exp = (m.exp - v / m.exp_r ** (m.valid - 1)) / m.exp_r
+  },
+  exponentialb: {
+    value: m => m.valid ? (m.exp * (1 - m.exp_r)) : undefined,
+    req:  ['exponential'], idx: 1
   }
 };
 
--- a/packages/vega-transforms/src/util/AggregateOps.js
+++ b/packages/vega-transforms/src/util/AggregateOps.js
@@ -133,8 +133,9 @@ export const AggregateOps = {
 export const ValidAggregateOps = Object.keys(AggregateOps).filter(d => d !== '__count__');
 
 function measure(key, value) {
-  return out => extend({
+  return (out, aggregate_param) => extend({
     name: key,
+    aggregate_param: aggregate_param,
     out: out || key
   }, base_op, value);
 }
@@ -143,8 +144,8 @@ function measure(key, value) {
   AggregateOps[key] = measure(key, AggregateOps[key]);
 });
 
-export function createMeasure(op, name) {
-  return AggregateOps[op](name);
+export function createMeasure(op, param, name) {
+  return AggregateOps[op](name, param);
 }
 
 function compareIndex(a, b) {
@@ -169,7 +170,7 @@ function resolve(agg) {
 function init() {
   this.valid = 0;
   this.missing = 0;
-  this._ops.forEach(op => op.init(this));
+  this._ops.forEach(op => (op.aggregate_param == null) ? op.init(this) : op.init(this, op.aggregate_param));
 }
 
 function add(v, t) {
--- a/packages/vega-transforms/src/util/WindowState.js
+++ b/packages/vega-transforms/src/util/WindowState.js
@@ -7,6 +7,7 @@ export default function WindowState(_) {
   const ops = array(_.ops),
         fields = array(_.fields),
         params = array(_.params),
+        aggregate_params = array(_.aggregate_params),
         as = array(_.as),
         outputs = this.outputs = [],
         windows = this.windows = [],
@@ -24,6 +25,8 @@ export default function WindowState(_) {
 
   ops.forEach((op, i) => {
     const field = fields[i],
+          param = params[i],
+          aggregate_param = aggregate_params[i] || null,
           mname = accessorName(field),
           name = measureName(op, mname, as[i]);
 
@@ -32,7 +35,7 @@ export default function WindowState(_) {
 
     // Window operation
     if (hasOwnProperty(WindowOps, op)) {
-      windows.push(WindowOp(op, fields[i], params[i], name));
+      windows.push(WindowOp(op, field, param, name));
     }
 
     // Aggregate operation
@@ -52,7 +55,7 @@ export default function WindowState(_) {
         m.field = field;
         measures.push(m);
       }
-      m.push(createMeasure(op, name));
+      m.push(createMeasure(op, aggregate_param, name));
     }
   });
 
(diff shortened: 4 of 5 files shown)


Xitian9:
Is there any interest in this?

domoritz:
Thanks for the pull request. I just approved the test run again.

Xitian9:
If/when this is merged, #3577 could be made more general by passing the requested quantiles as a parameter.

shcheklein:
Hey @Xitian9 thanks for the effort! This is exactly what we are looking for in this case https://github.com/iterative/vscode-dvc/issues/3837 . In ML tracker tools it's very common to use the exponential moving average. I'm still looking into this (and learning along the way if there are other ways to achieve this).

@domoritz are there any blockers left to merge this? (maybe we can also join and help). Also, if this is merged will it become available automatically in vega-lite?

packages/vega-transforms/src/Aggregate.js
```diff
@@ -11,6 +11,7 @@ import {accessorFields, accessorName, array, error, inherits} from 'vega-util';
  * @param {Array<function(object): *>} [params.groupby] - An array of accessors to groupby.
  * @param {Array<function(object): *>} [params.fields] - An array of accessors to aggregate.
  * @param {Array<string>} [params.ops] - An array of strings indicating aggregation operations.
+ * @param {Array<object>} [params.aggregate_params=[null]] - An optional array of parameters for aggregation operations.
```
domoritz, line 14: Do we need to repeat the name `aggregate` here or is it already obvious because of the context of the aggregate?

domoritz (approved):
Looks reasonable to me. The API question I have is whether it makes sense for the parameter to be a separate property or somehow part of the ops. I guess this is consistent with the rest of the API. 

@jheer needs to make a final pass and merge

shcheklein:
Okay, was looking into this one more time and I don't get so far is why this is considered the _moving_ average. It's producing a single value. @Xitian9 could you clarify please the use case for this? can it be used to produce a moving average, or may be I'm missing something.

packages/vega-transforms/src/Aggregate.js
Xitian9: I was going by the naming agreed upon in https://github.com/vega/vega-lite/issues/4439. We cannot just call it params.params, as this would conflict with the params argument in windowing functions. Anytime you would want to apply an aggregate with parameter to over a window with a parameter, you would run into problems.

Xitian9:
> Can it be used to smooth a linear plot or does it calculate a single number?

The aggregate in itself produces a single number (which is why I call it `exponential` instead of `exponentialmovingaverage` or something like that). However, it can easily be combined with a windowing function to produce a smoothing, which is the use case I have in mind as well.

> @domoritz are there any blockers left to merge this? (maybe we can also join and help). Also, if this is merged will it become available automatically in vega-lite?

I don't believe it would automatically be available in vega-lite, and that would need to be updated as well.

shcheklein:
@jheer sorry for bothering you, anything we can do (help with) here to unblock this and move forward? Please let me know.

Xitian9:
Is there any appetite for this? If not I'll delete the PR.

shcheklein:
@Xitian9 please keep it for a while :) We definitely need this! I can't make decisions here and / or review, but please keep it possible. I hope folks will get some time to merge it. Thanks for your work.

shcheklein:
Hey folks, @jheer , any chance we can move it forward? Could we at least identify what is blocking this PR atm?

jheer:
Thanks for the PR, sorry for the delay in reviewing!

I checked with @jonmmease and expanding aggregates this way does not have negative downstream consequences for VegaFusion (other than that weighted exponential aggregates will not be optimized by VegaFusion at the moment).

julieg18:
Hello, I'm working on adding `aggregate_parameters` into vega-lite (https://github.com/vega/vega-lite/issues/9200), but when I tested using the params in a vega editor spec, all I could get was `NaN`:

image

I tried a couple different field variations but `NaN` was always the result. Have I misunderstood what `aggregate_params` is supposed to consist of? Here's the [example vega editor spec](https://vega.github.io/editor/#/url/vega/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykSArJQBWENgDsQAGhAATJJhSoA2qHFIEcNCAhIANnAgB9OAA8NOA9JDF9DQ2lUgyAJzYMcaAAwzdBiGgALJQATF4CMjhILhrelF4JAJwAvlKgru6eqD46+g6owYFeIQBskdGx2fGBiSWp6W4eaACMvnkBqEIhza0gUTEIccUlABz1zo1ZvX75wQDsIwIh5QNDI93jGU2oy7n+aEJLgQDMK5Ve1XMhm5NouzMdx5QlC7v95-FzzQI3md5t+wKlBGIzKfQqgyqXkScxSaQmf1Q03aB2eIUSIzOkIuXm+cIaiORgKExxhgSxQ0CIWu8K2WXuKNQTzmpMSFKhiVJyQAuuN1JptO8EEZbHp7NZRfYOhcfvD+Vp0GYcBI4OJMDR9EYkMQyNYxAwXFAFXtDCZzAhLFoZJgYuIIAAzNguSFOTAATxwxqQZFcfEwVpA9pocD0sg6yhNAW5MjYeEcICVKrVGr0IGjICQ4ZALkMDD0mDTMjpTDd8bphYzPpzFH9RiFWaFIrscEo8kUaZ5naAA) I was using.

packages/vega-transforms/src/Aggregate.js
```diff
@@ -45,6 +46,7 @@ Aggregate.Definition = {
   'params': [
     { 'name': 'groupby', 'type': 'field', 'array': true },
     { 'name': 'ops', 'type': 'enum', 'array': true, 'values': ValidAggregateOps },
+    { 'name': 'aggregate_params', 'type': 'field', 'null': true, 'array': true, 'default': [null] },
```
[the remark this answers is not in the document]
domoritz: @Xitian9

packages/vega-transforms/src/util/AggregateOps.js
```diff
@@ -127,14 +127,25 @@ export const AggregateOps = {
...
     req:  ['max', 'values'], idx: 3
+  },
+  exponential: {
+    init: (m, r) => { m.exp = 0; m.exp_r = r; },
+    value: m => m.valid ? (m.exp * (1 - m.exp_r) / (1 - m.exp_r ** m.valid)) : undefined,
+    add:  (m, v) => m.exp = m.exp_r * m.exp + v,
+    rem:  (m, v) => m.exp = (m.exp - v / m.exp_r ** (m.valid - 1)) / m.exp_r
+  },
+  exponentialb: {
+    value: m => m.valid ? (m.exp * (1 - m.exp_r)) : undefined,
```
domoritz, line 138: Isn't an init missing here? Otherwise `m.exp_r` isn't assigned.

domoritz:
@julieg18 hmm, yeah, something is wrong here. The types look wrong and the implementation may also be missing an init.

I think the aggregate parameter is just a value, not a field judging from the code but maybe I'm missing something: [Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykSArJQBWENgDsQAGhAATJJhSoA2qHFIEcNCAhIANnAgB9OAA8NOA9JDF9DQ2lU79D1ABZKAJgAMAgL5SoLoGEGgebt6eAGwBQS6hqEKeAIzJsc4hYZQA7AAcAp7pwa5CBW4AzEXxaOWUUXmFgRmuHrm5MU3FCUJRngCcuVWZiZTlfdluQ6612WN9fgC66eqa2mY4EnDimDT6RkjEZNZiDABOUFroXSbmCJZaMpinSOIQAGZspwiOoJgAnjhLiAkGQyKc+JgHiA3jQ4HpZAllM1QgsZGw8I4QOtNttdnoQKjgYiQOCIAw9JgCTIQWCIXAjDgkM8EIjvJQBEtOX4gA) (still fails because of some error somewhere). 

@Xitian9 @shcheklein can you post an example you used to test this feature?

packages/vega-transforms/src/Window.js
```diff
@@ -37,6 +38,7 @@ Window.Definition = {
     { 'name': 'groupby', 'type': 'field', 'array': true },
     { 'name': 'ops', 'type': 'enum', 'array': true, 'values': ValidWindowOps.concat(ValidAggregateOps) },
     { 'name': 'params', 'type': 'number', 'null': true, 'array': true },
+    { 'name': 'aggregate_params', 'type': 'field', 'null': true, 'array': true, 'default': [null] },
```
domoritz, line 41: Is the default correct here? There may be more than two ops.

packages/vega-transforms/src/Aggregate.js
domoritz, line 14: Shouldn't this be `[params.aggregate_params]` since we cannot assume a single op?

packages/vega-transforms/src/Window.js
Xitian9: The code will pad aggregate_params with nulls as necessary to match the number of ops.

Xitian9:
> @Xitian9 @shcheklein can you post an example you used to test this feature? Also, could you send a follow up pull request to add docs.

Test code is here: https://github.com/vega/vega/blob/main/packages/vega-transforms/test/aggregate-test.js

packages/vega-transforms/src/Aggregate.js
Xitian9: The code will pad aggregate_params with nulls as necessary to match the number of ops.

Xitian9:
> Hello, I'm working on adding `aggregate_parameters` into vega-lite ([vega/vega-lite#9200](https://github.com/vega/vega-lite/issues/9200)), but when I tested using the params in a vega editor spec, all I could get was `NaN`:
> image
> 
> I tried a couple different field variations but `NaN` was always the result. Have I misunderstood what `aggregate_params` is supposed to consist of? Here's the [example vega editor spec](https://vega.github.io/editor/#/url/vega/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykSArJQBWENgDsQAGhAATJJhSoA2qHFIEcNCAhIANnAgB9OAA8NOA9JDF9DQ2lUgyAJzYMcaAAwzdBiGgALJQATF4CMjhILhrelF4JAJwAvlKgru6eqD46+g6owYFeIQBskdGx2fGBiSWp6W4eaACMvnkBqEIhza0gUTEIccUlABz1zo1ZvX75wQDsIwIh5QNDI93jGU2oy7n+aEJLgQDMK5Ve1XMhm5NouzMdx5QlC7v95-FzzQI3md5t+wKlBGIzKfQqgyqXkScxSaQmf1Q03aB2eIUSIzOkIuXm+cIaiORgKExxhgSxQ0CIWu8K2WXuKNQTzmpMSFKhiVJyQAuuN1JptO8EEZbHp7NZRfYOhcfvD+Vp0GYcBI4OJMDR9EYkMQyNYxAwXFAFXtDCZzAhLFoZJgYuIIAAzNguSFOTAATxwxqQZFcfEwVpA9pocD0sg6yhNAW5MjYeEcICVKrVGr0IGjICQ4ZALkMDD0mDTMjpTDd8bphYzPpzFH9RiFWaFIrscEo8kUaZ5naAA) I was using.

"exponential" takes a number as an argument, and you have provided the string "param_value". Do you still get NaN if you supply a double?

Xitian9:
> @julieg18 hmm, yeah, something is wrong here. The types look wrong to me.
> 
> I think the aggregate parameter is just a value, not a field judging from the code but maybe I'm missing something:

I'm not a javascript programmer, so I could definitely believe I got the types wrong. The type needs to be generic enough to accept any sort of value, as different ops will have different argument types. For "exponential" that's a number, but for other as-yet-unimplemented ops it could be something of arbitrary complexity.

julieg18:
> "exponential" takes a number as an argument, and you have provided the string "param_value". 

Makes sense, thanks! As previously mentioned, I can only provide strings since currently the type is set to fields. 

> Do you still get NaN if you supply a double?

Apologies, I'm not sure what you mean by double? Do you mean something like this:

image

I would still get `NaN` in [this case](https://vega.github.io/editor/#/url/vega/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykSArJQBWENgDsQAGhAATJJhSoA2qHFIEcNCAhIANnAgB9OAA8NOA9JDF9DQ2lUgyAJzYMcaAAwzdBiGgALJQATF4CMjhILhrelF4JAJwAvlKgru6eqD46+g6owYFeIQBskdGx2fGBiSWp6W4eaACMvnkBqEIhza0gUTEIccUlABz1zo1ZvX75wQDsIwIh5QNDI93jGU2oy7n+aEJLgQDMK5Ve1XMhm5NouzMdx5QlC7v95-FzzQI3md5t+wKlBGIzKfQqgyqXkScxSaQmf1Q03aB2eIUSIzOkIuXm+cIaiORgKExxhgSxQ0CIWu8K2WXuKNQTzmpMSFKhiVJyQAuuN1JptGYcBI4OJMDR9EYkMQyNYxAwXFAtOgHiZzAhLFoZJgYuIIAAzNguSFOTAATxwypASDIrj4mC1IH1NDgelkHWU4NWAIc3JkbDwjh0DEGMiFIrFEr0ID91o9Xo0Rlsens1hchgYekwMZkNrtFAdRneCA94kzeixSbsWl5teSQA)

domoritz:
> Makes sense, thanks! As previously mentioned, I can only provide strings since currently the type is set to fields.

The editor will only warn you. Numbers are still passed to Vega.

julieg18:
> The editor will only warn you. Numbers are still passed to Vega.

Perhaps I'm misunderstanding something, but [inputting numbers causes errors to be thrown](https://vega.github.io/editor/#/url/vega/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykSArJQBWENgDsQAGhAATJJhSoA2qHFIEcNCAhIANnAgB9OAA8NOA9JDF9DQ2lUgyAJzYMcaAAwzdBiGgALJQATF4CMjhILhrelF4JAJwAvlKgru6eqD46+g6owYFeIQBskdGx2fGBiSWp6W4eaACMvnkBqEIhza0gUTEIccUlABz1zo1ZvX75wQDsIwIh5QNDI93jGU2oy7n+aEJLgQDMK5Ve1XMhm5NouzMdx5QlC7v95-FzzQI3md5t+wKlBGIzKfQqgyqXkScxSaQmf1Q03aB2eIUSIzOkIuXm+cIaiORgKExxhgSxQ0CIWu8K2WXuKNQTzmpMSFKhiVJyQAuuN1JptO8EEZbHp7NZRfYOhcfvD+Vp0GYcBI4OJMDR9EYkMQyNYxAwXFAFXtDCZzAhLFoZJgYuIIAAzNguSFOTAATxwxqQZFcfEwVpA9pocD0sg6yhNAW5MjYeEcICVKrVGr0IGjICQ4ZALkMDD0mDTMjpTDd8bphYzPpzFH9RiF4ZlvKbySAA) and nothing get rendered, in the view or data viewer: 

image

julieg18:
> Types are now fixed thanks to https://github.com/vega/vega/pull/3846, but as expected, we are still hitting the Unexpected token ']' error when trying to [run a spec](https://vega.github.io/editor/#/url/vega/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykSArJQBWENgDsQAGhAATJJhSoA2qHFIEcNCAhIANnAgB9OAA8NOA9JDF9DQ2lU79D1ABZKAJgAMAgL5SoLoGEGgebt6eAGwBQS6hqEKeAIzJsc4hYZQA7AAcAp7pwa5CBW4AzEXxaOWUUXmFgRmuHrm5MU3FCUJRngCcuVWZiZTlfdluQ6612WN9fgC66eqa2mY4EnDimDT6RkjEZNZiDABOUFroXSbmCJZaMpinSOIQAGZspwiOoJgAnjhLiAkGQyKc+JgHiA3jQ4HpZAllM1QgsZGw8I4QOtNttdnoQKjgYiQOCIAw9JgCTIQWCIXAjDgkM8EIjvJQBEtOX4gA).

Actually, I tested `aggregate_params` today with the latest version of vega in my local vega editor and everything seems to be working! 

image

Not sure why I was getting the error previously, but unless it shows up again, I'll get back to https://github.com/vega/vega-lite/issues/9200

domoritz:
Awesome. Thanks for the update. I'm glad it's working now.

PR #3718, opened 2023-04-06
feat: constant regression

domoritz:
Merge after #3717

### DIFF
--- a/packages/vega-statistics/src/regression/constant.js
+++ b/packages/vega-statistics/src/regression/constant.js
@@ -5,9 +5,7 @@ export default function (data, x, y) {
   for (const d of data) {
     const val = y(d)
     if (x(d) == null || val == null || isNaN(val)) continue;
-
-    n++;
-    mean += (val - mean) / n;
+    mean += (val - mean) / ++n;
   }
   const predict = () => mean;
 
--- a/packages/vega-statistics/src/regression/constant.js
+++ b/packages/vega-statistics/src/regression/constant.js
@@ -1,14 +1,19 @@
-import { sum } from 'd3-array';
 import rSquared from './r-squared';
 
 export default function (data, x, y) {
-  const validData = data.filter(d => x(d) != null && y(d) != null),
-    meanY = sum(validData.map(y)) / validData.length,
-    predict = () => meanY;
+  let mean = 0, n = 0;
+  for (const d of data) {
+    const val = y(d)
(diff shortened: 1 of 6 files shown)


packages/vega-statistics/src/regression/constant.js
```diff
@@ -0,0 +1,14 @@
+import { sum } from 'd3-array';
+import rSquared from './r-squared';
+
+export default function (data, x, y) {
+  const validData = data.filter(d => x(d) != null && y(d) != null),
+    meanY = sum(validData.map(y)) / validData.length,
```
jheer, line 6: Overall the PR looks great. However, there is a minor discrepancy here with how we compute averages in the Aggregate transform. There, we use an online algorithm which should be more numerically stable than a standard sum-and-divide. We may want to be consistent throughout.

jheer: Yes, NaNs should be considered "invalid" just like null/undefined. That would be consistent with Aggregate.

packages/vega-statistics/src/regression/constant.js
```diff
@@ -0,0 +1,17 @@
...
+    const val = y(d)
+    if (x(d) == null || val == null || isNaN(val)) continue;
+    mean += (val - mean) / ++n;
+  }
+  const predict = () => mean;
+
+  return {
+    coef: [mean],
+    predict: predict,
+    rSquared: rSquared(data, x, y, mean, predict)
```
thomascamminady, line 15: `rSquared` is always `0` for a constant fit. We could skip that. 
I guess this would bring efficiency at the cost of loosing consistency in the implementation when compared to other orders?

## Proof
Here's the `rSquared` implementation and with `predict` being equal to the mean being equal to `uY`, `SSE` and `SST` are equal and `1 - SSE / SST` equals zero. 
```
export default function(data, x, y, uY, predict) {
  let SSE = 0, SST = 0;

  visitPoints(data, x, y, (dx, dy) => {
    const sse = dy - predict(dx),
          sst = dy - uY;

    SSE += sse * sse;
    SST += sst * sst;
  });

  return 1 - SSE / SST;
}```

thomascamminady (commented):
Possible performance gain by skipping `rSquared` computation since it is always zero.

packages/vega-statistics/src/regression/constant.js
domoritz: Thanks. I updated the code.

PR #3724, opened 2023-04-19
fix: handle consecutive escapes for splitAccessPath

suchanlee:
There was a bug in `splitAccessPaths` where if there were consecutive `\\`s it would skip a character due to an extra increment in the index

For example:
- `\\a\\b\\c` -> `a\\bc` (which is wrong but perhaps harmless)
- `\\[\\"foo\\"\\]` -> `Access path missing open bracket` error due to skipping the last `\\` following `\\"` and getting in the `]` conditional although it's been escaped

Added tests for these cases

### DIFF
--- a/packages/vega-util/src/splitAccessPath.js
+++ b/packages/vega-util/src/splitAccessPath.js
@@ -20,8 +20,7 @@ export default function(p) {
   for (i=j=0; j<n; ++j) {
     c = p[j];
     if (c === '\\') {
-      s += p.substring(i, j);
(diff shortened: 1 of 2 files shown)


domoritz (approved):

suchanlee:
Thanks a bunch for approving! What's the process for getting it merged?

domoritz:
@jheer usually merges and releases at some point. If you need this in your package now, I recommend patching the package dependency.

PR #3728, opened 2023-04-23
feat(vega-cli): return error code on failure

davidanthoff:
This makes it easier for tools that run the cli tools to detect when something went wrong.

Same idea as https://github.com/vega/vega-lite/pull/8858.


domoritz (approved):

jheer:
Thanks!

davidanthoff:
One quick question: what is the release cycle for vega, i.e. how long do you think it will take for a new version with this to ship?

jheer:
v5.25.0 has been released!

PR #3736, opened 2023-04-28
fix: correct tickMinStep calculation logic

kanitw:
If the domain is [0, 1], the span is 1.  The ideal tick count based on minStep should output 2, not 1. 

Not sure how we suppose to test vega-scale. I don't see many test files here.


jheer:
LGTM. Feel free to add a new test cases in the `ticks-test` file. Thanks!

jheer (approved):

PR #3753, opened 2023-05-10
chore: bump dtslint version

domoritz:


jheer:
@domoritz Newer versions of dtslint result in a large cascade of typing errors in the `vega-typings` test suite that I do not have the time to make sense of and repair. If someone can sort this out, I would be more than happy to upgrade to current versions.

PR #3759, opened 2023-05-31
Make right axis rotate the same way as left

### LINKED ISSUE #3571: Should right axis title rotate the same way as left axis title by default
Looking at Vega-Lite dual axis example:

image

Readers currently have to tilt their head in two different ways to read the chart.

Alternatively, if the right axis label title gets rotated in the same way, it might be easier to read. 

image

kanitw:
Fix #3571


domoritz (approved):

jheer:
LGTM, but the test scenes in `packages/vega` need to be updated to match these changes (see the `GENERATE_SCENES` flag in `packages/vega/test/scene-test.js`).

domoritz (changes requested):
Actually, I think I like the current behavior. The axis labels are facing the chart (and are correctly rotated when I told my head to the left to read the left label and right when I tilt to the right. Yes, it's different between left and right. 

So I am now leaning against this change.

joelostblom:
My personal preference would also favor the current behavior. In terms of other libraries, [ggplot aligns with the current behavior ](https://ggplot2.tidyverse.org/reference/sec_axis.html) but others such as seaborn/mpl and tableau seems to align with the suggested behavior.

PR #3770, opened 2023-06-16
Support pointer events & updated mouse events to pointer events whenever possible

alliefeldman:
Will be swapping Vega-Lite from using 'mouse' events to using 'pointer' events. Both 'pointer' and 'mouse' events are now included in Vega's schema.


docs/vega-core.min.js
domoritz, line 1: don't include this file in the pull request, please

same with the other files in docs.

packages/vega-typings/tests/spec/invalid/signal.ts
domoritz, line 1: Why is this file gone?

domoritz (approved):
Looks great. 

I was wrong about one thing. Please also change `docs/examples/*.vg.json` since stye use mouse events right now but those are better as pointer events.

domoritz:
Please make sure that the tests pass.

PR #3781, opened 2023-07-19
Support pointer events & updated mouse events to pointer events whenever possible

domoritz:
This enables easier support for touch events and also updates all the examples to work both with a mouse cursor and touch.


domoritz (commented):
Let me know when you manually tested Vega after these changes. The code looks good to me.

Issue #3771, opened 2023-06-20
Inconsistent versions of vega-functions imported, causes failure to add expressionFunctions


DirkKaris:
After installing the most recent version of Vega (5.25.0), `yarn why -R vega-functions` shows
```
└─ bi-app@workspace:packages/bi-app
   └─ vega@npm:5.25.0 (via npm:5.25.0)
      ├─ vega-functions@npm:5.13.2 (via npm:~5.13.2)
      ├─ vega-parser@npm:6.2.0 (via npm:~6.2.0)
      │  └─ vega-functions@npm:5.13.1 (via npm:^5.13.1)
      └─ vega-view@npm:5.11.1 (via npm:~5.11.1)
         └─ vega-functions@npm:5.13.1 (via npm:^5.13.1)
```

This appears to because the [most recent version-bump commit](https://github.com/vega/vega/commit/a27d94baa9b3190590d4d6dea045cc4ee2718856) did not update vega-parser/package.json or vega-view/package.json, while [earlier similar commits](https://github.com/vega/vega/commit/1365926e397328402a1d5370ffad998e34d3816b) did.

The problem this causes is that when calling vega-embed with an expressionFunction in the options argument, that expressionFunction is set into one version of vega-functions's `codeGenerator` export, but then looked for in the other version, and then parsing the spec fails because the custom expressionFunction is not found.

DirkKaris:
This is the only inconsistency that I _know_ causes a problem, but other packages show multiple imports as well, e.g. 

```
dirk@Dirks-MacBook-Pro omni % yarn why -R vega-time
└─ bi-app@workspace:packages/bi-app
   └─ vega@npm:5.25.0 (via npm:5.25.0)
      ├─ vega-crossfilter@npm:4.1.1 (via npm:~4.1.1)
      │  └─ vega-dataflow@npm:5.7.5 (via npm:^5.7.5)
      │     ├─ vega-format@npm:1.1.1 (via npm:^1.1.1)
      │     │  └─ vega-time@npm:2.1.1 (via npm:^2.1.1)
      │     └─ vega-loader@npm:4.5.1 (via npm:^4.5.1)
      │        └─ vega-format@npm:1.1.1 (via npm:^1.1.1)
      ├─ vega-dataflow@npm:5.7.5 (via npm:~5.7.5)
      ├─ vega-encode@npm:4.9.2 (via npm:~4.9.2)
[...]

PR #3791, opened 2023-08-04
feat: bump vega typings to 1.0

domoritz:
This will make it easier to release patch versions of the typings.

### DIFF
--- a/packages/vega-typings/package.json
+++ b/packages/vega-typings/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vega-typings",
-  "version": "0.24.2",
+  "version": "1.0.0",
   "description": "Typings for Vega.",
   "types": "types",
   "repository": "vega/vega",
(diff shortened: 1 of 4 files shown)


mcnuttandrew:
Looks like 5.1 breaks ?? for older node versions (https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-1.html). This may necessitate adding a minimum engine version

packages/vega-typings/package.json
```diff
@@ -21,8 +21,8 @@
   "dependencies": {
     "@types/geojson": "7946.0.4",
     "vega-event-selector": "^3.0.1",
-    "vega-expression": "^5.0.1",
-    "vega-util": "^1.17.1"
+    "vega-expression": "^5.1.0",
```
mcnuttandrew, line 24: did you mean to version these as well?

domoritz: Yeah, I think otherwise this leads to issues like #3771

mcnuttandrew:
Per outside discussion, Lgtm

Issue #3715, opened 2023-04-04
Reuse textMetrics.width in text truncation logic


jonmmease:
[VlConvert](https://github.com/vega/vl-convert) overrides the vega-scenegraph [`textMetrics.width`](https://github.com/vega/vega/blob/d0b886881bd27da626b0d33f2db23bd17f6f4277/packages/vega-scenegraph/src/util/text.js#L7-L13) function to compute text widths using [resvg](https://github.com/RazrFalcon/resvg) (since Node canvas is not available in Deno). This gives us accurate text layouts, as the `textMetrics.width` function is called directly by the text layout logic.

But as reported in https://github.com/vega/vl-convert/issues/30, the text truncation logic doesn't work properly. I think the issue is that the `truncate` function doesn't call `textMetrics.width`, but instead uses [`widthGetter`](https://github.com/vega/vega/blob/d0b886881bd27da626b0d33f2db23bd17f6f4277/packages/vega-scenegraph/src/util/text.js#L77)

https://github.com/vega/vega/blob/d0b886881bd27da626b0d33f2db23bd17f6f4277/packages/vega-scenegraph/src/util/text.js#L89-L93

`widthGetter` determines whether to use `_measureWidth` or `_estimateWidth` based on whether `textMetrics.width === measureWidth`. It's not clear to me why `widthGetter` is needed. Could we get rid of `widthGetter` and have `truncate` call `textMetrics.width()` directly?

Thanks!

PR #3792, opened 2023-09-04
Fall back to textMetrics.width for user defined width function

jonmmease:
Closes https://github.com/vega/vega/issues/3715

## Overview
vl-convert runs Vega and Vega-Lite inside the Deno JavaScript runtime to perform static image export. Since node canvas is not compatible with Deno, vl-convert overrides the default vega-scenegraph [textMetrics.width](https://github.com/vega/vega/blob/d0b886881bd27da626b0d33f2db23bd17f6f4277/packages/vega-scenegraph/src/util/text.js#L7-L13) function to use a Rust function to compute text metrics.

This works really well in practice, the only issue we've run into is that text truncation logic ends up using the `estimateWidth` logic path rather than the overridden `textMetrics.width`.  This PR updates the `widthGetter` function to fall back to using `textMetrics.width` when this function is not equal to Vega's own `measureWidth` and `estimateWidth` functions.

cc @jheer as it looks like you wrote the widthGetter and text truncation logic 😄

### DIFF
--- a/packages/vega-scenegraph/src/util/text.js
+++ b/packages/vega-scenegraph/src/util/text.js
@@ -79,10 +79,13 @@ function widthGetter(item) {
     // we are using canvas
     const currentFont = font(item);
     return text => _measureWidth(text, currentFont);
-  } else {
+  } else if (textMetrics.width === estimateWidth) {
     // we are relying on estimates
     const currentFontHeight = fontSize(item);
     return text => _estimateWidth(text, currentFontHeight);
+  } else {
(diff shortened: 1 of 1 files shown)


jonmmease:
friendly ping. This is an update I'd really like to have for VlConvert 1.0, which is the milestone at which VlConvert will have full coverage of the features of the `altair_saver` Python library and the Vega CLI utility.

jheer:
Missing test coverage for new branch, but otherwise looks good.

PR #3810, opened 2023-10-10
feat: Add experimental hybrid renderer

jonmmease:
## Background

Vega currently bundles two renderer implementations: SVG and Canvas. The SVG renderer creates vector images, but suffers from poor performance for charts with a large number of marks items (e.g. a scatter plot with thousands of points). The Canvas renderer utilizes HTML canvas for rendering and has much better performance for large marks. The downside of the Canvas renderer is that it generates raster images, so the rendered charts get blurry when zooming in with the browser.

## Overview

This PR introduces a "Hybrid Renderer" that layers an SVG renderer on top of a Canvas renderer and provides configuration to specify which mark types are rendered to each layer. A motivating use case is to render text marks to SVG and all other mark types to Canvas. For most charts (those without large text marks), this will have equivalent performance to the pure Canvas renderer but also having the advantage of rendering the chart's text in vector SVG format.

## Updates to Renderer API
This PR adds a `markTypes` argument to the `.render()` method of the base Renderer and the SVG/Canvas Renderer subclasses. When defined, the renderer should only render mark types that are included in the `markTypes` array. When undefined, all mark types are rendered (as before).

## Hybrid Handler
The `HybridHandler` is a subclass of `CanvasHandler` that attaches event callbacks to the Canvas layer. The `HybridRenderer` sets `pointer-events: "none"` for the top SVG layer so that all mouse events are handled by the lower Canvas layer.

## Playing with it
I updated `packages/vega/test/index.html` to include "hybrid" as a third renderer option. The particular configuration of the hybrid renderer can be configured using the `setHybridRendererOptions` function.

```
vega.setHybridRendererOptions({svgMarkTypes: ["text"], svgOnTop: true, debug: false});
```

If you set `debug` to `true`, the top two renderers will be laid out vertically instead of overlayed. This makes it easier to see what's going on.


packages/vega-scenegraph/test/svg-renderer-test.js
```diff
@@ -188,7 +188,7 @@ tape('SVGRenderer should support enter-item redraw', t => {
   rect2.bounds = new Bounds().set(70, 10, 120, 60);
   rects.items.push(rect2);
 
-  var svg = compensate(r.render(scene, [rect1, rect2]).svg());
+  var svg = compensate(r.render(scene).svg());
```
jonmmease, line 191: Then `render` method previously had a single argument (the scene) so the second argument didn't do anything. Now the second argument is the optional array of mark types, so I removed the second argument in these tests

packages/vega-scenegraph/src/HybridRenderer.js
```diff
@@ -0,0 +1,96 @@
...
+   * @param {object} scene - The root mark of a scenegraph to render.
+   * @param {Array} markTypes - Array of the mark types to render.
+   *                            If undefined, render all mark types
+   */
+  _render(scene, markTypes) {
+    const allMarkTypes = markTypes ?? [
+      'arc', 'area', 'image', 'line', 'path', 'rect', 'rule', 'shape', 'symbol', 'text', 'trail'
+    ];
+    const canvasMarkTypes = allMarkTypes.filter((m) => !OPTS.svgMarkTypes.includes(m));
+    this._svgRenderer.render(scene, OPTS.svgMarkTypes);
```
lsh, line 76: I am curious about the logical impact of this setup. I'm fine with a constraint like "one gets one canvas and one svg layer," but I think the crux of the design work will be ironing out this constraint. For instance, should we have an option of which is on top? I could see one example being that text rendering occurs on an svg context beneath a canvas context that has animated content (but which should be allowed to run on top of the axis).

domoritz:
This will be awesome together with the webgpu renderer since we could skip supporting text to make it much easier. @lsh

packages/vega-scenegraph/src/HybridRenderer.js
jonmmease: Yeah, those are good points. I added support for placing the canvas layer on top in [4a44501](https://github.com/vega/vega/pull/3810/commits/4a44501051fa519e73690fef9c3ef29a0a2d6773).

jonmmease:
> This will be awesome together with the webgpu renderer since we could skip supporting text to make it much easier

Yeah, the HybridRenderer doesn't actually end up having a lot of logic, so one option would be to just have a `webgpu` renderer type, and internally it would use the same kind of logic as HybridRenderer to layer a WebGPU canvas with an SVG renderer layer.  Or we could extend the HybridRenderer with WebGPU support.

jonmmease:
ping @vega/vega for review

lsh (approved):
Looks good to me! Only concern is that we may want to label this experimental for now to give us a chance to nail down the API later on.

jonmmease:
> Only concern is that we may want to label this experimental for now to give us a chance to nail down the API later on.

Do you mean in the docs/changelog? I updated the PR title to add "experimental" so this will propagate through to the changelog.

jheer:
This is awesome. Thanks @jonmmease!

PR #3827, opened 2023-11-09
feat: Expose aggregate parameters through vega-typings

Xitian9:
This exposes the changes made in #3686 in vega-typings.

* feat(vega-typings): expose aggregate_params in AggregateTransform
* feat(vega-typings): expose exponential and exponentialb

### DIFF
--- a/packages/vega-typings/types/spec/transform.d.ts
+++ b/packages/vega-typings/types/spec/transform.d.ts
@@ -86,6 +86,7 @@ export interface AggregateTransform {
   groupby?: FieldRef[] | SignalRef;
   fields?: (FieldRef | null)[] | SignalRef;
   ops?: (AggregateOp | SignalRef)[] | SignalRef;
+  aggregate_params?: object[];
   as?: (string | SignalRef | null)[] | SignalRef;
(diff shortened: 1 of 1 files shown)


packages/vega-typings/types/spec/transform.d.ts
```diff
@@ -86,6 +86,7 @@ export interface AggregateTransform {
   groupby?: FieldRef[] | SignalRef;
   fields?: (FieldRef | null)[] | SignalRef;
   ops?: (AggregateOp | SignalRef)[] | SignalRef;
+  aggregate_params?: object[];
```
domoritz, line 89: Shouldn't this be `number[]`?

Xitian9: It should take any sort of input. Different ops will take different parameters. "exponential" takes a number, but "cook_breakfast" might take a list of string ingredients, and "calculate_gravitational_field" might take a complicated object describing the local spacetime field.

domoritz: Let's change it when we actually have aggregates that use those types.

PR #3841, opened 2023-12-07
docs: Addition of an example: Zoomable Circle Packing

Giammaria:
Added an example: Zoomable Circle Packing under the _Interaction Techniques_ section.


domoritz:
Let's leave it in the existing section and refactor when there are more examples. Moving to another section is easy.

jheer:
Thanks @Giammaria!

docs/examples/zoomable-circle-packing.md
```diff
@@ -0,0 +1,11 @@
+---
+layout: example
+title: Zoomable Circle Packing Example
+permalink: /examples/zoomable-circle-packing/index.html
+spec: zoomable-circle-packing
+image: /examples/img/zoomable-circle-packing.png
+---
+
+This is an extension of the [Circle Packing Example](../examples/circle-packing). This version was implemented by [@giammaria](https://github.com/Giammaria). It incorporates a timer to facilitate zoom and fade animations, offering a technique beneficial for drill-down behavior and exploration. For simplicity, the example reveals only a few text marks upon zoom, corresponding to the user-selected node. Alternatively, you can render various mark types, axes, legends, etc., to provide additional insights into the selected node."
```
domoritz, line 9: is there an extra " at the end?

Giammaria: Good catch. There is. I can update it

jwoLondon:
Is `marknames` doing anything [here](https://vega.github.io/vega/examples/zoomable-circle-packing/)? I don't see it (plural version with an array) in the schema/documentation, and removing it from the example spec doesn't seem to make any difference.

Giammaria:
> Is `marknames` doing anything [here](https://vega.github.io/vega/examples/zoomable-circle-packing/)? I don't see it (plural version with an array) in the schema/documentation, and removing it from the example spec doesn't seem to make any difference.

It is not. I should have removed it before submitting.

domoritz:
Please send a follow up pr to update the spec.

jwoLondon:
I think also there is a mistake with the `hover` encoding of `circles`. The `strokeWidth` setting works, but there is an incorrect `color` setting (fill?). I'm not quite sure what the intention is here so not sure how best this would be fixed other than removing it entirely (which doesn't seem to affect the visualization).

I really like the example, so it would be great to see this with code as simplified as possible while showcasing the eased zooming transitions.

PR #3843, opened 2023-12-12
feat: Add observable10 palette

mcnuttandrew:
Announced [here](https://twitter.com/mbostock/status/1734623702157873369/photo/1) observable has a new color palette called observable10. I don't have a great sense if this type of keeping of with the color-pal-joneses would be seen as useful here, but I figured I'd add it and see what you all thought.


domoritz:
Screenshots. Looks nice to me. Let's add it. Do you know whether the colors are the same for dark and light mode or whether these are two palettes? 

image
image
image

domoritz (approved):

domoritz:
Before we merge, I'd like to get some sign off from other steering committee members.

mcnuttandrew:
I went into the code to double check if its just one pal---it does appear to be, see [here](https://github.com/observablehq/plot/blob/main/src/scales/schemes.js)---and realized my image-based measurement of the colors was slightly incorrectly, colors should now be right

domoritz:
Can we pull in https://github.com/d3/d3-scale-chromatic/pull/51 directly? Otherwise we should check for license/attribution requirements.

arvind:
Odd question but: would the Observable folks be happy if we were to include the color palette? I'm not sure if color palettes are licensed but, given the name, I wonder if they want their palette to be more readily identifiable with their brand/Plot? I would hate to step on their toes.

(The fact that they've contributed it back to D3 makes me think it's probably ok, but it'd be good to get their explicit approval.)

mcnuttandrew:
these are fair points! Elsewhere Maureen also noted that "If you want to add it, let’s please reach out to observable and get both the correct colors and and permission. Lots of work goes into making a good general purpose palette, let’s respect that."

- From a license sense, i suspect that this fully above board (https://github.com/observablehq/plot/blob/main/LICENSE) if an attribution is added. Although I'll admit im no expert in that area. 
- From a being community members/digital neighbors sense, yes I think it would be probably be good to check in.
- If it would be preferable, I definitely can try to add support through d3-scale-chromatic, but i'd like to be sure that's what's best before going down that road given that's its not part of vega now. Pulling it in might suggest some restructuring to palette.js more generally, but would probably have the benefit that more palettes would come for free. (although tbh  i'm having some trouble node-gyp based issues getting the local development environment operating, which is my biggest blocker for just trying it out)

jheer:
Let's just ask @mbostock: Mike, do you (and Observable) support adding your new `obserable10` color palette as an option within Vega / Vega-Lite, without directly depending on d3-scale-chromatic?

mbostock (approved):
👍👍 Go for it! Thanks for asking.

Fil:
Please hold on, we're making a slight adjustment to the palette.

domoritz:
I would prefer to hold off and just import https://github.com/d3/d3-scale-chromatic/pull/51 directly.

Fil:
The scheme has been updated (reordered) in plot [0.6.13](https://github.com/observablehq/plot/releases/tag/v0.6.13).

Per our CHANGELOG:

> In 0.6.12, we inadvertently published an earlier draft; this fixed version better aligns the order of colors with *tableau10*.

The correct scheme (reference below) is identical to [d3-scale-chromatic#51](https://github.com/d3/d3-scale-chromatic/pull/51). Please, update this PR to match (or follow @domoritz  and use d3-scale-chromatic).

I want to thank @mcnuttandrew and the team here for your enthusiasm and reactivity. It's by reviewing this PR that I noticed the discrepancy, and our mistake! (And fortunately it was not too late to hot-fix.)

reference

(Note the **green in position 4**.)

jheer:
For simplicity and to avoid bloat I would recommedn that we _not_ add a new dependency in this PR -- just add the palette within our existing setup.

If we want to add a d3-scale-chromatic dependency we should alsoo reconsider our palette system. At the moment (for better or worse) a direct replacement will "break" some of the adjustments in place for Vega color schemes.

domoritz (changes requested):
Let's replace with an import so we are up to date with the latest colors.

mcnuttandrew:
Alright with some offline help from @domoritz, I have gotten the scale-chromatic import working as requested. The version of d3-scale-chromatic that is current published (v3.0.0) does not include observable 10, and so i have just updated the stringified colors per Fil's updates. Once the new version is published we can just add a link so we'll always have the most up to date colors. 

The way i'm including the scale-chromatic schemes is trying to be as un-distruptive as possible, which necessitates a slightly odd hack in which schemes from scale-chromatic are converted from hex[] back into the strings of hexes that generates them. A more holistic rethinking of how palettes work is certainly possible, but in this modestly scoped PR I'm merely trying to make it possible to keep up to date within the current architecture.

domoritz:
Should we wait for the release then?

arvind:
Sorry, just to double check my understanding, wasn't the preference that @jheer mentioned above to _not_ use a d3-scale-chromatic because there are "adjustments in place for Vega color schemes"?

mcnuttandrew:
> Should we wait for the release then? 

Sure! I'll follow up once things d3-scale-chromatic versions

> Sorry, just to double check my understanding, wasn't the preference that @jheer mentioned above to not use a d3-scale-chromatic because there are "adjustments in place for Vega color schemes"?

This PR is kinda doing both! The way scale-chromatic is added here does not change the current color support in any way. It totally could be more thoughtfully rethought into something fancier, but for now its just providing a link to the small set of schemes that can be supported without any changes to vega's color support

domoritz:
I would prefer an import over copy but I don't know what the adjustments in place are.

keckelt:
So, to recap: 

* The PR adds d3-scale-chromatic as a dependency, 
* vega color schemes that exist in d3-scale-chromatic are no longer hardcoded but imported
* the imported d3-scale-chromatic seems are converted to a string of hex values.
  The difference is, that the strings are now dynamically generated (from d3) instead of being hardcoded.

However, the colors remain hardcoded in the documentation. I don't expect them to change, but if they do there would be a mismatch between the documentation and the colors actually used. How do you see this? Should we test whether the colors are still as expected?

packages/vega-scale/src/palettes.js
```diff
@@ -62,19 +74,22 @@ export const continuous = {
   darkRed: '3434347036339e3c38cc4037e75d1eec8620eeab29f0ce32ffeb2c'
 };
 
+const toStream = (scheme) => scheme.map((x) => x.slice(1).toLowerCase()).join('');
```
domoritz, line 77: Would it be cheaper to lowercase at the end after joining?

mcnuttandrew:
Yep @keckelt that is an accurate summary!

> However, the colors remain hardcoded in the documentation.

I've wondered about the documentation and if it might be better to auto-generate from the palettes, that way there isn't as much manual/error prone work to generate the colors for the docs. I'd be happy to write a script (in a different PR) for generating them if that'd be helpful

> I don't expect them to change, but if they do there would be a mismatch between the documentation and the colors actually used. How do you see this? Should we test whether the colors are still as expected?

I am confident that the colors (as manifest in this PR) are exactly the same. Here's the script I was running inside of the tests to figure out what's what.

```js
import * as scaleChromatic from 'd3-scale-chromatic'
Object.keys(discrete).forEach((name) => {
  const vegaScheme = discrete[name];
  const chromaticName = `scheme${name[0].toUpperCase()}${name.slice(1)}`;
  if (!scaleChromatic[chromaticName]) {
    return;
  }
  const chromaticScheme = toStream(scaleChromatic[chromaticName]);
  if (vegaScheme === chromaticScheme) {
    console.log(name, chromaticName, 'matched!')
  }
});
```
Which returned
```sh
category10 schemeCategory10 matched!
accent schemeAccent matched!
dark2 schemeDark2 matched!
paired schemePaired matched!
pastel1 schemePastel1 matched!
pastel2 schemePastel2 matched!
set1 schemeSet1 matched!
set2 schemeSet2 matched!
set3 schemeSet3 matched!
```

Something like this could be added as a test to the vega-scale test suite, but it also seems weird to test a dep like that? Happy to add it if you want though.

keckelt:
> Something like this could be added as a test to the vega-scale test suite, but it also seems weird to test a dep like that? Happy to add it if you want though.

Yes, I also find it a bit weird. That was the only thing I noticed when I looked through the PR. That's why I wanted to point it out again.

mcnuttandrew:
my understanding is that i was blocked by https://github.com/d3/d3-scale-chromatic/pull/51 landing the new observable 10

domoritz:
Should we make in the stream until that's released? I think the scheme doesn't change often if at all.

mcnuttandrew:
I am not entirely convinced that its true that it won't change, which I suspect is why it's idling.
 
In any case, I'm not sure what else needs to be done to wrap it up?  I think i addressed the requested changes? but maybe im missing something obvious?

Fil:
It will not change, but we need to take a moment to cut a new release (of d3-scale-chromatic but also of its dependents: d3, Plot).

domoritz:
What's the performance impact of running `toStream` for every scheme? If we put the compiled strings in the code, we don't have to run the method every time Vega is loaded. But maybe the performance impact is super tiny. 

It's probably tiny given that the method is just

```js
const toStream = (scheme) => scheme.map((x) => x.slice(1)).join('').toLowerCase();
```

mcnuttandrew:
It appears to be trivial. it's a little hard to test because it is already so fast, but it appears add about 5 milliseconds (measured from within the test suite).

domoritz:
Okay. Then let's wait for https://github.com/d3/d3-scale-chromatic/pull/51 and then merge this.

mbostock:
d3-scale-chromatic 3.1.0 has now been published.

mcnuttandrew:
@domoritz i think this is ready to go now!

packages/vega-scale/src/palettes.js
```diff
@@ -62,19 +75,21 @@ export const continuous = {
   darkRed: '3434347036339e3c38cc4037e75d1eec8620eeab29f0ce32ffeb2c'
 };
 
+const toStream = (scheme) => scheme.map((x) => x.slice(1)).join('').toLowerCase();
 export const discrete = {
-  category10: '1f77b4ff7f0e2ca02cd627289467bd8c564be377c27f7f7fbcbd2217becf',
+  accent: toStream(schemeAccent),
+  category10: toStream(schemeCategory10),
```
domoritz, line 81: I think it's funny that we are basically reversing the color generation from https://github.com/d3/d3-scale-chromatic/blob/2c52792197299346b7bdb94322bb4dff8f554fea/src/categorical/category10.js#L3.

mbostock: Small warning here: these are currently hexadecimal colors, but the format might change in the future. I don’t think in practice we will change them, but I don’t want to commit to never changing them. And certainly we might introduce new color schemes in the future that using a different format (_e.g._, to use the wider p3 color space).

mcnuttandrew: > I think it's funny that we are basically reversing the color generation
this is an irony very much not lost on me :)

> Small warning 
That's a very good point, but my intuition is that's a bridge that can be crossed when it's encountered. What would be the right way to guard against this change? Including a note saying to look out for that? Include a check in toStream that throws if it encounters not hex characters?

domoritz: Yeah, we can cross that bridge when we come to it. 

For now, I wonder whether we can remove the `toStream` method and recognize if we have an array of colors in the palette and skip the `colors` method in schemes.js.

mcnuttandrew: ok that appears to be fine

domoritz:
Thank you @mcnuttandrew

PR #3844, opened 2023-12-13
fix: do not cache `devicePixelRatio`

lsh:
Currently we cache the `devicePixelRatio` on the creation of the renderer and never check to update it again. The problem with this behavior is that zooming in or out of a page will change the `devicePixelRatio` (which can be observed via `window.onresize`). In turn, a render will be blurry on zoom. This PR makes us recalculate the `devicePixelRatio` on resize, which is negligibly expensive.

### DIFF
--- a/packages/vega-scenegraph/src/util/canvas/resize.js
+++ b/packages/vega-scenegraph/src/util/canvas/resize.js
@@ -2,14 +2,12 @@ function devicePixelRatio() {
   return typeof window !== 'undefined' ? window.devicePixelRatio || 1 : 1;
 }
 
-var pixelRatio = devicePixelRatio();
-
 export default function(canvas, width, height, origin, scaleFactor, opt) {
   const inDOM = typeof HTMLElement !== 'undefined'
               && canvas instanceof HTMLElement
               && canvas.parentNode != null,
         context = canvas.getContext('2d'),
-        ratio = inDOM ? pixelRatio : scaleFactor;
+        ratio = inDOM ? devicePixelRatio() : scaleFactor;
 
   canvas.width = width * ratio;
   canvas.height = height * ratio;
--- a/packages/vega-view/src/size.js
+++ b/packages/vega-view/src/size.js
@@ -81,6 +81,10 @@ export function resizeView(viewWidth, viewHeight, width, height, origin, auto) {
       view._viewWidth = viewWidth;
     }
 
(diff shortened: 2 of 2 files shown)


domoritz (approved):
Looks good. Is there a unit test we could write for this?

lsh:
I think that would be valuable, but do we have to worry about how that interacts with react?

domoritz:
Are you concerned about the global variable or which part? Vega already has a bunch of globals so I don't see any additional concerns. I'd focus on what's best for Vega for now and then worry about compatibility with React.

tuner:
You could also listen to `devicePixelRatio` changes using media queries: https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio#monitoring_screen_resolution_or_zoom_level_changes

I think this solution would also cover cases where the browser window is dragged from a low-dpi display to a hi-dpi display.

packages/vega-view/src/View.js
```diff
@@ -113,6 +113,30 @@ export default function View(spec, options) {
...
   if (options.container) view.initialize(options.container, options.bind);
+  
+  // based on https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio#monitoring_screen_resolution_or_zoom_level_changes
+  let remove = null;
+  const updatePixelRatio = () => {
+    if (remove != null) {
+      remove();
+    }
+    const mqString = `(resolution: ${window.devicePixelRatio}dppx)`;
+    const media = matchMedia(mqString);
```
domoritz, line 124: Why not inline it?

```suggestion
    const media = matchMedia(`(resolution: ${window.devicePixelRatio}dppx)`);
```

packages/vega-view/src/View.js
```diff
@@ -385,5 +389,8 @@ inherits(View, Dataflow, {
 
   // -- SAVE / RESTORE STATE ----
   getState,
-  setState
+  setState,
+
+  // RESIZE HANDLER
```
domoritz, line 394: Is that up to date?

domoritz (approved):
Needs docs and ts typings but otherwise looks good.

packages/vega-view/src/View.js
lsh: Could you clarify the question? The only change is adding a new method via the helper.

kanitw (approved):

PR #3845, opened 2023-12-13
refactor: replace indexOf with equivalent includes and startsWith

domoritz:


lsh (approved):
Thanks for cleaning this up!

PR #3864, opened 2024-01-22
refactor: convert some of the Scenegraph types to classes

lsh:
Vega currently uses the `inherits` method in a bunch of places mimic classes, and many of the classes are ES5 style classes. This PR upgrades some of the scenegraph types to ES6 classes and uses built in inheritance so that we get benefits of editor support and our code is more idiomatic.


domoritz:
This could be a refactor:

domoritz (approved):
This is great. How does this affect performance? I think the previous inherits copied properties so they wouldn't need to go through the prototype chain. 

Otherwise, I think this makes code more readable so +1.

PR #3874, opened 2024-02-13
fix: add missing `aggregate_params` to window transform type

julieg18:
`aggregate_params` can also be used in [window transforms](https://github.com/vega/vega/blob/main/packages/vega-transforms/src/Window.js#L41), but the type `WindowTransform` is not updated. 

Related to https://github.com/vega/vega-lite/pull/9225#discussion_r1450647420, followup to #3686


domoritz (approved):
Thank you

PR #3896, opened 2024-03-17
turn off all handlers in View.finalize()

cmerrick:
We traced a memory leak in our application back to dangling event listeners that had been registered with `addEventListener`. Removing them from outside the library was possible, but tedious - this is much cleaner.

### DIFF
--- a/docs/docs/api/view.md
+++ b/docs/docs/api/view.md
@@ -79,7 +79,7 @@ var view = await new vega.View(runtime, {
 view.<b>finalize</b>()
 [<>](https://github.com/vega/vega/blob/master/packages/vega-view/src/finalize.js "Source")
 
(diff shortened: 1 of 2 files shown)


domoritz:
This looks reasonable to me but I wonder whether the previous behavior with intentional @jheer @arvind?

domoritz (approved):

jheer:
LGTM. I _think_ the prior reasoning was that those listeners would be garbage collected with the View and its contents. Either that was a faulty assumption or observed memory leaks are due to other lingering references to “stale” content.

jheer (approved):

domoritz:
Thank you @cmerrick

cmerrick:
Thanks for the quick review and merge!

PR #3904, opened 2024-04-19
fix: Respect tickMinStep for binned scale

alexxu-db:
Currently, we do not pass in the `count` (calculated based on `tickMinStep`) while generating tick values for binned scale. 

This causes performance issues especially when there exist a huge amount of bins. Attempting to render too many ticks will sometimes even crash the browser tab. 

Before:
Screenshot 2024-04-19 at 3 39 57 PM

After:
Screenshot 2024-04-19 at 3 39 26 PM

(`tickMinStep` both set to `0.1`)

### DIFF
--- a/packages/vega-scale/src/ticks.js
+++ b/packages/vega-scale/src/ticks.js
@@ -101,7 +101,7 @@ export function validTicks(scale, ticks, count) {
  * @return {Array<*>} - The generated tick values.
  */
 export function tickValues(scale, count) {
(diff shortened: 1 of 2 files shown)


kanitw (approved):

kanitw:
LGTM. I'll defer to @lsh to merge and release.

lsh (approved):
Agree with @kanitw, looks good to me.

Issue #3893, opened 2024-03-14
Mobile webview tap on the chart doesn't display tooltips properly since vega version 5.26.0


ZhaohuaZeng-at:
Please:

- [x] Check for duplicate issues. Please file separate requests as separate issues on GitHub.
- [x] Describe how to reproduce the bug.
- [x] Use the latest Vega version, if possible. Always indicate which version you are using.
- [x] If you are using Kibana, first read the [instructions for posting Kibana Vega issues](https://gist.github.com/nyurik/736c34970ba3c32f3fe3d45d66719b3e).
- [x] Include error messages or gifs/screenshots of problematic behavior, if applicable.
- [x] Provide an example specification and accessible data. Try to provide the simplest specification that reproduces the problem. To share a specification, use the [Vega-Editor](https://vega.github.io/editor/) and click "Share" to embed a working link, or provide an example spec in JSON, wrapped by triple backticks like this:

Hi there, I think there is a bug introduced in 5.26.0 that makes tooltips not show up on tap on a mobile device. 

[...]

PR #3907, opened 2024-04-23
fix: Revert pointer events to mouseevents

lsh:
Reverts #3781 in order to mitigate #3893.


domoritz (changes requested):
I don't think we should revert everything. Just the change to tooltips is enough.

lsh:
Closing in favor of #3909

PR #3909, opened 2024-04-25
fix: Revert TooltipHideEvent to MouseOutEvent

kadamwhite:
Alternative approach to #3907, which selectively reverts only one line from #3781 in order to mitigate #3893.

**Untested**; I haven't been able to build a test bundle of the project as the `yarn` install is failing on my M1 mac due to missing build dependencies.


lsh (approved):
I played around with this PR for a while and it didn't seem to break anything, though I also couldn't recreate the issue outside of codesandbox/codepen. Either way, this is a safe change, so looks good to me.

PR #3920, opened 2024-05-07
chore: migrate to ESLint flat config

lsh:
ESLint is removing the `.eslintrc` and `.eslignore` configuration options in version 9. Ahead of that transition, I am migrating us to use the new flat configuration format. I also remove `tslint` and `dtslint`, since those dependencies are being replaced by ESLint itself.


packages/vega-geo/src/util/contours.js
```diff
@@ -62,8 +62,8 @@ export default function() {
   // Marching squares with isolines stitched into rings.
   // Based on https://github.com/topojson/topojson-client/blob/v3.0.0/src/stitch.js
   function isorings(values, value, callback) {
-    var fragmentByStart = new Array,
-        fragmentByEnd = new Array,
+    var fragmentByStart = [],
```
lsh, line 65: Note that this was the only violator of the "don't use `new Array` lint, and it didn't feel worth it to do a follow up just for that.

kanitw (approved):

kanitw:
no strong opinion here.

Issue #3924, opened 2024-05-19
developer-experience: deployment previews for pull requests


hydrosquall:
## Enhancement Description

Goal: Reduce time and effort to validate change requests, and improve review quality for vegachange requests (PRs). By making a deployment preview easy, it is easier to test user facing behaviors. Similar to https://github.com/vega/vega-lite/issues/9276

Motivating use-cases

- When new PRs are opened, enable maintainers to review and test edge cases
- Create a test environment that shields users having having to buid locall
- See discussion w/ @domoritz here: https://github.com/vega/vega-lite/pull/9262#issuecomment-1950624437

## Expected behavior

- When a new PR is opened, build a copy of the vega vega/editor that uses this branch's copy of vega

## Notes

- I'm going to try doing this with Cloudflare pages, similarly to what I did for vega-lite. 
- I previously tried doing this with pure github actions, but was unsuccessful ( https://github.com/vega/vega-lite/pull/9277#issuecomment-1975194856 )

PR #3925, opened 2024-05-19
build: add deploy preview of the Vega editor

hydrosquall:
## Motivation

- Fix https://github.com/vega/vega/issues/3924

## Changes

- Set explicit yarn and node versions so that the build step can succeed in Cloudflare CI
- Move installing data out of the postinstall as it was causing the initial install to break. I moved it to a docs building step
- Add a build script for setting up a site deploy preview

## Testing

image

- Visit the preview URL: https://cameron-yick-deploy-previews.vega-628.pages.dev/
- Check that console logs from both `vega` and a subpackage like `vega-view` are both visible. 
  - (I'll remove these statements before merging)

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,8 +59,5 @@
   "engines": {
     "node": ">=18.18.0"
   },
-  "packageManager": "yarn@1.22.19",
-  "volta": {
-    "node": "18.18.0"
-  }
+  "packageManager": "yarn@1.22.19"
 }
--- a/package.json
+++ b/package.json
@@ -9,14 +9,15 @@
     "clean": "lerna clean --yes && lerna exec -- rimraf build && lerna exec -- rimraf LICENSE && rimraf node_modules yarn.lock",
     "data": "rsync -r node_modules/vega-datasets/data/* docs/data",
     "docs": "cd docs && bundle exec jekyll serve -I -l",
-    "docsbuild": "yarn data && cd packages/vega/build && cp vega.js vega.min.js* vega-core.js vega-core.min.js* vega-schema.json ../../../docs/",
+    "docsbuild": "cd packages/vega/build && cp vega.js vega.min.js* vega-core.js vega-core.min.js* vega-schema.json ../../../docs/",
     "license": "lerna exec -- cp ../../LICENSE .",
     "release": "yarn license && lerna publish from-package",
     "serve": "serve packages/vega/ -l 8080",
     "test": "yarn test:no-lint && yarn lint",
     "test:no-lint": "lerna run test --ignore vega-typings && lerna run test --scope vega-typings",
     "lint": "ESLINT_USE_FLAT_CONFIG=true yarn eslint -c eslint.config.mjs .",
-    "format": "yarn lint --fix"
+    "format": "yarn lint --fix",
+    "postinstall": "scripts/postinstall.sh"
   },
   "dependencies": {},
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -9,15 +9,14 @@
     "clean": "lerna clean --yes && lerna exec -- rimraf build && lerna exec -- rimraf LICENSE && rimraf node_modules yarn.lock",
     "data": "rsync -r node_modules/vega-datasets/data/* docs/data",
     "docs": "cd docs && bundle exec jekyll serve -I -l",
-    "docsbuild": "cd packages/vega/build && cp vega.js vega.min.js* vega-core.js vega-core.min.js* vega-schema.json ../../../docs/",
+    "docsbuild": "yarn data && cd packages/vega/build && cp vega.js vega.min.js* vega-core.js vega-core.min.js* vega-schema.json ../../../docs/",
     "license": "lerna exec -- cp ../../LICENSE .",
     "release": "yarn license && lerna publish from-package",
     "serve": "serve packages/vega/ -l 8080",
     "test": "yarn test:no-lint && yarn lint",
     "test:no-lint": "lerna run test --ignore vega-typings && lerna run test --scope vega-typings",
     "lint": "ESLINT_USE_FLAT_CONFIG=true yarn eslint -c eslint.config.mjs .",
-    "format": "yarn lint --fix",
-    "postinstall": "yarn data"
+    "format": "yarn lint --fix"
   },
   "dependencies": {},
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -53,5 +53,6 @@
     "typescript-eslint": "^7.8.0",
     "vega-datasets": "^2.8.0"
   },
-  "workspaces": ["packages/*"]
+  "workspaces": ["packages/*"],
+  "packageManager": "yarn@1.22.19"
 }
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "repository": "vega/vega",
   "scripts": {
     "build": "lerna run build && yarn docsbuild",
+    "build-editor-preview": "scripts/build-editor-preview.sh",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -52,5 +53,11 @@
     "typescript-eslint": "^7.8.0",
     "vega-datasets": "^2.8.0"
   },
-  "workspaces": ["packages/*"]
+  "workspaces": [
+    "packages/*"
+  ],
+  "engines": {
```
hydrosquall, line 59: image

package.json
```diff
@@ -52,5 +53,11 @@
...
+    "node": ">=18.18.0"
+  },
+  "packageManager": "yarn@1.22.19"
```
hydrosquall, line 62: Otherwise

```
16:26:44.651 | Detected the following tools from environment: yarn@3.6.3, nodejs@18.17.1
-- | --
16:26:44.651 | Installing project dependencies: yarn
16:26:45.212 | ➤ YN0070: Migrating from Yarn 1; automatically enabling the compatibility node-modules linker 👍
16:26:45.212 |  
16:26:45.342 | ➤ YN0000: ┌ Resolution step
16:26:46.174 | ➤ YN0032: │ canvas@npm:2.11.2: Implicit dependencies on node-gyp are discouraged
16:26:46.421 | ➤ YN0032: │ nan@npm:2.17.0: Implicit dependencies on node-gyp are discouraged
16:26:47.253 | ➤ YN0032: │ fsevents@npm:2.3.2: Implicit dependencies on node-gyp are discouraged
16:26:49.121 | ➤ YN0061: │ @npmcli/move-file@npm:2.0.1 is deprecated: This functionality has been moved to @npmcli/fs
16:26:49.642 | ➤ YN0061: │ request@npm:2.88.2 is deprecated: request has been deprecated, see https://github.com/request/request/issues/3142
16:26:49.660 | ➤ YN0061: │ uuid@npm:3.4.0 is deprecated: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
16:26:49.815 | ➤ YN0061: │ har-validator@npm:5.1.5 is deprecated: this library is no longer supported
16:26:50.972 | ➤ YN0032: │ @parcel/watcher@npm:2.0.4: Implicit dependencies on node-gyp are discouraged
16:26:51.046 | ➤ YN0032: │ node-addon-api@npm:3.2.1: Implicit dependencies on node-gyp are discouraged
16:27:01.937 | ➤ YN0002: │ @nrwl/devkit@npm:16.10.0 doesn't provide nx (p048f2), requested by @nx/devkit
16:27:01.937 | ➤ YN0060: │ root-workspace-0b6124@workspace:. provides eslint (p1305c) with version 8.53.0, which doesn't satisfy what typescript-eslint and some of its descendants request
16:27:01.938 | ➤ YN0000: │ Some peer dependencies are incorrectly met; run yarn explain peer-requirements <hash> for details, where <hash> is the six-letter p-prefixed code
16:27:01.942 | ➤ YN0000: └ Completed in 16s 599ms
16:27:02.015 | ➤ YN0000: ┌ Post-resolution validation
16:27:02.016 | ➤ YN0028: │ The lockfile would have been modified by this install, which is explicitly forbidden.


```

package.json
```diff
@@ -5,18 +5,18 @@
...
-    "docsbuild": "cd packages/vega/build && cp vega.js vega.min.js* vega-core.js vega-core.min.js* vega-schema.json ../../../docs/",
+    "docsbuild": "yarn data && cd packages/vega/build && cp vega.js vega.min.js* vega-core.js vega-core.min.js* vega-schema.json ../../../docs/",
     "license": "lerna exec -- cp ../../LICENSE .",
     "release": "yarn license && lerna publish from-package",
     "serve": "serve packages/vega/ -l 8080",
     "test": "yarn test:no-lint && yarn lint",
     "test:no-lint": "lerna run test --ignore vega-typings && lerna run test --scope vega-typings",
     "lint": "ESLINT_USE_FLAT_CONFIG=true yarn eslint -c eslint.config.mjs .",
-    "format": "yarn lint --fix",
-    "postinstall": "yarn data"
```
hydrosquall, line 19: If we must keep this as a postinstall step and can't mopve this to docsbuild, an alternate approach would be to replace this with a guard so that I can disable the postinstall step with an env variable.

package.json
```diff
@@ -52,5 +52,14 @@
...
   },
-  "workspaces": ["packages/*"]
+  "workspaces": [
+    "packages/*"
+  ],
+  "engines": {
+    "node": ">=18.18.0"
+  },
+  "packageManager": "yarn@1.22.19",
+  "volta": {
```
hydrosquall, line 62: Recommended node version manager - can drop this if it's distracting / people prefer to bring their own node version management solution

https://volta.sh/

As it stands, you won't be able to build the project if you have an older version (like node 18.15) running locally.

package.json
hydrosquall: Convinced myself it would be less of a change for existing devs to take this route.

https://github.com/vega/vega/pull/3925/commits/d4bacf02452ea28f356da5edbed7e18f6d2bf8cd

scripts/build-editor-preview.sh
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+# Build the editor site and replace main build with local copy of vega-lite
+echo "Installing Vega"
+
+# Link every package to make sure the editor uses the local version
+for package in packages/*; do
```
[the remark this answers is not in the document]
hydrosquall: Lerna can handle running a command in every package, but for linking it delegates to yarn / npm https://lerna.js.org/docs/legacy-package-management#migrating-from-lerna-bootstrap-lerna-add-and-lerna-link-in-lerna-v7-and-later

It's supposed to prefer the local version if that packages the package version but the [bug](https://vega-js.slack.com/archives/C0NMEF7EF/p1716001329224679) we're currently facing in vega is that multiple versions of the subpackages are active at once

image

image

The yarn monorepo solution for this is the `workspaces:*` [protocol](https://stackoverflow.com/a/66701215/5129731) so we can make the package prefer the version in this monorepo, the specific version number would be replaced on the publish step. To use it we would have to upgrade to yarn v2 and up. 

I think that move is worth it, but didn't want to get that tied up in the scope of this PR.

scripts/build-editor-preview.sh
```diff
@@ -0,0 +1,50 @@
...
+#Build - assumes lerna handles the topological sort
+yarn lerna run build
+
+echo "Installing vega editor"
+git clone https://github.com/vega/editor.git
+
+cd editor
+yarn --frozen-lockfile --ignore-scripts
+
+# Make sure we prefer the local version to the one from npm
```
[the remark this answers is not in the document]
hydrosquall: I'd agree this is a hack, ideally we shouldn't need to install these from NPM

Ideally we can remove this after we run through and make sure there's only 1 version of each package ( @lsh mentioned he might get to look into it in the vega slack , I didn't open a new issue for it )

scripts/postinstall.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+# Run this script as long as the env variable
+# DISABLE_POSTINSTALL_SCRIPT is not set.
```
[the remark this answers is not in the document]
hydrosquall: In the CI dashboard config: https://dash.cloudflare.com/<HASH>/pages/view/vega/settings/environment-variables

image

PR #3930, opened 2024-06-11
chore: bump many workspace deps

lsh:
The goal of this PR is to reduce out of date interdependencies in the workspace. I basically just went through and ran `ncu -ui` in each folder, and updated any minor or patch versions. The only project where this didn't work is in `vega-typings`, which has some issues related to the ts-schema-generator that we can look into later (and are semi-unrelated to the goal of this PR anyway).

### DIFF
--- a/packages/vega/package.json
+++ b/packages/vega/package.json
@@ -51,7 +51,7 @@
     "vega-statistics": "~1.9.0",
     "vega-time": "~2.1.1",
     "vega-transforms": "~4.11.1",
-    "vega-typings": "~1.1.0",
+    "vega-typings": "~1.3.0",
     "vega-util": "~1.17.2",
     "vega-view": "~5.12.1",
     "vega-view-transforms": "~4.5.9",
--- a/packages/vega-crossfilter/package.json
+++ b/packages/vega-crossfilter/package.json
@@ -20,9 +20,9 @@
     "prepublishOnly": "yarn test && yarn build"
   },
   "dependencies": {
-    "d3-array": "^3.2.2",
+    "d3-array": "^3.2.4",
     "vega-dataflow": "^5.7.5",
-    "vega-util": "^1.17.1"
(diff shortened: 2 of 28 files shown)


domoritz:
Why not use `ncu -uiws`?

lsh:
`ncu -uiws` didn't seem to bump all of the deps (particularly it didn't bump any `vega-*` package when I tried)

packages/vega/package.json
```diff
@@ -51,7 +51,7 @@
     "vega-statistics": "~1.9.0",
     "vega-time": "~2.1.1",
     "vega-transforms": "~4.11.1",
-    "vega-typings": "~1.1.0",
+    "vega-typings": "~1.3.0",
```
domoritz, line 54: This seems to be the only non automatic version bump.

domoritz (approved):
Looks like these are all within bumps except one.

lsh:
I wonder if `ncu -uiws` didn't work because we use lerna as opposed to NPM workspaces. We should probably migrate to that as a follow up

hydrosquall (commented):
This might work as a first step, but in a follow up, every package that had its version changed, that package will also need to have a new version. (For example vega-functions was at `5.14.0` before this change. Upgrading the versions will cause a version bump for it, maybe 5.14.1.

Unfortunately, that will cause packages that reference vega-functions (like vega-view) to need to bump which copy of vega-function they're importing too.

 The amount of churn here can be limited by running the upgrade  in topological sort order (so that [leaves](https://gist.github.com/hydrosquall/40ece3c6c618fe834a28885986eaa3cd#file-vega-dependency-graph-no-d3-mmd) are bumped first like vega-util, and roots like vega-view get bumped last. 

To keep this already large change lower-risk as possible, we might want to avoid the d3- deps, and just do the the vega- ones in this PR.

lsh:
Closed in favor of #3932

PR #3932, opened 2024-06-12
chore: bump all `vega-*` deps in topological order

lsh:
Take two on #3930. Using @hydrosquall's diagram on the topological sort to try and bump things in order.


domoritz (approved):

hydrosquall (approved):
Looks good to me. 

Thanks for figuring out how to work this out of the [diagram](https://gist.github.com/hydrosquall/40ece3c6c618fe834a28885986eaa3cd#file-vega-dependency-graph-no-d3-mmd) . 

If this becomes a recurring thing we have to do, I'll try to update the mermaid script so that it either gives us a runbook of the order in which to visit the packages, or applies the update + version bumps for us  automatically (with the non vega-* deps excluded).

PR #3936, opened 2024-06-19
fix: bounds calc for empty visualization with absent legend title

### LINKED ISSUE #2881: Empty visualization with null legend title
I created an empty visualization (named dataset) and I wanted to have no legend title. This combination apparently breaks the chart [see here](https://vega.github.io/editor/#/url/vega-lite/N4IgJghgLhIFygHYQLYFN4gGYHscgF8AaEFCAJwGtMAbAS0QxLUQGMcwGBzeUAD17Y6aGmEwCSUAJ4AHDHBBQ0KGTnIQaISXXQBVRHSiZIUwiVMIhIsQtOTZ8kAEcArhERRD0OgDcmICC4ucjQuaEcAZxcUMxB2GjVBLGFRcS0QGlCWG1BPKEz4RBcaGgIygiA)

Is there any other workaround?

betalb:
Some empty visualizations will generate svg element of huge size

Following criteria must be met to trigger bug:

* visualization has legend, but has no legend title (undefined/null)
* has no data
* autosize.type is pad

In this case bounds calculation will create svg element width `width` and `height` attributes equal to `Number.MAX_VALUE`, but runtime value, accessible through `clientWidth` / `clientHeight`, will depend on browser, as they are doing capping (i.e. Firefox caps value at 17_895_698).

Example of vega spec

```json
{
  "autosize": {
    "type": "pad"
  },
  "background": "red",
  "width": 500,
  "height": 500,
  "data": [
    {
      "name": "data",
      "values": []
    }
  ],
  "scales": [
    {
      "name": "color",
      "type": "ordinal",
      "domain": {
        "data": "data",
        "field": "series",
        "sort": true
      },
      "range": "category"
    }
  ],
  "legends": [
    {
      "stroke": "color",
      "symbolType": "stroke"
    }
  ]
}
```

And also vega-lite spec that produces broken vega spec

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v5.json",
  "width": 200,
  "height": 200,
  "data": {
    "name": "data",
    "values": []
  },
  "encoding": {
    "color": {
      "field": "series",
      "legend": {
        "title": ""
      }
    }
  },
  "mark": {
    "type": "line"
  }
}
```

closes #2881

### DIFF
--- a/packages/vega-typings/tests/spec/valid/legends-empty.ts
+++ b/packages/vega-typings/tests/spec/valid/legends-empty.ts
@@ -1,0 +1,35 @@
+import { Spec } from 'vega';
+
+export const spec: Spec = {
+  $schema: 'https://vega.github.io/schema/vega/v5.json',
+  autosize: {
+    type: 'pad'
+  },
+  background: 'red',
+  width: 500,
+  height: 500,
+  data: [
+    {
+      name: 'data',
+      values: []
+    }
+  ],
+  scales: [
+    {
+      name: 'color',
+      type: 'ordinal',
+      domain: {
+        data: 'data',
+        field: 'series',
+        sort: true
+      },
+      range: 'category'
+    }
+  ],
+  legends: [
+    {
+      stroke: 'color',
+      symbolType: 'stroke'
+    }
+  ]
+};
--- a/packages/vega-view-transforms/src/layout/legend.js
+++ b/packages/vega-view-transforms/src/layout/legend.js
@@ -124,6 +124,11 @@ function legendBounds(item, b) {
   // aggregate item bounds
   item.items.forEach(_ => b.union(_.bounds));
 
+  if (b.empty()) {
+    b.x2 = item.padding;
+    b.y2 = item.padding;
+  }
+
   // anchor to legend origin
   b.x1 = item.padding;
   b.y1 = item.padding;
--- a/packages/vega/test/scenegraphs/legends-empty.json
+++ b/packages/vega/test/scenegraphs/legends-empty.json
@@ -1,0 +1,59 @@
+{
+  "marktype": "group",
+  "name": "root",
+  "role": "frame",
+  "interactive": true,
+  "clip": false,
+  "items": [
+    {
+      "items": [
+        {
+          "marktype": "group",
+          "role": "legend",
+          "interactive": false,
+          "clip": false,
+          "items": [
+            {
+              "items": [
+                {
+                  "marktype": "group",
+                  "role": "legend-entry",
+                  "interactive": false,
+                  "clip": false,
+                  "items": [
+                    {
+                      "items": [
+                        {
+                          "marktype": "group",
+                          "role": "scope",
+                          "interactive": false,
+                          "clip": false,
+                          "items": [],
+                          "zindex": 0
+                        }
+                      ],
+                      "x": 0,
+                      "y": 0
+                    }
+                  ],
+                  "zindex": 0
+                }
+              ],
+              "x": 518,
+              "y": 0,
+              "width": 0,
+              "height": 0,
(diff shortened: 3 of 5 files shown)


betalb:
I've updated PR:
 * rebased onto latest main
 * added test

What else is required to push this PR further?

hydrosquall:
Hey @betalb , thanks for looking into this. 

I've have not seen an SVG with scientific notation used for dimensions before
(from pasting your vega spec into the [editor](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAdxoATemgBMABh1L4NMlm16lapJhSpQspAjhoQFq4pDEkghnAhoA2gC6AL5KcLJQbGo0smRooBGCbABOcSAAZjRwgmpOEHBJmb5KguRhOTYgmHQlTiBB9UrISQDWqZgAnjiO6MKyjvVAA) )

image

I can see that your fix passes the spec, I think we will want to test your change in a vega editor as well. 

To enable next steps, I have 2 questions

1. Assuming you have this working locally, could you please share a screenshot of the SVG dimensions in the browser when running Vega Editor with your patched copy of Vega?
2. If the bounds are empty, do you think we should just be avoiding calling `legendBounds` altogether? What would you expect to see in the legend in cases where there is no data?

https://github.com/vega/vega/blob/a07dd38af4b4a6227676c94144efe3895f0ecddd/packages/vega-view-transforms/src/layout/legend.js#L99

betalb:
Here is the screenshot of dimensions from vega editor after the fix
image

Until you traverse group hierarchy, you won't know if legend bounds are really empty, and this is done as 1st step in legendBounds function https://github.com/vega/vega/blob/560aeecbbe99edb3022071fbe8c756d04f2e57bc/packages/vega-view-transforms/src/layout/legend.js#L125

And regarding the question "What would you expect to see in the legend in cases where there is no data?", I'm not sure. I have 2 options on my mind:
 1. Produce markup for svg, include configured paddings and render it, even though visually it will be empty and just eating a little bit of pixels. This is the how current fix works
 2. Exclude legend completely from scenegraph if it is "logically" empty. This will produce svg with smaller dimensions

In 2nd case below spec (with empty title)
```
{
  "$schema": "https://vega.github.io/schema/vega-lite/v5.json",
  "width": 200,
  "height": 200,
  "data": {"name": "data", "values": []},
  "encoding": {"color": {"field": "series", "legend": {"title": ""}}},
  "mark": {"type": "line"}
}
```

will produce identical result to
```
{
  "$schema": "https://vega.github.io/schema/vega-lite/v5.json",
  "width": 200,
  "height": 200,
  "data": {"name": "data", "values": []},
  "encoding": {},
  "mark": {"type": "line"}
}
```
image

hydrosquall:
@betalb thanks for the clear explanation and editor screenshot 👍. I hadn't noticed earlier that legendBounds mutates `b`, which is why we had to enter legendBounds to know that b was empty. 

> Exclude legend completely from scenegraph if it is "logically" empty. This will produce svg with smaller dimensions

This mode makes sense to me, do you have a sense for what the diff would look like if we took the current approach? Would it be a matter of not pushing a legend when there's no data?

https://github.com/vega/vega/blob/a3af49e9d7004895d1c06417718611106dcb5418/packages/vega-view-transforms/src/ViewLayout.js#L70

It looks like the current approach prevents the infinitely sized svg, but it's a bit unexpected from reading the code why we have this line. If we keep it, can we add a comment explaining why `b.empty()` needs those item.padding values for `x2/y2` (as opposed to returning early?)

PR #3937, opened 2024-06-28
docs: Fix broken link in global-development.md example

dsmedia:
Update the broken link in the global-development.md file.


joelostblom (approved):
Thanks for updating the link!

PR #3955, opened 2024-08-02
docs: Update packed bubble example

PBI-David:
Update the packed bubble chart to use a square root scale so as not to be misrepresentative.

### DIFF
--- a/docs/examples/packed-bubble-chart.md
+++ b/docs/examples/packed-bubble-chart.md
@@ -8,4 +8,6 @@ image: /examples/img/packed-bubble-chart.png
 
 A packed bubble chart displays relatively sized circles arbitrarily packed together. You can change the shape of the packing by adjusting the x or y gravity. This example shows dummy categorical data where node areas are proportional to the amount value.
 
+This Vega example made by David Bacci [@PBI-David](https://github.com/PBI-David).
+
 {% include example spec=page.spec %}
--- a/docs/examples/packed-bubble-chart.vg.json
+++ b/docs/examples/packed-bubble-chart.vg.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://vega.github.io/schema/vega/v5.json",
-  "description": "Warming Stripes Chart. Originally designed by Ed Hawkins.",
+  "description": "A packed bubble chart.",
   "usermeta": {
     "author": "David Bacci",
     "github": "https://github.com/PBI-David/Deneb-Showcase"
--- a/docs/examples/packed-bubble-chart.vg.json
+++ b/docs/examples/packed-bubble-chart.vg.json
@@ -1,5 +1,10 @@
 {
   "$schema": "https://vega.github.io/schema/vega/v5.json",
+  "description": "Warming Stripes Chart. Originally designed by Ed Hawkins.",
+  "usermeta": {
+    "author": "David Bacci",
+    "github": "https://github.com/PBI-David/Deneb-Showcase"
+  },
   "width": 800,
   "height": 500,
   "padding": {"left": 5, "right": 5, "top": 20, "bottom": 0},
@@ -44,8 +49,9 @@
   "scales": [
     {
       "name": "size",
+      "type": "sqrt",
       "domain": {"data": "table", "field": "amount"},
-      "range": [100, 3000]
+      "range": [0, 5000]
     },
     {
       "name": "color",
--- a/docs/examples/packed-bubble-chart.vg.json
+++ b/docs/examples/packed-bubble-chart.vg.json
@@ -49,9 +49,8 @@
   "scales": [
     {
       "name": "size",
-      "type": "sqrt",
       "domain": {"data": "table", "field": "amount"},
-      "range": [0, 5000]
+      "range": [0, 10000]
     },
     {
       "name": "color",
@@ -72,7 +71,7 @@
           "yfocus": {"signal": "cy"}
         },
         "update": {
-          "size": {"signal": "pow(2 * datum.amount, 2)", "scale": "size"},
+          "size": {"signal": "datum.amount", "scale": "size"},
           "stroke": {"value": "white"},
           "strokeWidth": {"value": 1},
           "tooltip": {"signal": "datum"}
@@ -87,7 +86,7 @@
             {
               "force": "collide",
               "iterations": 2,
-              "radius": {"expr": "sqrt(datum.size) / 2"}
+              "radius": {"expr": "sqrt(datum.size)/2"}
             },
             {"force": "center", "x": {"signal": "cx"}, "y": {"signal": "cy"}},
             {"force": "x", "x": "xfocus", "strength": {"signal": "gravityX"}},


domoritz:
Unfortunately neither a sqrt nor a linear scale is really representative. People don't interpret size by either.

PBI-David:
When comparing circles, I thought people compared areas which is why I thought a sqrt scale should be used. Every dataviz tool that has a bubble plot that I'm aware of use a square root scale otherwise the proportions are way off (a value twice as large is not represented by a circle twice the size). I don't mind either way but thought this would be more aligned with what I see elsewhere.

PBI-David:
Couple of links to illustrate the point in d3: 

https://benclinkinbeard.com/d3tips/log-and-power-scales/


https://observablehq.com/@johnhaldeman/tutorial-on-d3-basics-and-circle-packing-heirarchical-bubb

PBI-David:
Alternatively, I could add a signal that allows a user to switch between the scales which could help illustrate the choice people need to make.

PBI-David:
Here is a screenshot to illustrate the issue. The smaller circle has a value of 1 and the larger circle has a value of 2. 

Using a linear scale ends up with this (radius is doubled):

image

image

Using a squareroot scale ends up with area being doubled:

image

image

domoritz:
Shouldn't we change this line instead: https://github.com/vega/vega/pull/3955/files#diff-e4577745914d531c6c6fe9b7d252369187ffabb2e1ae1b3d8481d7bb07771fe3R75

PBI-David:
@domoritz  possibly. Although if I simplify to just the field value, I can't get the areas or diameters to make sense using linear or sqrt scales. I'm probably missing something obvious though. What do you think it should be changed to? Keeping it as proposed but changing just the scale allows:

linear = double diameter
sqrt = double area

domoritz:
How about [Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykSArJQBWENgDsQAGhAB3GgBN6aABwAGNTPg0yWNAI0ycSBQpriyaUABs4AM0z6ZAJx17UAmZjY40AJk0QJjZsNgQ0NQBfGSQGbwgaAC84NBBxCRSZBLJxJGsINABtUFyEFPQoAA9pEAYcBSRMcrlFegACAja-EGiSpDLUqABPGrqGptTtXUwOrp6pPoH0MmckYjohgA0a4jyGcrVKPxkmcwUrEHMcONTVi0yQBHMImWRq1ABGSN60-uaVtYbACaOz2B0oHxOZwuVxu6DuZAeT0kqECbzQX0iAF0ZOMUKhir8liBMEgmLZQdZ9gUCaAoI1yGxnCN0ABBGr9NgMcSOVFHFQ-elNMhMlkgABCHIQXJ5EUoAgEgoZIuZqQAwlKZbzDgAWADMSuFotSABFNdztZQAJxfBYgIWM1XoACi5tlfJUtrpyuN6AAYm7LQIDXaHSqxQBxQNyj5Ww2OsUACWjHoA7PHw6kAJIpw5+AWhn1OkAAKVz8sVhaNxYA0uX9RnfSAADLlm2N4sAWXLno7YoAcuXg33UgB5cuxkfoAAKPfTWOxWXpthphNKzQSyRqCjCSGeqFAeNSpPJDzsNDg1nO6E5FvmIAR5UKmgCGgXdvXgzY1iZNUwQxwZomTMXJrG3Xd90PRp8RJMkKRkc9L2ve0ixGaIHyQe5bn6XxF0eJBnAAa1XRZmnSBQ4AKLwAI3IYEGCMCEOcMILiPdATwpdC4HEKA2Aoi5uKaZwLnPawwIPEBoDyZpeJ-YSEIvK9BlQ+9KjsNgoAYGlQGyUDBmqdChnUzTtMknQ9IqND0LGBkLk3codPMvJUnGBgEEoW9ZSXaTUns+8IEwZjCIckBdipZpZBgOgUnQgKgrgAB1VoYAuML9gxdDvG-TAaF8CTdOc9BXPCb5MruCB1OccJaRJGjUkqqAHmi1YcokGkPkMSTSRyqA0DsPIIDgBCmUakiQAamTv2sRQmqExp2HEGljgwswtIEyocGE9AIAAR2cTAAApisoeyAEoCG6b47QmwZBLgeSQHeRyckK+0DJkFlnos+0rOukbmmqGR3ke4y1qyQLuLIZR8qc8SQABdZ-22K7QBu9ARg+1IjI0sGuucSHoa+16EeBHpsXJu1-0A484EqRwmJYiS2LSPjKPvbjeP4iS7q20A8nM1KwVunl7vvJgkEG6bxBCtLmieUxOOGnkAGUkhloXPkrcaJEwBK4DcXlQFl1IGPOdDRPEo2NbkKKJky2nDfGxTkOOsNjSu2p6lsiSnqdpDUnekBPr9pT0bJ74FyAA)

PBI-David:
@domoritz  looks good to me - I have updated it. The areas are slightly off but this is probably down to my screen ruler rounding to integers.

domoritz:
Thanks!

domoritz (approved):

kanitw:
> PS I did create the original example.

Gotcha. Thanks for contributing!

PBI-David:
> > PS I did create the original example.
> 
> Gotcha. Thanks for contributing!

Thanks for creating such a versatile and powerful dataviz library 😉

BTW, I have a lot advanced examples on my GitHub that I can create PRs for if you think they might be useful. The sankey and gantt charts have been **very** popular.

domoritz:
That's awesome. I wonder how we can maybe make it easier to have community examples that don't have to be part of this repo. We could do something similar to Vega-Lites listing but maybe with screenshots to make it easy to go to them. 

Thoughts?

PBI-David:
That makes sense to me. It would be easier to maintain and contribute as I see great Vega examples all the time that most are unaware of. FYI, Vega/VL have become very popular in the Power BI space via a container called Deneb. There is growing demand from that user base for a grammar of graphics approach to dataviz. I know the gantt chart is used extensively at some very high profile companies and organisations.

[Gantt](https://evyb.short.gy/nCRa3r)

domoritz:
That's super cool. I've previously thought about ways to let people add examples to an index in the Vega editor but it became too much work to develop/maintain. A community example section here similar to https://vega.github.io/vega-lite/examples/#community-examples but with images (as suggested in https://github.com/vega/vega-lite/issues/3120) would be pretty awesome. Would you like to take a stab at it? We could copy the approach in Vega-Lite as well.

PBI-David:
Happy to have a go but I don't have an environment so everything would need to be done via the web. If it is just a new header and a list of links, then I can submit a PR for that.

domoritz:
We can start with a list just like Vega-Lite.

PBI-David:
Hmm - just taken a look and I'm not even sure I can do this simple change without a working environment to test. VL and Vega documentation use different systems so I can't just copy it across and hope for the best. I believe this is the correct page:

https://github.com/PBI-David/vega/blob/main/docs/_layouts/example.html

Can I just add raw HTML as a bulleted list and heading? Will it be styled correctly?

Even though it is a simple addition, this probably needs to be done by someone who can preview how the changes will appear. Once the skeleton is in place, I can add a few examples I have seen on the web. 

In parallel, I'll try and get a local environment running again but I hit too many blockers on Windows last time and just gave up.

domoritz:
The setup probably works in the subsystem for Linux.

hydrosquall:
I think it might be possible to avoid the need for a local build system if
we deployed a build of the docs in addition to the editor (as discussed
here
https://github.com/vega/vega-lite/issues/9276#issuecomment-2040896404 )

In the meantime, I think the core of the change could be submitted as a
block of markdown in this thread where the wording / links can be reviewed
without any build system.

 Any of the maintainers can assist with moving it into the appropriate Md /
html file. I am happy to help with this in the absence of someone else
volunteering, as I spent quite a bit of time getting Ruby etc to work when
I was reformatting the site for a custom docs plugin last year.

Issue #3954, opened 2024-07-31
fix: Update `LegendType` typescript definition to support 3rd (discrete) type


hydrosquall:
## Motivation

The source code handles 3 types of legend at runtime

https://github.com/vega/vega/blob/ddd2bbff9897415be968fb60e2afa6c695c9eec4/packages/vega-parser/src/parsers/legend.js#L35-L36

https://github.com/vega/vega/blob/ddd2bbff9897415be968fb60e2afa6c695c9eec4/packages/vega-parser/src/parsers/legend.js#L80-L81

The types definitions only permit 2 types of legend at compile time.

https://github.com/vega/vega/blob/8a7f2f233b7fec430e52fbe19a544c1f290be591/packages/vega-typings/types/spec/legend.d.ts#L43

This mismatch contributed to a bug like https://github.com/vega/vega-lite/issues/9399

## Proposal

- We update `LegendType` to support the third type of legend, `discrete` (implicitly discrete gradient vs continuous gradient). Note this may also impact 1 or more [docstrings](https://github.com/vega/vega/blob/8a7f2f233b7fec430e52fbe19a544c1f290be591/packages/vega-typings/types/spec/legend.d.ts#L65-L67) .

https://github.com/vega/vega/blob/8a7f2f233b7fec430e52fbe19a544c1f290be591/packages/vega-parser/src/parsers/guides/constants.js#L27

PR #3957, opened 2024-08-02
fix(legend): add third mode, `discrete`, to `LegendType`

hydrosquall:
## Motivation

- Fix https://github.com/vega/vega/issues/3954

## Changes

- Adds a type to align with the runtime behavior, + docstring update

https://github.com/vega/vega/blob/ddd2bbff9897415be968fb60e2afa6c695c9eec4/packages/vega-parser/src/parsers/legend.js#L35-L36

- [x] (Checked w/ @jheer to confirm if this is the intended behavior based on the commit that introduced this, https://github.com/vega/vega/commit/ddd2bbff9897415be968fb60e2afa6c695c9eec4

## Notes

- Broader question: would we be interested to have portions of the vega sub-packages converted to TS so we don't have to sync `.d.ts` files with JS source code as a separate step?

### DIFF
--- a/docs/docs/legends.md
+++ b/docs/docs/legends.md
@@ -13,7 +13,7 @@ Properties for specifying a legend. Legends accept one or more [scales](../scale
 
 | Property      | Type                           | Description    |
 | :------------ |:------------------------------:| :------------- |
-| type          | {% include type t="String" %}  | The type of legend to include. One of `symbol` for discrete symbol legends, or `gradient` for a continuous color gradient. If `gradient` is used only the _fill_ or _stroke_ scale parameters are considered. If unspecified, the _type_ will be inferred based on the scale parameters used and their backing scale types.|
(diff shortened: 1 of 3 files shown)


domoritz:
> Broader question: would we be interested to have portions of the vega sub-packages converted to TS so we don't have to sync .d.ts files with JS source code as a separate step?

Yes, if we can figure out a way that doesn't degrade the developer experience. So for example if I change code in a linked Vega package, the Vega editor (in local dev mode) should update.

domoritz (changes requested):
Please update the docs at https://vega.github.io/vega/docs/legends/ as well.

hydrosquall:
> Yes, if we can figure out a way that doesn't degrade the developer experience. So for example if I change code in a linked Vega package, the Vega editor (in local dev mode) should update.

Got it! I split this out to a separate issue. https://github.com/vega/vega/issues/3971

> Please update the docs

A separate followup would be to generate sections of the docs files (like this props table) from the JSDoc comments in the source code. For the time being, I made an manual update, thanks for the flagging this one.

domoritz (approved):

PR #3962, opened 2024-08-26
feat: add sort expressions

joelostblom:
I think it would be useful to have sort function in the expression language. My immediate use case is for https://github.com/vega/altair/pull/3394#pullrequestreview-2231115960, where we need to sort an array returned by `domain()`.

I'm not at all sure what the best implementation is here, or what tests are needed, so this PR is partly to move the discussion I started with @hydrosquall on slack and get additional input. I think that ideally there would just be one `sort` function that handles both numbers, strings, and mixed types, but that seems considerably more complex and I'm not comfortable writing it so I'm suggesting these simple ones which @hydrosquall showed me on slack and that directly maps to the built in JS array methods.

### DIFF
--- a/docs/docs/expressions.md
+++ b/docs/docs/expressions.md
@@ -522,6 +522,10 @@ Returns an array containing an arithmetic sequence of numbers. If _step_ is omit
 <b>slice</b>(<i>array</i>, <i>start</i>[, <i>end</i>])<br/>
 Returns a section of _array_ between the _start_ and _end_ indices. If the _end_ argument is negative, it is treated as an offset from the end of the array (_length(array) + end_).
 
+<a name="sort" href="#sort">#</a>
+<b>sort</b>(<i>array</i>)<br/>
+Sorts the array in natural order using [ascending from Vega Utils](https://vega.github.io/vega/docs/api/util/#ascending).
+
 <a name="span" href="#span">#</a>
 <b>span</b>(<i>array</i>)<br/>
 Returns the span of _array_: the difference between the last and first elements, or _array[array.length-1] - array[0]_.
--- a/packages/vega-expression/README.md
+++ b/packages/vega-expression/README.md
@@ -128,6 +128,7 @@ This package provides the following constants and functions:
 - [`lastindexof`](https://vega.github.io/vega/docs/expressions/#lastindexof)
 - [`reverse`](https://vega.github.io/vega/docs/expressions/#reverse)
 - [`slice`](https://vega.github.io/vega/docs/expressions/#slice)
+- [`sort`](https://vega.github.io/vega/docs/expressions/#sort)
 
 **String Functions**
 
--- a/packages/vega-interpreter/package.json
+++ b/packages/vega-interpreter/package.json
@@ -22,6 +22,9 @@
     "test": "TZ=America/Los_Angeles tape 'test/**/*-test.js'",
     "prepublishOnly": "yarn test && yarn build"
   },
+  "dependencies": {
+    "vega-util": "^1.17.2"
+  },
   "devDependencies": {
     "vega": "*"
   }
--- a/packages/vega-interpreter/src/functions.js
+++ b/packages/vega-interpreter/src/functions.js
@@ -1,3 +1,5 @@
+import { ascending } from "vega-util";
+
 const slice = Array.prototype.slice;
 
 const apply = (m, args, cast) => {
@@ -62,8 +64,7 @@ export default {
   lastindexof:  function() { return apply('lastIndexOf', arguments); },
   slice:        function() { return apply('slice', arguments); },
   reverse:      x => x.slice().reverse(),
-  sort:         x => x.slice().sort(),
-  sortNumeric:  x => x.slice().sort((a, b) => a - b),
+  sort:         x => x.slice().sort(ascending),
 
   // string functions
   parseFloat:   parseFloat,
--- a/packages/vega-interpreter/src/functions.js
+++ b/packages/vega-interpreter/src/functions.js
@@ -1,4 +1,4 @@
-import { ascending } from "vega-util";
+import { ascending } from 'vega-util';
 
 const slice = Array.prototype.slice;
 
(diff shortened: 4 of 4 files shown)


domoritz:
Thanks. I added a more generic comparator so we don't need two methods and added docs as well.

joelostblom:
That's great, thank you! Having just one method is much preferred.

hydrosquall:
Hey @joelostblom, thanks for putting this together!

Using the branch preview ( https://sort-functions.vega-628.pages.dev/#/ ), could you link to a sample Vega or Vega-Lite definition that would make use of this new expression? We probably want at least 1 case each for each happy path (numbers, strings, maybe one with missing data). I'm not sure we need to have an especially "nice" behavior for the mixed datatype case other than to not crash. 

> what tests are needed, 

I looked to see how tests in this package `vega-interpreter` are written ( https://github.com/vega/vega/tree/main/packages/vega-interpreter/test ), and found that it appears to involve adding an additional JSON file each to to https://github.com/vega/vega/tree/main/packages/vega/test/specs-valid and https://github.com/vega/vega/tree/main/packages/vega/test/scenegraphs . Once we confirm that this expression does what you need in the preview deploy, we can copy that JSON over to the test suite.

joelostblom:
Sorry for the delay in getting back to you @hydrosquall . I was trying out the branch preview editor right now, and it seems like it doesn't recognize the `sort` function. Try replacing `reverse` with `sort` on line 15 in the following spec and you will see that it errors with `Unrecognized function: sort`. [Open the Chart in the Vega Editor](https://sort-functions.vega-628.pages.dev/#/url/vega/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykSArJQBWENgDsQAGhBMkUANZkATmwbiAJmhAB3GHTjSQOJBo01xZNAJk6aG+mgBMABhkRMATwA2h9FDhvbyMNJEwUVABtUHEkBD8QMQZlAIB9FyNiJG8GOAg0aJAIkABBIyZnAA4AXwBdaqkYuITQ8PSjJJSEzrSMmUxlJHEIADM2ZQQC0C8cBLGJhm8UGTgADxxlbWU4UmUIOAAKSIByEeOpAAJjpHOrzFvjpmPagEojJHz0TDzMEDq69w0MixbyfQqxeLaTypDxwHCZbK5ZwuBpNSHoeBArBGBg4VoJOSaCAmAIHDRsZAWA7HTzHF6UXyWeiXACMlxclAELwuACoLtDYfCASBkMoFGC0QlReKjDMEt9Vr93F5fAUQArfrUZCNVJNUKBWsVDe0GiA4OIoGwNH5QLj8WhQCMaEEHSAsjkCUtFH8ZNboMoaDhMOxJPrEkCQdoADrqn6oC4xi4AaguBxoEAAatl7GSwgwEJEY98PDHXhcAPwXVr5wuxksgWoXeMxmNJ6sFos-UtvU2rV0QCPZbR2BwwIwIRa-VAcgSmzz9qDZBLz7U0QJaL4-H1FbwR13upH+c3fTamuT7XfiG1uxFS+waVXVJ-C6BLiUgCHL2WeWbaHBsCwlRAclKVDA0wiNCD2lXddtGLICxGUKcBlyU1Bksa9BX7QdgnQAVviFU0TDMCwrGnTlhSQVY8imRJF1VPCjHGNdxCnEBfBGICVHsNARmyfZ+joBi6yAgAvCxrT7ad-iAA)

domoritz:
I tried this spec btw. It definitely worked for me before but I also get `Unrecognized function: sort` right now. 

```
{


  "signals": [
    {
      "name": "foo",
      "update": "sort([1,2,3])"
    }
  ]
}

```

hydrosquall:
@joelostblom - I’m puzzled about how the other expressions for sequences
are working given that they’re only listed in one place. This is my first
time seeing a function added to the core lib rather than adding via the
extensions API.

From grepping the code for “unrecognized”, I think we might have to add
this new sort function to Vega-expression too. This package is used when
Vega is used in strict mode to be compliant with CSP.

https://github.com/vega/vega/blob/main/packages/vega-expression/src/functions.js

joelostblom:
> From grepping the code for “unrecognized”, I think we might have to add
this new sort function to Vega-expression too. 

Hmm, doesn't that sound odd given that other array functions such as `reverse` are not listed there?

Give that @domoritz mentioned the spec worked for him before, maybe there is something going on with the deploy preview?

domoritz:
I couldn't get it to work locally either yesterday. Need to dig a bit more what I did before

hydrosquall:
From what I can see, the editor doesn’t use the strict interpreter, it uses
the faster default “Function” constructor

https://vega.github.io/vega/usage/interpreter/

https://github.com/vega/editor/blob/master/src/components/renderer/renderer.tsx#L168-L171

hydrosquall:
Reverse is also defined and exported from the Vega-functions package -
maybe we have to add it to the shared function context object.

https://github.com/vega/vega/blob/main/packages/vega-functions/src/codegen.js#L236

(What I’m taking away from this PR is we may want to have a guide to adding
new functions for contributors, if doing so involves changing files in
multiple packages)

hydrosquall:
I can't speak to why the relevant code is missing from that localhost screenshot (that does seem odd), but adding the method to `codegen.js` seems to have successfully added `sort` to the editor preview. (see the test link on https://github.com/vega/vega/pull/3973 ). 

image

From revisiting the package descriptions, it looks like `vega-interpreter` is only used for the CSP compliant mode. Since the editor doesn't run in CSP mode, it didn't know about this function. `vega-expression` is used in "regular" Function constructor mode, it seems reasonable to me that if we only edited one, it should probably be the one under `codegen.js`. 

(Apologies @joelostblom , as I steered you incorrectly when I suggested `vega-interpreter` as the package to change!)

domoritz:
Moved to https://github.com/vega/vega/pull/3973

Issue #2181, opened 2019-11-21
histogram counts/labeling is surprising when nice=false


zanarmstrong:
If I create data that is simply the integers from 0 to 29 (including 0 and 29), I would expect a histogram to look something like this:

[histogram](https://vega.github.io/editor/#/url/vega-lite/N4KABGBEAkDODGALApgWwIaQFxUQFzwAdYsB6UgN2QHN0A6agSz0QFcAjOxge1IRQyUa6ALQAbZskoBmOgBY6ABjoArWNwB2kADTgo8TQDNG1bGFAQIkDACcA1mYuWrebtzF5GhMxtZixehAAvrrOkBSMyADujoGWkCgm+GbSioqhzlBRjAAmLGZyaXFBeiF6kDnoeJg4TlYa6KjIZhVVogBs7QBMAIwAHADsPcgD7IbsAJzdAKwDOXIT0-ATE12Kcj1d7YqYpRmt1bDIeLCxYZXVIp29g8Oj41Nds-OLy6vrm9s1YADacRB1TKQAAeZkU-zAZUy5ghVlBOB6EKhmUBYXhYC6SIyzlR8XR0ixENxcIKhOhxKg6OmZJRsMpZnaNJxdJBZgGTMsFNZOD6HIBLPREz5MOhJIR4OhyOZovpCMRkuxnIFZk2wq56J6BIVROVCLkat1YB61O15MNPUZptpMu5RvZVulott-QNNo1QodSrdZjWrqd6N6fuhtq6mM9-O9OC6WsyUq9-p9+vDIoTUZNscVEdTGMtGZ1kYx9rzZoLXV5yfVPo9sbiAF09uVkBoDDlGBpTLU4rbcZB2G2zk6NIx4M0cHgbKxkMLIMZkGIci1QZmoHgAJ6EUdQACOrHQGk81U8VEgxUVkFXA7C6Go1Bswjwm8gBlY+50EMga43LR3e4PVUYx7FA2Vi2A4OC9ugNgnkEQA)

![nice](https://user-images.githubusercontent.com/7397271/69289067-37690300-0bb0-11ea-8fee-ccc209bfe265.png)

This is what I see what nice=true or nice is not specified. 

[...]

PR #3967, opened 2024-09-05
fix: return adjusted stop from bin transform

### LINKED ISSUE #3966: results of binning outside computed domain
Issue was originally raised on SO: [Issue with Vega Altair Heatmap: Cell Appearing Outside the Heatmap](https://stackoverflow.com/questions/78925999/issue-with-vega-altair-heatmap-cell-appearing-outside-the-heatmap).

A (very) minimal working example that showcase the problem is the following ([Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUATAAZKARkoArCGwB2IADQgAJkkwpUoYkkEM4ENAG1QADzTSAvhttp+syWIcBdJyGQAnAGs0UEwATxw4NBB-OChMTRBMNjZBTBocNEx-M184VSg2bRpVMhCQO0MQJhLy5Bsa1QtUAGZfADMaOEFtaLsHAaA)):
```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v5.20.1.json",
  "data": {"values": [{"x": 0}, {"x": -0.25}]},
  "mark": {"type": "rect", "tooltip": true},
  "encoding": {"x": {"bin": {"maxbins": 3}, "field": "x"}}
}
```

The results of binning is outside the computed domain of the axis.

domoritz:
Fixes #3966. It's not entirely correct since 0 should be in the bin before 0 but that's a floating point issue with a difficult underlying issue. This fix at least doesn't make bins stick out of the bounds anymore. 

Screenshot 2024-09-05 at 16 47 20

Improved example from https://github.com/vega/vega/issues/2181
Screenshot 2024-09-05 at 16 52 48

### DIFF
--- a/packages/vega-statistics/test/bin-test.js
+++ b/packages/vega-statistics/test/bin-test.js
@@ -77,15 +77,6 @@ tape('bin generates boundaries with minimum step size', t => {
   t.end();
 });
 
-tape('bin generates boundaries around the data', t => {
-  const b = bin({extent:[-0.25, 0], maxbins:3});
-  t.equal(b.start, -0.30000000000000004);  // floating point issues can come from this
-  t.equal(b.stop, 0);
-  t.equal(b.step, 0.1);
-
-  t.end();
-});
-
 tape('bin generates boundaries for given span size', t => {
   let b = bin({extent:[0, 100], span:10, maxbins:10, nice:false});
   t.equal(b.start, 0);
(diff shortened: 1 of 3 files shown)


lsh (approved):
Thanks for the fix!

domoritz:
@jheer this is undoing your change from https://github.com/vega/vega/pull/2182/files. Do you recall the reasoning for returning the unadjusted stop?

PR #3973, opened 2024-09-21
feat: add sort function to `vega-functions` (and `vega-interpreter`)

hydrosquall:
## Motivation

This PR tests a [fix](https://github.com/vega/vega/pull/3962#issuecomment-2362682733) for https://github.com/vega/vega/pull/3962 , with the goal of adding a `sort` function to the default set of Vega functions in the expression language.

## Changes

- Includes @domoritz and @joelostblom's doc and `vega-interpreter` commits from https://github.com/vega/vega/pull/3962, rebased against latest `main`
- Added + tests sort in `vega-functions`: https://github.com/vega/vega/pull/3973/commits/5e6190d84a0fb9aff2a33b0ca7658b5b5f4208b2#diff-328f6311e6ab7be84b1167d9ca245bea77f0ae6e078ffcd4b71bb7b2340e810b

## Testing

- This worked locally when using `yarn link vega`
- This is also working in the [Cloudflare preview](https://cameron-yick-sort-functions.vega-628.pages.dev/#/url/vega/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykSArJQBWENgDsQAGhBMkUANZkATmwbiAJmhAB3GHTjSQOJBo01xZNAJk6aG+mgBMABhkRMATwA2h9FDhvbyMNJEwUVABtUHEkBD8QMQZlAIB9FyNiJG8GOAg0aJAIkABBIyZnAA4AXwBdaqkYuITQ8PSjJJSEzrSMmUxlJHEIADM2ZQQC0C8cBLGJhm8UGTgADxxlbTFlTAAKSIByEYOpAAIDpBPzzCuDpgPagEojJHz0TDzMEDq69xoyWLeN6FWLxbSeVIeOA4TLZXLOFwNJpg9Dwf5YIwMHCtBJyTQQEwBXYaNjICy7A6eA6PSi+Sz0M4ARjOLkoAkepwAVKcIVCYb8QMhlApgciEkKRUYZgkPqsvu4vL4CiBZV9ajIRqpJqhQK1inr2g0QHBxFA2Bo-KAsTi0KARjQgraQFkcrilopvjILdBlDQcJh2JIdYl-oDtAAdFWfVCnSOnADUp12NAgADVsvZiWEGAhIpGPh5I09TgB+U6tHN5qOFkC1U4xyOR+MV3P5z5F55G1ZOiCh7LaOwOGBGBCLL6oVkCI2eHtQbIJGcamiBLTvT6eoreUNOl3w-wmj6bI1yCCBCyW51w8X2DRK6r3gXQeeikCghdSzyzbQ4NgWeUgEkySDXUwn1UD2iXFdtALf9tnHAZciNQZLAvPkez7YJ0F5D5+SNEwzAsKwJzZAUkFWPIpkSOclSwoxxmXcRxxAXwRn-FR7DQEZslPfo6Bo6t-wALwsC1uwnH4gA) 

Uses test spec JSON:

```
{
  "background": "white",
  "padding": 5,
  "width": 20,
  "style": "cell",
  "data": [
    {"name": "source_0", "values": [{"a": "A", "b": 28}]},
    {
      "name": "data_0",
      "source": "source_0",
      "transform": [
        {"type": "formula", "expr": "sort(['f', 'a', 't', 'b'])", "as": "test"}
      ]
    }
  ],
  "signals": [
    {"name": "y_step", "value": 20},
    {
      "name": "height",
      "update": "bandspace(domain('y').length, 1, 0.5) * y_step"
    }
  ],
  "marks": [
   
  ],
  "scales": [
    {
      "name": "y",
      "type": "point",
      "domain": {"data": "data_0", "field": "test", "sort": true},
      "range": {"step": {"signal": "y_step"}},
      "padding": 0.5
    }
  ],
  "axes": [
    {
      "scale": "y",
      "orient": "left",
      "grid": false,
      "title": "test",
      "zindex": 0
    }
  ]
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,5 +59,8 @@
   "engines": {
     "node": ">=18.18.0"
   },
-  "packageManager": "yarn@1.22.19"
+  "packageManager": "yarn@1.22.19",
+  "volta": {
+    "node": "20.17.0"
+  }
 }
--- a/package.json
+++ b/package.json
@@ -59,8 +59,5 @@
   "engines": {
     "node": ">=18.18.0"
   },
-  "packageManager": "yarn@1.22.19",
-  "volta": {
-    "node": "20.17.0"
-  }
+  "packageManager": "yarn@1.22.19"
 }
--- a/docs/docs/expressions.md
+++ b/docs/docs/expressions.md
@@ -522,6 +522,10 @@ Returns an array containing an arithmetic sequence of numbers. If _step_ is omit
 <b>slice</b>(<i>array</i>, <i>start</i>[, <i>end</i>])<br/>
 Returns a section of _array_ between the _start_ and _end_ indices. If the _end_ argument is negative, it is treated as an offset from the end of the array (_length(array) + end_).
 
+<a name="sort" href="#sort">#</a>
+<b>sort</b>(<i>array</i>)<br/>
+Sorts the array in natural order using [ascending from Vega Utils](https://vega.github.io/vega/docs/api/util/#ascending).
+
 <a name="span" href="#span">#</a>
 <b>span</b>(<i>array</i>)<br/>
 Returns the span of _array_: the difference between the last and first elements, or _array[array.length-1] - array[0]_.
--- a/packages/vega-expression/README.md
+++ b/packages/vega-expression/README.md
@@ -128,6 +128,7 @@ This package provides the following constants and functions:
 - [`lastindexof`](https://vega.github.io/vega/docs/expressions/#lastindexof)
 - [`reverse`](https://vega.github.io/vega/docs/expressions/#reverse)
 - [`slice`](https://vega.github.io/vega/docs/expressions/#slice)
+- [`sort`](https://vega.github.io/vega/docs/expressions/#sort)
 
 **String Functions**
 
--- a/packages/vega-functions/index.js
+++ b/packages/vega-functions/index.js
@@ -65,7 +65,8 @@ export {
   lastindexof,
   replace,
   reverse,
-  slice
+  slice,
+  sort
 } from './src/functions/sequence';
 
 export {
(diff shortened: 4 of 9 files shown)


joelostblom:
This is great, thank you @hydrosquall !

One thing that I noticed is that when there is a mix of numbers and strings, it seems like this function sorts them as subarrays: first it sorts all the string until the first number, then all the numbers, then all the string after the last number:

image

I would expect the output to be `["0", "1", "2", "3", "4", "a", "b"]`. But to be honest, I don't think this is a show stopper; it would be rare that anyone passes in an array with mixed object types and it is easy to just pass in all strings instead.

hydrosquall:
Thanks for having a close look @joelostblom , it's cool to see how those subarrays are handled in Javascript sorting!

I'm not sure what to expect if we mix datatypes beyond the known sentinel values like `null` and `undefined`. The sorting would likely look strange if we mixed `boolean` and object ( `{name : "hello"}` ) values into the array too. I think the current behavior of not having sort coerce datatypes is preferred. If the user needed their `1` to be treated as strings they have the option of coercing them to strings before the sort, or vice versa (parsing string as int).

package.json
```diff
@@ -59,5 +59,8 @@
   "engines": {
     "node": ">=18.18.0"
   },
-  "packageManager": "yarn@1.22.19"
+  "packageManager": "yarn@1.22.19",
+  "volta": {
+    "node": "20.17.0"
+  }
```
domoritz, line 65: Let's not include this in this pull request. 
```suggestion
  "packageManager": "yarn@1.22.19"
```

domoritz (approved):
Thanks. Looks great.

joelostblom:
Thank you both!

package.json
hydrosquall: Oops thanks for tidying this for me. I need to figure out how to set node version for just this folder without modifying the package json.

Issue #3475, opened 2022-04-19
Mouse signals confused by transform: scale(...)


isaksky:
In our app we have a full-screen mode where we want the charts to be made bigger. In order to do this, we use `transform: scale(...)`. The problem we are having is that the scale makes it so the mouse signals no longer track correctly. The same happens in [this example](https://vega.github.io/vega/examples/stock-index-chart/), when you add a scale.

Here is what it looks like normally:

![image](https://user-images.githubusercontent.com/607986/164113063-112cc5ea-4d5e-43c6-80c7-d4059ea3683d.png)

But if you add a `transform: scale(...)`, the tracking is off:

![image](https://user-images.githubusercontent.com/607986/164113482-f5f6bc1f-008c-4e51-9a14-6e914a1c61aa.png)

(The position of the chart is wrong in the screenshot above, but that can be solved in user code.)

Can the mouse signals be fixed to be made aware of the transform?

PR #3978, opened 2024-10-15
fix: respect element offset when calculating point

letelete:
## Motivation

- https://github.com/vega/vega/issues/3475
- https://github.com/vega/vega/issues/3600

## Changes

Added a function to map element points to the correct element offset, addressing interaction issues caused by scale transformations.

| Before | After |
| ---	| ---	|
| <video src="https://github.com/user-attachments/assets/a25616aa-9a40-47e8-8e2c-1fc854f03575">  |  <video src="https://github.com/user-attachments/assets/439cdc39-b200-4142-94d1-7604d8559fa1"> |

## Testing

This works locally when testing on:
- Firefox 132.0b7 (64-bit)
- Safari 17.6 (19618.3.11.11.5)
- Chrome Chrome 129.0.6668.91 (Official Build) (arm64)

Uses test spec:
- [layout-splom](https://github.com/vega/vega/blob/fb2e60274071033b4c427410ef43375b6f314cf2/packages/vega/test/scenegraphs/layout-splom.json)

## Notes

This fix addresses mouse signal distortion from the scaled root Vega element. It’s designed as a minimal-impact solution, allowing rescaling while preserving mouse interactions with minimal changes to the codebase.

**Concerns:**
- Potential for unknown edge cases.
- This issue might be better addressed by leveraging existing mechanisms (e.g., during renderer initialization).

### DIFF
--- a/packages/vega-scenegraph/src/util/point.js
+++ b/packages/vega-scenegraph/src/util/point.js
@@ -1,7 +1,78 @@
 export default function(event, el) {
   const rect = el.getBoundingClientRect();
-  return [
-    event.clientX - rect.left - (el.clientLeft || 0),
-    event.clientY - rect.top - (el.clientTop || 0)
-  ];
+  return mapPointToElementOffset(
+    [
+      event.clientX - rect.left - (el.clientLeft || 0),
+      event.clientY - rect.top - (el.clientTop || 0)
+    ],
+    el
+  );
+}
+
+function mapPointToElementOffset([x, y], el) {
+  const rect = el.getBoundingClientRect();
+  const [offsetWidth, offsetHeight] = getElementOffset(el);
+
+  const scaleX = offsetWidth ? rect.width / offsetWidth : 1;
+  const scaleY = offsetHeight ? rect.height / offsetHeight : 1;
+
+  return [x / scaleX, y / scaleY];
+}
+
+/**
+ * Retrieves the `offsetWidth` and `offsetHeight` of the given element.
+ * If not available, creates a container within the element to determine it.
+ *
+ * @param {Element} el - The target element.
+ * @returns {[offsetWidth: number, offsetHeight: number]} The element's offset.
+ */
+function getElementOffset(el) {
+  const offsetWidth = el.offsetWidth;
+  const offsetHeight = el.offsetHeight;
+  if (offsetWidth !== undefined && offsetHeight !== undefined) {
+    return [offsetWidth, offsetHeight];
+  }
+
+  let offsetTarget = el.querySelector(
+    '[data-offset-target-container]>[data-offset-target]'
+  );
+
+  if (!offsetTarget) {
+    el.style.position = 'relative';
+
+    const namespace = 'http://www.w3.org/2000/svg';
+
+    const offsetTargetContainer = document.createElementNS(
+      namespace,
+      'foreignObject'
+    );
+    offsetTargetContainer.setAttribute('x', '0');
+    offsetTargetContainer.setAttribute('width', '100%');
+    offsetTargetContainer.setAttribute('height', '100%');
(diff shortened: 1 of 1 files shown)


packages/vega-scenegraph/src/util/point.js
```diff
@@ -1,7 +1,78 @@
...
+      position: 'absolute',
+      left: '0',
+      top: '0',
+      width: '100%',
+      height: '100%'
+    });
+
+    offsetTargetContainer.appendChild(offsetTarget);
+    el.appendChild(offsetTargetContainer);
+  }
```
letelete, line 75: This feels like a hack.

hydrosquall:
Thanks for this contribution and the repro videos. I found the link from your [SO post](https://stackoverflow.com/questions/78652538/fit-vega-lite-react-vega-visualization-to-container-size). 

Do you have thoughts about how we might test that this change is working? I do wonder about whether the offset is something we can or should compute once in Vega-View (as you have suggested) at graph init time, rather making querying the dom potentially part of every scenegraph render.

domoritz:
@letelete This looks like a great fix that I would like to get in. I am also concerned about performance implications. Could you respond to @hydrosquall's comments?

jefflang80:
Has there been any movement on this fix?  I'm keen to see this bug resolved, too. Thanks!

letelete:
> @letelete This looks like a great fix that I would like to get in. I am also concerned about performance implications. Could you respond to @hydrosquall's comments?

Technically, the performance implications shouldn’t be a concern, since the function that calculates the offset creates the overlay lazily and does so only once per element instance. If the implementation looks promising, I could try profiling the before vs. after states to verify the actual impact.

Regarding the alternative approach—initializing during the graph init phase—I was hoping to get your guidance, as I’m not very familiar with the codebase. Could you point me to the files where such initialization could be implemented?

letelete:
cc: @domoritz @hydrosquall

hydrosquall:
Thanks for continuing to look into this @letelete ! It's a good callout that the impact is mitigated a bit by the fact a new DOM element is only created once per view, but we need to be careful about this function if it's called very frequently since we're also adding the overhead of extra `el.querySelector` calls. 

In terms of perf, aside from initializing the new DOM element, this solution will call getBoundingClientRect 3 times per call (and potentially trigger a reflow each of those times), I think we could simplify that part of the cost by passing `rect` through instead of `el` , since the `el` should not be changing in between those 3 calls.

Regarding init code, you may want to look around the constructor for the View class: 

https://github.com/vega/vega/blob/9177097be8bd59cda56530682e4df3aac01c575d/packages/vega-view/src/View.js#L58

domoritz:
Thanks for the fix. It's great to have interactions working even when the browser is zoomed. Great to hear that the impact on performance should still be minimal. 

I'll page @jheer and @arvind who might have advice for how to address the issue more elegantly (if possible).

letelete:
> Thanks for the fix. It's great to have interactions working even when the browser is zoomed. Great to hear that the impact on performance should still be minimal.
> 
> I'll page @jheer and @arvind who might have advice for how to address the issue more elegantly (if possible).

Thanks! I won't be able to work on it much longer (I'd have to file a permission request for OSS contribution from Google). Feel free to alter this PR to suit your needs.

domoritz:
Do you mean you can't work on it anymore or that you have a limited window where you can work on it? 

Btw, I think vega is in google3. Would be great to have general permission for google to work on vega since it's used by quite a few products as far as I know.

hydrosquall:
Hey @Ablu , thanks for your interest. If you have availability to build on this solution but to  adapt the solution based on this comment on a new branch, we would be glad to review + try to bring this in. 

https://github.com/vega/vega/pull/3978#issuecomment-2865050574

Issue #3987, opened 2024-11-14
inline clip-path instead of id ref


pluswave:
## Enhancement Description

I have a hexo blog with hexo-filter-kroki filter, which in turn, can render MULTIPLE vega-lite diagrams in ONE markdown file.  the problem is that, when I set 
```json
{
  "mark": { "clip": true }
}
```
the result svg will have two parts to do this, part one is something like

```xml

<g  ... clip-path="url(#clip1)"  ... >

```

part two is something like

```xml
<defs>
<clipPath id="clip1">
...
</clipPath>
</defs>
```

however, for multiple diagrams inlined in one html document, the id of clipPath conflicts and the renderer in browser is not correct except the first graph. To resolve this issue,  I suggest make the clip-path attribute inline for part one , and completely discards part two.

<details>
  <summary><h2>Checklist</h2></summary>

- [x] I checked for duplicate issues.
</details>

domoritz:
Moved. It definitely feels like something we should fix to allow multiple charts on a page. Would you feel comfortable making a pull request? An easy work around may be to generate IDs that don't collide.

hydrosquall:
> An easy work around may be to generate IDs that don't collide.

What if Vega graphs were given an optional"id" or "prefix" the view is generated that is used with the clip prefix? This way the responsibility of generating a unique identifier is pushed to the caller.

I thought about using a timestamp for the ID, but that would make the compile function return a different result each call since it's no longer a pure function.

For this particular case, I wonder if the decision on whether to inline vs use refs the Vega view or Vega-lite compile step could become a build option, since this seems like a runtime config rather than a setting that would be stored directly in a VL spec.

domoritz:
I'm curious why this issue happens. Vega does increase the clip id whenever you use a clip id. Could it be that you have multiple instances of Vega?

pluswave:
> I'm curious why this issue happens. Vega does increase the clip id whenever you use a clip id. Could it be that you have multiple instances of Vega?

Yes, as follow:

```dot
    hexo-plugin-kroki -rpc_call-> kroki_server
```

Every time [kroki](https://kroki.io) server response, it use a new instance of Vega.

domoritz:
Oh I see. So you render the chart and then just embed the SVG.

jurriaan:
We also have this issue with gradients (as they are defined like this `<linearGradient id="gradient_0"`). We now manually transform the ids to make sure there is no overlap.

PR #3988, opened 2024-11-25
svg clip path: now inlined instead of id ref (#3987)

pluswave:


domoritz:
Is there ever a case where we would reuse the clip masks and therefore using a reference would be more efficient?

pluswave:
> Is there ever a case where we would reuse the clip masks and therefore using a reference would be more efficient?

I don't know. but I suggest to change again if it is true. maybe add another option (inline) :

First  to make it more space efficient, there should be three or more same clip-path refs;

Second, in some cases, id ref is not allowed at all.  ( I know one case is [wechat pub media](https://mp.weixin.qq.com), as an author, any id attribute is not allowed in html format when upload content)

packages/vega-scenegraph/src/util/svg/clip.js
```diff
@@ -21,6 +14,10 @@ export default function(renderer, item, size) {
     c.width = size.width || 0;
     c.height = size.height || 0;
   }
-
-  return 'url(#' + id + ')';
+  if (c.path){
+    return `path(${c.path}) view-box`;
+  }
+  else{
```
domoritz, line 20:
```suggestion
  } else {
```

domoritz (approved):
Thank you. Just one style comment. Also, please merge main since we made some changes.

domoritz (changes requested):
Thank you. We will also need to replace the gradient since that also uses an id.

PR #3991, opened 2024-12-09
fix: axis "description" not overriding default aria-label

### LINKED ISSUE #9432: Axis "description" is not overriding default aria-label
## Bug Description

The `description` property with alt.Axis is not overriding the default aria-label. Below is the chart spec similar to what I'm trying to achieve. To confirm that the description is not overriding the default aria-label I'm inspecting the chart's code on the browser. For this example, the aria-label for the x-axis is:

image

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v5.20.1.json",
  "config": {
    "view": {
      "continuousHeight": 300,
      "continuousWidth": 300
    }
  },
  "data": {
    "name": "data-a"
  },
  "datasets": {
    "data-a": [
      {
        "a": "A",
        "b": 28
      },
      {
        "a": "B",
        "b": 55
      },
      {
        "a": "C",
        "b": 43
      },
      {
        "a": "D",
        "b": 91
      },
      {
        "a": "E",
        "b": 81
      },
      {
        "a": "F",
        "b": 53
      },
      {
        "a": "G",
        "b": 19
      },
      {
        "a": "H",
        "b": 87
      },
      {
        "a": "I",
        "b": 52
      }
    ]
  },
  "encoding": {
    "x": {
      "axis": {
        "description": "test",
        "labelAngle": 0
      },
      "field": "a",
      "type": "nominal"
    },
    "y": {
      "field": "b",
      "type": "quantitative"
    }
  },
  "mark": {
    "type": "bar"
  }
}
```

MarcBaeuerle:
## PR description
closes vega/vega-lite#9432

Issue rose because `mark.items[0]` didn't contain the `description` that was provided, so it defaulted to the aria attribute generators. The description was present in `mark.description` instead.

<details>
  <summary><h2>Checklist</h2></summary>

- [x] This PR is atomic (i.e., it fixes one issue at a time).
- [x] The title is a concise [semantic commit message](https://www.conventionalcommits.org/) (e.g. "fix: correctly handle undefined properties").
- [x] `yarn test` runs successfully

### DIFF
--- a/packages/vega-scenegraph/src/util/aria.js
+++ b/packages/vega-scenegraph/src/util/aria.js
@@ -100,12 +100,11 @@ function ariaMark(mark) {
 
 function ariaGuide(mark, opt) {
   try {
-    const item = mark.items[0],
-          caption = opt.caption || (() => '');
+    const caption = opt.caption || (() => '');
     return bundle(
       opt.role || GRAPHICS_SYMBOL,
       opt.desc,
-      item.description || caption(item)
+      mark.description || caption(mark)
     );
   } catch (err) {
     return null;


domoritz (changes requested):
Thank you for the pull request. 

Unfortunately, this breaks the default descriptions. 

Before
Screenshot 2025-03-16 at 20 46 03

After
Screenshot 2025-03-16 at 20 44 59

Could you look into supporting the existing behavior while also fixing the bug with custom descriptions? 

It would be great to add a test case to make sure we support these two use cases.

PR #3998, opened 2025-01-13
Dwootton/add-bin-tresholds

dwootton:
Used to make thresholds an explicit parameter for Vega's bin transform. 

This is particularly useful when the bin thresholds one wants to set need to be customized via signals.

PR #3999, opened 2025-01-14
test: replace `flights-5k.json` external reference

dwootton:
This PR resolves a previous issue with one of our tests referencing an external dataset in vega_datasets. I've updated the spec to reference the same dataset but as a local copy such that the test doesn't fail if the external dataset gets updated.


dangotbanned (approved):
Thanks for fixing this @dwootton

I'll leave this open until the end of the day, in case anyone from (https://github.com/vega/vega/pull/3998#issuecomment-2589551144) has any feedback

PR #4000, opened 2025-01-16
feat(vega-typings): add Typescript Types for `vega-loader`

hydrosquall:
## Motivation

- I was learning about `vega-loader`, and decided to move the information from the [README](https://github.com/vega/vega/blob/main/packages/vega-loader/README.md) to the typings file. 
- This removes 3 `any` types from `Loader` too!

## Changes

- This should have no runtime impact, it only updates Typescript definitions in `vega-typings`
- Note: `vega-loader`'s README does not mention that `sanitize` takes in `context` and `response` as additional options - this is something I noticed from `console.log`'ing the output of the options.  For now I only document that in the Typescript types. 
   - I could update the README for vega-loader as well if we want consumers to know about this, or leave it out if this is supposed to be an internal implementation detail.

### DIFF
--- a/packages/vega-typings/types/runtime/index.d.ts
+++ b/packages/vega-typings/types/runtime/index.d.ts
@@ -32,7 +32,7 @@ type BaseLoaderOptions = {
   /** Allows caller to explicitly set loading mode (local or network request). File mode only applies to server-side rendering. */
   mode: 'file' | 'http';
   /** Default protocol for protocol-relative URIs, defaults to HTTP */
-  defaultProtocol: string;
+  defaultProtocol: 'file' | 'http' | string;
   /** browser `target` attribute for hyperlinks. Only used when sanitizing URI values for use as hyperlink */
   target: string;
   /** browser `rel` attribute for hyperlinks. Only used when sanitizing URI values for use as hyperlink */
@@ -49,7 +49,7 @@ type BaseLoaderOptions = {
 /**
  * Informs loader which context the URI will be used in.
  */
-type OptionsWithContext =
+type LoaderOptionsWithContext =
   | (Partial<BaseLoaderOptions> & {
       /**
        * Describes context in which the URI will be used.
@@ -74,8 +74,8 @@ type OptionsWithContext =
  * https://github.com/vega/vega/tree/main/packages/vega-loader
  */
 export interface Loader {
-  load: (uri: string, options?: OptionsWithContext) => Promise<string>;
-  sanitize: (uri: string, options: OptionsWithContext) => Promise<{ href: string }>;
+  load: (uri: string, options?: LoaderOptionsWithContext) => Promise<string>;
+  sanitize: (uri: string, options: LoaderOptionsWithContext) => Promise<{ href: string }>;
   http: (uri: string, options: Partial<RequestInit>) => Promise<string>;
   file: (filename: string) => Promise<string>;
 }
--- a/packages/vega-typings/types/runtime/index.d.ts
+++ b/packages/vega-typings/types/runtime/index.d.ts
@@ -25,10 +25,58 @@ export function timeFormatLocale(definition: object): void;
 // Parser
 export function parse(spec: Spec, config?: Config, opt?: { ast?: boolean }): Runtime;
 
+/** Types documented in vega-loader's README.md */
+type BaseLoaderOptions = {
+  /** Base URL prefix prepended to provided URI */
+  baseUrl: string;
+  /** Allows caller to explicitly set loading mode (local or network request). File mode only applies to server-side rendering. */
+  mode: 'file' | 'http';
+  /** Default protocol for protocol-relative URIs, defaults to HTTP */
+  defaultProtocol: string;
+  /** browser `target` attribute for hyperlinks. Only used when sanitizing URI values for use as hyperlink */
+  target: string;
+  /** browser `rel` attribute for hyperlinks. Only used when sanitizing URI values for use as hyperlink */
+  rel: string;
+  /** http request parameters passed to underlying [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) */
+  http: RequestInit;
+  /**
+   * Specify a [crossOrigin](https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_enabled_image) attribute for images. Only used when sanitizing URI values with option context: "image"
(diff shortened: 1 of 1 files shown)


hydrosquall:
I'm unclear about how to find out what the errors are in the [build](https://github.com/vega/vega/actions/runs/12818145678/job/35742955281?pr=4000) (which seem to be formatting related)

image

I've run both `yarn lint` and `yarn format` on my source file `packages/vega-typings/types/runtime/index.d.ts` , neither of which make changes to the current head of this branch.

image

Will open a thread in vega-dev about this.

packages/vega-typings/types/runtime/index.d.ts
```diff
@@ -25,10 +25,60 @@ export function timeFormatLocale(definition: object): void;
...
+
+/**
+* Sanitize accepts all options from Loader constructor
+*/
+type SanitizeOptions =
+  | (Partial<LoaderOptions> & {
+        /**
+         * Describes context in which the URI will be used.
+         */
+        context: 'href' | 'image';
```
hydrosquall: From code search for how `context: '` gets called in the monorepo, I found just 3 sites for href and image.

https://github.com/vega/vega/blob/0b652cd06d9d02172e0740f0ae4544f863f16aa6/packages/vega-scenegraph/src/Handler.js#L138

https://github.com/vega/vega/blob/0b652cd06d9d02172e0740f0ae4544f863f16aa6/packages/vega-scenegraph/src/ResourceLoader.js#L19

https://github.com/vega/vega/blob/0b652cd06d9d02172e0740f0ae4544f863f16aa6/packages/vega-scenegraph/src/ResourceLoader.js#L35

packages/vega-typings/types/runtime/index.d.ts
```diff
@@ -25,10 +25,60 @@ export function timeFormatLocale(definition: object): void;
...
+    })
+  | {
+        /**
+         * context:dataflow is used when loader is used for getting data into vega-dataflow
+         * https://vega.github.io/vega/docs/data/
+         */
+        context: 'dataflow';
```
hydrosquall: Added by vega-dataflow: 

https://github.com/vega/vega/blob/0b652cd06d9d02172e0740f0ae4544f863f16aa6/packages/vega-dataflow/src/dataflow/load.js#L42-L43

packages/vega-typings/types/runtime/index.d.ts
```diff
@@ -25,10 +25,58 @@ export function timeFormatLocale(definition: object): void;
...
+ * Loader object is used for loading data, images, and links (hrefs) from a Uniform Resource Identifier (URI)
+ * https://github.com/vega/vega/tree/main/packages/vega-loader
+ */
 export interface Loader {
-  load: (uri: string, options?: any) => Promise<string>;
-  sanitize: (uri: string, options: any) => Promise<{ href: string }>;
-  http: (uri: string, options: any) => Promise<string>;
+  load: (uri: string, options?: OptionsWithContext) => Promise<string>;
+  sanitize: (uri: string, options: OptionsWithContext) => Promise<{ href: string }>;
+  http: (uri: string, options: Partial<RequestInit>) => Promise<string>;
```
hydrosquall, line 79: Setting `RequestInit` based on the signature of what `fetch` would accept:

https://github.com/vega/vega/blob/0b652cd06d9d02172e0740f0ae4544f863f16aa6/packages/vega-loader/src/loader.js#L173-L175

packages/vega-typings/types/runtime/index.d.ts
```diff
@@ -25,10 +25,58 @@ export function timeFormatLocale(definition: object): void;
...
+   * Specify a [crossOrigin](https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_enabled_image) attribute for images. Only used when sanitizing URI values with option context: "image"
+   * If this property is defined and maps to a value of `null` or `undefined`, then a `no-cors` fetch will be performed for the `Image`. This property can be used to override Vega's default behavior of using `crossOrigin="anonymous"`, which allows images loaded from a different host to be included in exported visualization images (and thereby avoid "tainted canvas errors"), so long as the server provides permission via proper CORS headers.
+   */
+  crossOrigin: 'anonymous' | 'use-credentials' | '' | undefined;
+};
+
+/**
+ * Informs loader which context the URI will be used in.
+ */
+type OptionsWithContext =
```
danmarshall, line 52: Perhaps rename this to LoaderOptionsWithContext ?

packages/vega-typings/types/runtime/index.d.ts
```diff
@@ -25,10 +25,58 @@ export function timeFormatLocale(definition: object): void;
...
+  | (Partial<BaseLoaderOptions> & {
```
domoritz, line 53: Why are the options optional now but context is not?

domoritz (approved):

packages/vega-typings/types/runtime/index.d.ts
```diff
@@ -25,10 +25,58 @@ export function timeFormatLocale(definition: object): void;
...
 export function parse(spec: Spec, config?: Config, opt?: { ast?: boolean }): Runtime;
 
+/** Types documented in vega-loader's README.md */
+type BaseLoaderOptions = {
+  /** Base URL prefix prepended to provided URI */
+  baseUrl: string;
+  /** Allows caller to explicitly set loading mode (local or network request). File mode only applies to server-side rendering. */
+  mode: 'file' | 'http';
+  /** Default protocol for protocol-relative URIs, defaults to HTTP */
+  defaultProtocol: string;
```
domoritz, line 35: Can we enumerate this?

hydrosquall: It looks like this can be file, http, a string that starts with `//`, or a bunch of other things: https://en.wikipedia.org/wiki/List_of_URI_schemes 

I'll include HTTP and file for a start, but leave a string at the end for those other cases.

packages/vega-typings/types/runtime/index.d.ts
hydrosquall: As far as I could see, `context` was provided every time that `Vega` is internally calling [sanitize](https://github.com/vega/vega/pull/4000#discussion_r1919527743) or [load](https://github.com/vega/vega/pull/4000#discussion_r1919527775), so it seemed worthwhile to document that this info is available to switch on

External users of `vega-loader` aren't required to know about this since it context is not mentioned in the previous [README](https://github.com/vega/vega/tree/main/packages/vega-loader) except for images. I wouldn't have known about it if I didn't happen to be console what sorts of data was getting  passed to `sanitize` and `load` when verifying the types.

I suppose we could mark these types as optional since the code seems to run OK if you use `vega-loader` without also using it inside of `vega`, but I figured erring on the side of strictness would be good here. Happy to adjust the type if you'd recommend changing it.

domoritz: I think it's fine but I wanted us to be aware of this.

domoritz (approved):

Issue #3990, opened 2024-12-03
Move all packages to ESM


domoritz:
We are planning to move all Vega packages to ESM only. Other packages like D3 have done this a while ago and it reduces future maintenance burden for us.

This will require a major release for all packages.

- [x] Vega https://github.com/vega/vega/pull/4018
- [x] Vega-Lite https://github.com/vega/vega-lite/pull/9527
- [x] Vega Embed https://github.com/vega/vega-embed/pull/1384
- [x] Vega Tooltip https://github.com/vega/vega-tooltip/pull/1000
- [x] Vega Themes https://github.com/vega/vega-themes/pull/651
- [x] Vega Datasets https://github.com/vega/vega-datasets/pull/673
- [ ] ~~ts-json-schema-generator~~

See https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c and https://github.com/d3/d3/issues/3469.

hydrosquall:
> it reduces future maintenance burden for us.

This sounds good to me, It could be beneficial for future maintainers reading about this move for us to to see which of these burdens are heaviest right now. The Sindresorhus link covers how, and the D3 post talks about node.js incompatibility, but perhaps we can have our own short recap of the why (especially given that given its interactive roots, I'd predict most vega users are in a browser rather than server-side rendering). 

Do you have a recommended article o list of highlights for what we gain or lose by moving to ESM? 

My understanding of the main consequences might include that we

1. Drop support for older browsers (pre 2015) using CJS  as we assume usage in this group is low 
2. Don't have to ship compiled JS in the repo anymore (i.e. https://github.com/vega/vega/blob/main/docs/vega.min.js ), making releasing changes easier (we can just edit the source files)
3. Align w/ JS ecosystem as a whole is moving towards ESM only (at 26% as of Nov 2024)
4. Get general ESM benefits: modern tools rely on ESM for developer benefits like tree-shaking, hot-reloading, etc per Anthony's [post](https://antfu.me/posts/move-on-to-esm-only)
5. Fewer complications when it comes to making proper types for exported objects when providing Typescript types 
6. Fewer complications when it comes to resolving dependencies w/ conflicting versions ( e.g.  #3932 )
7. Smaller package sizes (by not having to ship both versions)

domoritz:
[...]

PR #4008, opened 2025-02-02
docs: Add Security Advisory Policy for `vega`

hydrosquall:
## Motivation

- Make it easy for researchers and engineers to have a safe path to reporting security vulnerabilities
- Keep Vega and its dependents (Vega-lite, altair, etc) secure

## Changes

- Add a `SECURITY.md` ([Security Policy](https://docs.github.com/en/code-security/getting-started/adding-a-security-policy-to-your-repository)), following a slack discussion in the Maintainers channel. This is based on
  - Electron https://github.com/electron/electron/blob/main/SECURITY.md
  - Microsoft: https://github.com/microsoft/.github/blob/main/SECURITY.md
- While linking this file from the homepage, I noticed we could make the `Vega` README a bit more like the `Vega-Lite` one (shorter and with more direct docs links). I can revert these changes if preferred, I figured it didn't hurt to propose them.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,4 +22,15 @@ The Vega project consists of a number of packages, organized into a single [mono
 
 For an overview of all packages, see the [vega `/packages` folder](/packages).
 
-To setup a development environment follow the [Build Instructions in README.md](https://github.com/vega/vega/#Build-Instructions). We use [Yarn workspaces](https://yarnpkg.com/lang/en/docs/workspaces/) and [lerna](https://github.com/lerna/lerna) to manage the monorepo packages.
+To setup a development environment follow the [Build Instructions in README.md](#build-instructions). We use [Yarn workspaces](https://yarnpkg.com/lang/en/docs/workspaces/) and [lerna](https://github.com/lerna/lerna) to manage the monorepo packages.
+
+## Build Instructions
+
+For a basic setup allowing you to build Vega and run examples:
+
+- Clone `https://github.com/vega/vega`.
+- Run `yarn` to install dependencies for all packages. If you don't have yarn installed, see https://yarnpkg.com/en/docs/install. We use [Yarn workspaces](https://yarnpkg.com/lang/en/docs/workspaces/) to manage multiple packages within this [monorepo](https://en.wikipedia.org/wiki/Monorepo).
+- Once installation is complete, run `yarn test` to run test cases, or run `yarn build` to build output files for all packages.
+- After running either `yarn test` or `yarn build`, run `yarn serve` to launch a local web server &mdash; your default browser will open and you can browse to the `"test"` folder to view test specifications.
+
+This repository includes the Vega website and documentation in the `docs` folder. To launch the website locally, first run `bundle install` in the `docs` folder to install the necessary Jekyll libraries. Afterwards, use `yarn docs` to build the documentation and launch a local webserver. After launching, you can open [`http://127.0.0.1:4000/vega/`](http://127.0.0.1:4000/vega/) to see the website.
--- a/README.md
+++ b/README.md
@@ -4,23 +4,14 @@
 <img src="https://vega.github.io/vega/assets/banner.png" alt="Vega Examples" width="900"></img>
 </a>
 
-**Vega** is a *visualization grammar*, a declarative format for creating, saving, and sharing interactive visualization designs. With Vega you can describe data visualizations in a JSON format, and generate interactive views using either HTML5 Canvas or SVG.
+**Vega** is a *visualization grammar*, a declarative format for creating, saving, and sharing interactive visualization designs. With Vega you can describe data visualizations in a JSON format, and generate interactive views using HTML5 Canvas or SVG.
 
-For documentation, tutorials, and examples, see the [Vega website](https://vega.github.io/vega). For a description of changes between Vega 2 and later versions, please refer to the [Vega Porting Guide](https://vega.github.io/vega/docs/porting-guide/).
+For [documentation](https://vega.github.io/vega/docs/), [tutorials](https://vega.github.io/vega/tutorials/), and [examples](https://vega.github.io/vega/examples/), see the [Vega website](https://vega.github.io/vega). For a description of changes between Vega 2 and later versions, please refer to the [Vega Porting Guide](https://vega.github.io/vega/docs/porting-guide/).
 
-## Build Instructions
-
-For a basic setup allowing you to build Vega and run examples:
-
-- Clone `https://github.com/vega/vega`.
-- Run `yarn` to install dependencies for all packages. If you don't have yarn installed, see https://yarnpkg.com/en/docs/install. We use [Yarn workspaces](https://yarnpkg.com/lang/en/docs/workspaces/) to manage multiple packages within this [monorepo](https://en.wikipedia.org/wiki/Monorepo).
-- Once installation is complete, run `yarn test` to run test cases, or run `yarn build` to build output files for all packages.
-- After running either `yarn test` or `yarn build`, run `yarn serve` to launch a local web server &mdash; your default browser will open and you can browse to the `"test"` folder to view test specifications.
-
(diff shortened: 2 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -22,4 +22,15 @@ The Vega project consists of a number of packages, organized into a single [mono
 
 For an overview of all packages, see the [vega `/packages` folder](/packages).
 
-To setup a development environment follow the [Build Instructions in README.md](https://github.com/vega/vega/#Build-Instructions). We use [Yarn workspaces](https://yarnpkg.com/lang/en/docs/workspaces/) and [lerna](https://github.com/lerna/lerna) to manage the monorepo packages.
+To setup a development environment follow the [Build Instructions in README.md](#build-instructions). We use [Yarn workspaces](https://yarnpkg.com/lang/en/docs/workspaces/) and [lerna](https://github.com/lerna/lerna) to manage the monorepo packages.
+
+## Build Instructions
```
hydrosquall, line 27: This section was lifted without modifications from `README.md`

README.md
```diff
@@ -4,23 +4,14 @@
...
-
-- Clone `https://github.com/vega/vega`.
-- Run `yarn` to install dependencies for all packages. If you don't have yarn installed, see https://yarnpkg.com/en/docs/install. We use [Yarn workspaces](https://yarnpkg.com/lang/en/docs/workspaces/) to manage multiple packages within this [monorepo](https://en.wikipedia.org/wiki/Monorepo).
-- Once installation is complete, run `yarn test` to run test cases, or run `yarn build` to build output files for all packages.
-- After running either `yarn test` or `yarn build`, run `yarn serve` to launch a local web server &mdash; your default browser will open and you can browse to the `"test"` folder to view test specifications.
-
-This repository includes the Vega website and documentation in the `docs` folder. To launch the website locally, first run `bundle install` in the `docs` folder to install the necessary Jekyll libraries. Afterwards, use `yarn docs` to build the documentation and launch a local webserver. After launching, you can open [`http://127.0.0.1:4000/vega/`](http://127.0.0.1:4000/vega/) to see the website.
+Try using Vega in the online [Vega Editor](https://vega.github.io/editor/#/examples/vega/bar-chart).
```
hydrosquall, line 11: This is based on a similar line in the `Vega-Lite` readme.md to encourage people to try the library right away without having to install anything.

README.md
```diff
@@ -4,23 +4,14 @@
...
 
 ## Internet Explorer Support
-For backwards compatibility, Vega includes a [babel-ified](https://babeljs.io/) IE-compatible version of the code in the `packages/vega/build-es5` directory. Older browser would also require several polyfill libraries:
+For backwards compatibility, Vega includes a [babel-ified](https://babeljs.io/) IE-compatible version of the code in the `packages/vega/build-es5` directory. Older browsers also require these polyfill libraries:
```
hydrosquall, line 14: I considered moving this section on IE into Contributing / Development as well, but decided to leave it in place. I'm not sure how relevant this section still is to keep on the homepage given the default MSFT browser has been Edge for 10 years.

SECURITY.md
```diff
@@ -0,0 +1,21 @@
+# Security
+
```
hydrosquall, line 2: I blended the Github and Electron README's together. 

We could alternately put this directly in our `.github` community folder to have a policy that applies to all repos in the `vega` org, but I figured it would be safe and meaningful to start with `vega/vega` (since this is the core browser runtime that every other package is importing).

SECURITY.md
```diff
@@ -0,0 +1,21 @@
...
+## Reporting Security Issues
+
+**Please do not report security vulnerabilities through public GitHub issues.**
+
+To report a security issue privately, please use the GitHub Security Advisory ["Report a Vulnerability"](https://github.com/vega/vega/security/advisories/new) tab.
+
+A Vega maintainer will send a response indicating next steps in handling your report. After the initial reply, the team will keep you informed of the progress towards a fix and announcement, and may ask for additional information or guidance.
+
+You can also report a vulnerability through the [npm contact form](https://www.npmjs.com/support) by selecting "I'm reporting a security vulnerability".
```
domoritz, line 13: Where does that go?

SECURITY.md
```diff
@@ -0,0 +1,21 @@
+# Security
```
domoritz, line 1: Is there a way to have an org wide document rather than per repo?

SECURITY.md
hydrosquall: I believe the NPM security team would contact the account owner for https://www.npmjs.com/package/vega . I'm including this mainly so that someone without a Github account has a way to reach us. 

If we would find this redundant with the Github security advisory feature, I can remove it.

SECURITY.md
hydrosquall: We have the option of putting this in `.github` like Microsoft does [here](https://github.com/microsoft/.github/blob/main/SECURITY.md). It will fill in the "security" tab if a repo doesn't already have one, just confirmed from checking an old Microsoft [example](https://github.com/microsoft/sudokumaster-wp?tab=security-ov-file)

image

I can relocate this draft to `vega/.github` if we prefer!

README.md
```diff
@@ -4,23 +4,14 @@
...
-- Once installation is complete, run `yarn test` to run test cases, or run `yarn build` to build output files for all packages.
-- After running either `yarn test` or `yarn build`, run `yarn serve` to launch a local web server &mdash; your default browser will open and you can browse to the `"test"` folder to view test specifications.
-
-This repository includes the Vega website and documentation in the `docs` folder. To launch the website locally, first run `bundle install` in the `docs` folder to install the necessary Jekyll libraries. Afterwards, use `yarn docs` to build the documentation and launch a local webserver. After launching, you can open [`http://127.0.0.1:4000/vega/`](http://127.0.0.1:4000/vega/) to see the website.
+Try using Vega in the online [Vega Editor](https://vega.github.io/editor/#/examples/vega/bar-chart).
 
 ## Internet Explorer Support
-For backwards compatibility, Vega includes a [babel-ified](https://babeljs.io/) IE-compatible version of the code in the `packages/vega/build-es5` directory. Older browser would also require several polyfill libraries:
```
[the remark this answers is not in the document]
hydrosquall: Makes sense to handle that in a separate issue 👍

SECURITY.md
hydrosquall: Added a modified proposal to the shared repo  https://github.com/vega/.github/pull/20 . It will fill in for any repos that are missing a local security advisory.

README.md
domoritz: Let's delete it in the next major version.

SECURITY.md
domoritz: Yeah, let's remove it and have one clear path.

SECURITY.md
```diff
@@ -0,0 +1,21 @@
...
+
+To report a security issue privately, please use the GitHub Security Advisory ["Report a Vulnerability"](https://github.com/vega/vega/security/advisories/new) tab.
+
+A Vega maintainer will send a response indicating next steps in handling your report. After the initial reply, the team will keep you informed of the progress towards a fix and announcement, and may ask for additional information or guidance.
+
+You can also report a vulnerability through the [npm contact form](https://www.npmjs.com/support) by selecting "I'm reporting a security vulnerability".
+
+## Languages
+
+Communications should be in English.
```
domoritz, line 17: I think we can delete that since we already have that for the org. If we ever get a request not in English, we can respond.

SECURITY.md
domoritz: Yeah, makes sense to have one for Vega specifically. We could also link to the org one for more info so we only have to maintain one place.

README.md
hydrosquall: Noting this as part of the ESM support (since switching over would end that compatibility): https://github.com/vega/vega/issues/3990

hydrosquall:
The main reason to have a dedicated one for Vega/Vega vs reusing the org
one is to have an updated link for the “report a vulnerability” tab, and so
people have a copy of this file when they git clone.

You are right that most of the document is the same as in the org link. We
could shorten this doc to just the pointer to the org link and the
intro/report a vulnerability link if we prefer. There’s already a link to
the org one at the bottom of the security.MD in this PR.

domoritz:
>You are right that most of the document is the same as in the org link. We
could shorten this doc to just the pointer to the org link and the
intro/report a vulnerability link if we prefer.

Yes, that was what I was thinking.

domoritz (approved):

Issue #3875, opened 2024-02-16
Add a function for Base64 encoding to the expressions


samuel-pina:
In Vega-Lite, we don't have the _image_  property of the image mark as we do in vega, we only have the url one. In some uses like in Deneb in Power BI, there are certain limitations for using the url property as it could be connecting with external content. If we are using base64 encoded images declared inline for the _url_ property, we can overcome this limitation but then we cannot alter the image (for instance, a svg) dynamically as we don't have a way in vega-lite to encode them later to base64. [I explain a little bit better the scenario in this post](https://github.com/deneb-viz/deneb/issues/422#issuecomment-1945778121).
At the end, it is possible to change the image dynamically in vega, but it would be great if we could modify the svg in vega-lite and then encode it in base64 for using it in the url property. Would it be possible to add such function to the expression language?

hydrosquall:
Hi @samuel-pina ,

Could you share more about what the function you're looking to have added would do? Are you looking for a binding to the `btoa` function? https://developer.mozilla.org/en-US/docs/Web/API/Window/btoa 

I'm not fully following where you'd get the data you want to encode to base64 from, but if what you need is addressed by `btoa` and its cousin `btoa`, it seems like something we could add.

samuel-pina:
Hello @hydrosquall ,

Thanks a lot for your answer. The basic idea was to change dynamically an svg image inside of Vega, circumventing the limitation for accessing external content using DENEB in Power BI.  With the current configuration, the property _url_ for an image mark could only be an url (so blocked by PowerBI in DENEB) or a base64 encoded image. If we used the later, then we weren't able to change the SVG image dynamically as we don't have the ability to decode the image for changing it later.
At the end I was able to find a workaround, as [explained here.](https://github.com/deneb-viz/deneb/issues/422#issuecomment-1945778121)

Thanks again,

Samuel

hydrosquall:
Hi @samuel-pina , thanks again for the idea and sharing the workaround. 

In light of the workaround, would you still be looking to have this feature request, or is it now unnecessary?

PBI-David:
@samuel-pina  I'm not sure if this helps but here is an example of data driven SVG paths. 

![Image](https://github.com/user-attachments/assets/95607622-356f-4166-b0f2-b45145b722c7)

```

{
  "$schema": "https://vega.github.io/schema/vega/v5.json",
  "description": "Dataviz by David Bacci: https://www.linkedin.com/in/davbacci/",
[...]

PR #4009, opened 2025-02-04
feat: Add base64 string encoder/decoder to `vega-expression` and `vega-interpreter`

hydrosquall:
## Motivation

- Allow users to modify SVG images that were Base64 encoded. This is useful in Vega-Lite environments where remote access to images was disabled, such as Deneb (PowerBI), so users are especially dependent on inline data definitions.
- See #3875 

## Testing

- Try test URL sandbox showing that images can be encoded/decoded [link](https://cameron-yick-feature-add-bas.vega-628.pages.dev/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEySYUqUMSSCGcCGgDaoDSACCikEzQAmABwBfBSbQgAQpeuoxYpy-QAwp5oACwAzA4Auk4gyABOANauTEhxlnCyUGzKNLJkaKAAngUgAGY0cILKrhlZynDVSpiFOHCusmwIuTogMQAeJeWV1eg09bKYdMVNLW3oHV2yPTFZgmxpmmUVVa4oM63tnd2CloLkGSOyDIKCDjGYcUiyEKXrCEagUDpQ12pzIGo2EwABRMTBsJDA1SYBgIShIACUCMsSAMo3Gk2avWcIC+gh+gj+yXBkOhsPhyKUqJqmWyDV60SAA)
  - Make sure it works with AND without vega-interpreter mode: https://github.com/vega/editor/issues/1460
- I did not unit test this change yet as the function is a core web API. However, if it would be useful to have at least 1 spec that uses both atob and btoa, I can add one. 

## Notes

- btoa / atob: check the status for node.js usage ( https://github.com/DefinitelyTyped/DefinitelyTyped/issues/65494 ) since the Node.js world prefers using `Buffer.from(` instead.
- Check why/if this is different from what we did for vega-functions in https://github.com/vega/vega/pull/3973
- To get the scenegraph for my test, I visited https://cameron-yick-feature-add-bas.vega-628.pages.dev/#/edited, used and used `VEGA_DEBUG.view` to get a sample scenegraph.

### DIFF
--- a/packages/vega-interpreter/src/functions.js
+++ b/packages/vega-interpreter/src/functions.js
@@ -75,6 +75,11 @@ export default {
   split:        function() { return apply('split', arguments, String); },
   replace:      function() { return apply('replace', arguments, String); },
   trim:         x => String(x).trim(),
+  // Base64 encode/decode
+  // Convert binary string to base64-encoded ascii
+  btoa:         x => btoa(x),
+  // Convert base64-encoded ascii to binary string
+  atob:         x => atob(x),
 
   // regexp functions
   regexp:       RegExp,
--- a/docs/docs/expressions.md
+++ b/docs/docs/expressions.md
@@ -523,7 +523,7 @@ Returns an array containing an arithmetic sequence of numbers. If _step_ is omit
 Returns a section of _array_ between the _start_ and _end_ indices. If the _end_ argument is negative, it is treated as an offset from the end of the array (_length(array) + end_).
 
 <a name="sort" href="#sort">#</a>
-<b>sort</b>(<i>array</i>)<br/>
+<b>sort</b>(<i>array</i>)<br/>{% include tag ver="5.31" %}<br/>
 Sorts the array in natural order using [ascending from Vega Utils](https://vega.github.io/vega/docs/api/util/#ascending).
 
 <a name="span" href="#span">#</a>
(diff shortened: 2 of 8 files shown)


packages/vega-interpreter/src/functions.js
```diff
@@ -75,6 +75,11 @@ export default {
   split:        function() { return apply('split', arguments, String); },
   replace:      function() { return apply('replace', arguments, String); },
   trim:         x => String(x).trim(),
+  // Base64 encode/decode
+  // Convert binary string to base64-encoded ascii
+  btoa:         x => btoa(x),
```
domoritz, line 80: Could this just be

```suggestion
  btoa:         btoa,
```

domoritz (approved):

packages/vega-interpreter/src/functions.js
hydrosquall: It could. A small difference is wrapping btoa in a function ensures that our public API ensures that only 1 argument is ever passed to `btoa`. 

This is similar to how writing `['1','2'].map(parseInt)` and `['1','2'].map((num) => parseInt(num))` are not equivalent

```js
['5','2'].map((num) => parseInt(num))
// returns [5, 2]
['5','2'].map(parseInt)
// returns [5, NaN]
```

The NaN appears as a result of calling `parseInt(2, 1)` since the index of the array is passed as the second arg of the mapping function, and changes the radix to something other than base 10. We want to avoid that.

domoritz (approved):
Please add a test

hydrosquall:
@domoritz I have added both a unit test for `vega-functions` and an interpreter-mode test for `vega-interpreter`. 

It felt a bit heavy to add 2 big JSONs just to test these two functions, but that appears to be the pattern for vega-interpreter (rather than testing just the function in isolation). Let me know if you have feedback about naming/content of the scenegraph test.

domoritz:
I had a similar reaction when I saw the tests but it's fine.

domoritz (approved):

PR #4024, opened 2025-03-06
docs:  fix docs for exponential regression

elliot-at-looker:
Calculation for exponential regression in the docs is wrong, should be y = a * e^(b * x) not y = a + e^(b * x)


domoritz (approved):
Thanks!

PR #4027, opened 2025-03-16
chore: v5.33.0

lsh:
Changes since v5.32.0

**docs**
* Add release number to docs for base64 encode/decode (atob, btoa) (via #4023). (Thanks @hydrosquall!)
* Fix docs for exponential regression (via #4024). (Thanks @elliot-at-looker!)


hydrosquall (approved):
Makes sense to me, thanks for the `vega-functions` and related dependent bumps too.

PR #4031, opened 2025-03-19
Merge next into main

domoritz:


domoritz:
I don't know why these commits are not in main but looks like it wasn't really ment to be.

PR #4037, opened 2025-03-25
fix: correct browser exports syntax (vega loader and canvas)

domoritz:
@hydrosquall maybe this fixes the issue with parcel picking up the wrong code


domoritz:
Replaced by https://github.com/vega/vega/pull/4039

PR #4039, opened 2025-03-25
fix: correct browser exports syntax

domoritz:
This removes UD bundles from everything but `vega`.

PR #4040, opened 2025-03-25
chore: v6.0.0

lsh:
changes since v5.33.0

**monorepo**
* Vega is now ESM only! (via #4018) (Thanks @domoritz!)
* (fix) Correct browser export syntax (via #4039) (Thanks @domoritz!)
* (fix) Fix source maps (via #4041) (Thanks @domoritz!)
* (fix) handle string datetime in interpreter (via #4042) (Thanks @domoritz!)

**vega-typings**
* Expose resetSVGDefIds (via #4012) (Thanks @domoritz!)

**docs**
* Update deps and datasets (via #4036) (Thanks @domoritz!)


docs/vega.js
```diff
@@ -41738,3 +41729,4 @@
   exports.zoomSymlog = zoomSymlog;
 
 }));
+//# sourceMappingURL=vega.js.map
```
domoritz, line 41732: looks like we should copy that as well. I'll fix it

domoritz, line 41732: looks like we should copy that as well. I'll fix it

domoritz (approved):

docs/vega.js
domoritz: Done in #4041

domoritz:
We don't need to block on the preview builds.

PR #4044, opened 2025-03-28
fix(preview): Use parcel `2.12.0` for building PR previews

hydrosquall:
## Motivation

- Same as https://github.com/vega/editor/pull/1475, but rolling back a minor version further back (`2.12.0`)
- Restore service to #3924 

## Testing

- Check if the cloudflare preview passes! (it does: https://cameron-yick-test-reduce-par.vega-628.pages.dev/ )

### DIFF
--- a/scripts/build-editor-preview.sh
+++ b/scripts/build-editor-preview.sh
@@ -22,6 +22,8 @@ git clone https://github.com/vega/editor.git
 
 cd editor
 yarn --frozen-lockfile --ignore-scripts
+# Parcel 2.13.0 and later break the preview build when run in CI.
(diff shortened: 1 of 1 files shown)


domoritz:
Nice. We should port these to the editor, though, right?

hydrosquall:
> Nice. We should port these to the editor, though, right?

We can do this in either place. Since the node build seems to be erroring for unrelated reasons, I'll open the PR in the editor (this way it will fix things for vega-lite too).

hydrosquall:
Replacing with https://github.com/vega/editor/pull/1476

PR #4047, opened 2025-03-29
feat: switch to npm, use native fetch

domoritz:


hydrosquall:
Nice, confirmed that swapping in native fetch makes sense after [node 18](https://nodejs.org/en/blog/announcements/v18-release-announce/) and lets us cut the bundle size.

We may want to note a minimum node version in the readme or development docs if that's the case.

domoritz:
Node 18 is the latest supported and it's also indicated in the package.json already. Might be enough.

PR #4048, opened 2025-03-29
fix: correct mode handling in loader

domoritz:
`mode` is already an argument to fetch (https://developer.mozilla.org/en-US/docs/Web/API/RequestInit#mode) so we need to change the name.

Technically it's a breaking change but we had an option collision so I think we should consider this a fix. Plus v6 hasn't been out for long. 

Follow up for #4047

### DIFF
--- a/packages/vega-loader/src/loader.js
+++ b/packages/vega-loader/src/loader.js
@@ -48,7 +48,7 @@ async function load(uri, options) {
 
   return opt.localFile
     ? this.file(url)
-    : this.http(url, options);
+    : this.http(url, options.http);
 }
 
 /**
@@ -161,7 +161,7 @@ async function fileReject() {
 /**
  * An http loader.
  * @param {string} url - The url to request.
- * @param {object} options - An options hash.
+ * @param {Partial<RequestInit>} options - An options hash.
  * @return {Promise} - A promise that resolves to the file contents.
  */
 async function httpLoader(url, options) {
--- a/packages/vega-loader/src/loader.js
+++ b/packages/vega-loader/src/loader.js
@@ -48,7 +48,7 @@ async function load(uri, options) {
 
   return opt.localFile
     ? this.file(url)
-    : this.http(url, options.http);
+    : this.http(url, options?.http);
 }
 
 /**
@@ -88,8 +88,8 @@ async function sanitize(uri, options) {
 
   // should we load from file system?
   loadFile = (isFile = uri.startsWith(fileProtocol))
-    || options.loaderMode === 'file'
-    || (options.loaderMode !== 'http')
+    || options.mode === 'file'
+    || (options.mode !== 'http')
       && !hasProtocol && fileAccess;
 
   if (isFile) {
--- a/packages/vega-loader/src/loader.js
+++ b/packages/vega-loader/src/loader.js
@@ -88,8 +88,8 @@ async function sanitize(uri, options) {
 
   // should we load from file system?
   loadFile = (isFile = uri.startsWith(fileProtocol))
-    || options.mode === 'file'
-    || (options.mode !== 'no-cors' && options.mode !== 'cors' && options.mode !== 'same-origin')
+    || options.loaderMode === 'file'
+    || (options.loaderMode !== 'http')
       && !hasProtocol && fileAccess;
 
   if (isFile) {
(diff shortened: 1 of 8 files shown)


domoritz:
I could make this backwards compatible with a hack but it feels like we should just rip the bandaid off now.

hydrosquall:
2 docs we could update alongside 

1. Readme.md for Vega-loader 
2. Types https://github.com/vega/vega/blob/main/packages/vega-typings/types/runtime/index.d.ts#L32-L33 

Re backwards compat - rather than supporting both, what do you think about logging a message or throwing an error? When the old option is detected? Doesn't use much code but still helps people know to update.

domoritz:
Good catch. I updated the docs and types.

I'm leaning against warnings because you already get an error if you pass `http` (at least I did because fetch says it's not a valid argument).

hydrosquall:
I think there might be a few uses of mode in the Vega-scenegraph tests 

https://sourcegraph.com/search?q=repo%3A%5Egithub%5C.com%2Fvega%2Fvega%24+mode&patternType=keyword&sm=0&__cc=1

I didn't know about the existing warning, I'll try it when I'm at a computer again

domoritz:
``` js
> fetch('www.google.com', {mode: 'http'})
Promise {
  <pending>,
  [Symbol(async_id_symbol)]: 71,
  [Symbol(trigger_async_id_symbol)]: 70
}
> Uncaught:
TypeError: Request constructor: http is not an accepted type. Expected one of navigate, same-origin, no-cors, cors.
```

domoritz:
Not sure how this worked until now tbh 😅

hydrosquall:
> ```js
> > fetch('www.google.com', {mode: 'http'})
> Promise {
>   <pending>,
>   [Symbol(async_id_symbol)]: 71,
>   [Symbol(trigger_async_id_symbol)]: 70
> }
> > Uncaught:
> TypeError: Request constructor: http is not an accepted type. Expected one of navigate, same-origin, no-cors, cors.
> ```
> Not sure how this worked until now 

I think I figured it out - `this.options.mode` is not passed to `fetch` it was only used to instantiate the loader. Only keys under `this.options.http` are  meant to be passed through to fetch.

https://github.com/vega/vega/blob/6ddd83a3988e0caff13d2a80e13139dc674a5d7b/packages/vega-loader/src/loader.js#L168

`loaderMode` is a custom enum we have in `vega` that is either file or http, which is not related to [`RequestInit.mode`](https://developer.mozilla.org/en-US/docs/Web/API/RequestInit#mode) . 

As a result, I don't think we can rely on the fetch error message to catch cases where folks might be using `mode` in the old way.

domoritz:
`const opt = extend({}, this.options.http, options)` does pass `options` as the last argument, though, no? So the full options object is being passed to the loader. 

So `loader.load` passes options to `http` to `opt` in `httpLoader` to `fetch`. I didn't walk through it with a debugger but that seems to be the path.

domoritz:
Ahh, I think there was a bug where we passed the loader options instead of just the http options in https://github.com/vega/vega/blob/18a55af6207fb183510659bdec3922eddc9a6022/packages/vega-loader/src/loader.js#L53.

hydrosquall:
> const opt = extend({}, this.options.http, options) does pass options as the last argument, though, no? So the full options object is being passed to the loader.

You're right, I didn't step back far enough. We are probably not erroring right now since `mode` is usually undefined, so nothing problematic gets passed through.
 
I think we should rewrite

````
const opt = extend({}, this.options.http, options)
```

to

```
const fetchOptions = extend({}, this.options.http)
```

since there isn't anything else in the requestInit that looks like it should be shared with our options.

domoritz:
@hydrosquall check the latest commit. I basically undid the changes and went back to mode. We now pass `options.http` instead of `options`. That seems to be consistent with the types and the readme.

packages/vega-loader/src/loader.js
```diff
@@ -89,7 +89,7 @@ async function sanitize(uri, options) {
   // should we load from file system?
   loadFile = (isFile = uri.startsWith(fileProtocol))
     || options.mode === 'file'
-    || (options.mode !== 'no-cors' && options.mode !== 'cors' && options.mode !== 'same-origin')
+    || (options.mode !== 'http')
```
hydrosquall, line 92: Makes sense

hydrosquall (approved):
I learned a lot more about vega-loader through reviewing this - thanks for investigating this with me!

packages/vega-loader/src/loader.js
domoritz: That's what it was before my changes yesterday. Thanks for helping me dig into this!

PR #4049, opened 2025-03-29
chore: remove sourcemaps for bundled files

domoritz:
They don't really help since we don't ship the sources in the package anyway. If you want sourcemaps, use ESM.

This change should make our packages smaller. See https://packagephobia.com/result?p=vega

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -156,7 +156,6 @@ export default function(commandLineArgs) {
       output: bundleOutputs({
         file: pkg.jsdelivr,
         format: 'umd',
-        sourcemap: true,
         globals,
         name: 'vega'
       }),
@@ -171,7 +170,6 @@ export default function(commandLineArgs) {
       output: bundleOutputs({
         file: pkg.jsdelivr.replace('.min.js', '-core.min.js'),
         format: 'umd',
-        sourcemap: true,
         globals,
         name: 'vega'
       }),


hydrosquall (approved):
I think a second impact is this will speed up the build times too! I don't have metrics before/after metrics on what this looked like before (perhaps this is something we could start tracking in github actions). 

From past work, I've seen sourcemaps generation may add noticeable overhead to compile time, and that we might want to save those for dev rather than production mode.

## Disk savings

image

It looks like we'll save close to 6 MB by not shipping sourcemaps per [runpkg](https://www.runpkg.com/?vega@6.1.2/build/vega.min.js) - nice!

image

Unfortunately the install footprint still seems like it's over 20 MB, but this might be something we get to address by addressing the maps on the various sub dependencies of vega

image

## Meta

Node.js test is [failing](https://github.com/vega/vega/actions/runs/14146185120/job/39635531569?pr=4049) on vega-interpreter, but I think it's unrelated to this change, so we can investigate it separately.

domoritz:
> Node.js test is [failing](https://github.com/vega/vega/actions/runs/14146185120/job/39635531569?pr=4049) on vega-interpreter, but I think it's unrelated to this change, so we can investigate it separately.

There is a flaky test. I just kept rerunning and then tests pass. Yes, would be lovely to investigate.

Issue #3903, opened 2024-04-12
Align legends to the left of the charts


domoritz:
When placing a legend at the top of a chart, it aligns with the left of the plotting area/axis but not the chart. This means that it's inconsistent with how titles align.

[Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEySYUqUAwBOgtCrVJOcNZkFx+AdxP042ylAjEQAXyXJtAa31Mk2xSBwslBsyjSyZGigAB5RIJg0CHAAqrJ0+ghy9AEAZjRwgsr6qphwAZgAnjhl6GzaYbJIekoJZjUgALJZMAAEbDk9tj0VJv5uIBVxSGRk2nyl+iEMspjlVe0AjgxIK3RqNKSuSiGCdXF5BUXo1mrw-i3r+rJsCOFNAdBNNaDKL0jhaAA2iAIMtcmxIkplNoaAAvWHmALaf7yJQQZ6WEAAXSUyIiNWBAGI4AB2XxiABMAUJUBJtNp1KQcFpcAAHNSAIw5ElkgAs1IAnLyAGxkopY8atRHoADqNjug0eSnMZCCV1AdXyK30mDYOFcLkldGloB2sDO6Ag6m0qxacGiq1l8rsPXCPQAyjY2gagA)

<img width="307" alt="Screenshot 2024-04-12 at 15 49 25" src="https://github.com/vega/vega/assets/589034/3a7931de-22a6-4068-8d1d-5fadcd5cba4c">

It would be nice to be able to align the legend with the left side of the chart.

[...]

PR #4052, opened 2025-03-30
feat: align label to start of chart

djbarnwal:
Addresses https://github.com/vega/vega/issues/3903

Added a property named `frame` similar to that of `title` which takes the value of `group` or `bounds` to align the legend to the chart when `orient` is either `top` or `bottom`. 

With `orient: top`, `anchor:start` and  `frame: group` (default)
image

With `orient: top`, `anchor:start` and  `frame: bounds`
image

With `orient: top`, `anchor:end` and  `frame: bounds` and axes Y `orient:right`
image

### DIFF
--- a/packages/vega-typings/types/spec/config.d.ts
+++ b/packages/vega-typings/types/spec/config.d.ts
@@ -9,7 +9,6 @@ import {
   Orientation,
   Padding,
   RangeScheme,
-  StringValue,
   SymbolShape,
   TextBaseline
 } from '../index.js';
@@ -598,7 +597,7 @@ export interface BaseLegendLayout {
   /**
    * The reference frame for the anchor position, one of `"group"` (the default, to anchor relative to the group width or height) or `"bounds"` (to anchor relative to the full bounding box).
    */
-  frame?: TitleFrame | StringValue;
+  frame?: TitleFrame;
 
   /**
    * A flag to center legends within a shared orient group.
--- a/packages/vega-typings/types/spec/config.d.ts
+++ b/packages/vega-typings/types/spec/config.d.ts
@@ -9,6 +9,7 @@ import {
   Orientation,
   Padding,
   RangeScheme,
+  StringValue,
   SymbolShape,
   TextBaseline
 } from '../index.js';
@@ -29,7 +30,7 @@ import { BaseLegend } from './legend.js';
 import { Locale } from './locale.js';
 import { BaseProjection } from './projection.js';
 import { InitSignal, NewSignal, SignalRef } from './signal.js';
-import { BaseTitle, TitleAnchor } from './title.js';
+import { BaseTitle, TitleAnchor, TitleFrame } from './title.js';
 
 export type KeepSignal<T> = T extends SignalRef ? SignalRef : never;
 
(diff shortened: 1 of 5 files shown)


domoritz:
Thanks for taking a stab at this. Would frame (similar to titles) be the more fitting term?

djbarnwal:
> Thanks for taking a stab at this. Would frame (similar to titles) be the more fitting term?

Sure, makes sense to adopt already existing property definitions. I have updated the PR to use `frame` as the property.

packages/vega-view-transforms/src/layout/legend.js
```diff
@@ -26,6 +26,7 @@ export function legendParams(g, orient, config, xb, yb, w, h) {
   const _ = lookup(config, orient),
         offset = offsets(g, _('offset', 0)),
         anchor = _('anchor', Start),
+        frame = _('frame', Group),
```
hydrosquall, line 29: Thanks for the new param! For this to show up in the JsonSchema, I believe we'll need to add this to around here in the Typescript types: 

https://github.com/vega/vega/blob/7862e6db14ae2fb02b21444921d7de60160a36a2/packages/vega-typings/types/spec/config.d.ts#L586

djbarnwal: Thanks for the review, I have added types for Legend Layout similar to that of `frame` property of Title.

packages/vega-typings/types/spec/config.d.ts
```diff
@@ -594,6 +595,11 @@ export interface BaseLegendLayout {
    */
   bounds?: LayoutBounds;
 
+  /**
+   * The reference frame for the anchor position, one of `"group"` (the default, to anchor relative to the group width or height) or `"bounds"` (to anchor relative to the full bounding box).
+   */
+  frame?: TitleFrame | StringValue;
```
hydrosquall: I'm under the impression from the comment just "group" and "bounds" are supported, which is covered by `TitleFrame`. What would be the case for supporting a `StringValue` on this prop?

packages/vega/test/specs-valid/scatter-plot-guides.vg.json
```diff
@@ -154,10 +154,20 @@
     }
   ],
 
+  "config": {
+    "legend": {
+      "layout": {
+        "frame": "bounds"
```
hydrosquall, line 160: This new default makes sense to me based on the picture in the PR description 👍

hydrosquall (approved):
Thanks for following up on this suggestion, the change looks good to me. 

I have one last question about the type signature on this new prop, but the change looks reasonable to me 👍

packages/vega-typings/types/spec/config.d.ts
djbarnwal: I was following the same type convention used for the title layout, which also supports `StringValue`. If it's not relevant I can remove it.

https://github.com/vega/vega/blob/398933a49e8249a41ad8aee65f99935346b67e71/packages/vega-typings/types/spec/title.d.ts#L71-L74

hydrosquall: The code won't crash if some random string is passed through, it'll just get the `default` behavior of `group`, but it won't do anything useful. I would vote to remove StringValue from both places, but if you want to keep the radius of this PR small just changing it in this location is fine too since I haven't gone to check the implementation of the `title` frame beforehand.

djbarnwal: Good point. Have removed the `StringValue` reference from layout's frame. I don't have a fair idea of why it was added for `title` so following Chesterton’s Fence, will keep it as is for now.

PBI-David:
@hydrosquall - is there anything stopping this from being merged?

PR #4053, opened 2025-03-30
chore: delete map files and old compatibility libraries

domoritz:


hydrosquall (approved):
Seems like a good idea to me so that default IDEs don't have to search through these files.  Noting that `vega-projections.min.js` does not seem like  a `.map` file just in case that was removed accidentally.

Taking it further, I think we might able to eventually remove the entire docs/assets folder from the main branch, and only have those on the github pages branch.

PR #4061, opened 2025-04-14
fix: support animated vega-lite thumbnail rendering with vg2png and vg2svg

jonathanzong:
this change fixes thumbnail generation for the animated examples in the vega-lite example gallery by evaluating the dataflow first so that the animation initializes to its first frame before finalizing and rendering

needed for VL animation docs: https://github.com/vega/vega-lite/pull/9535


packages/vega-cli/src/render.js
```diff
@@ -43,10 +43,19 @@ export default function(type, callback, opt) {
...
-        }).finalize(); // clear any timers, etc
-        return (type === 'svg'
-            ? view.toSVG(scale)
-            : view.toCanvas(scale * ppi / 72, opt)).then(_ => callback(_, arg));
+        });
+        // Run the dataflow to evaluate signals first (this returns a Promise)
+        return view.runAsync()
+            .then(() => {
+                // Now finalize after the signals have been evaluated
+                view.finalize();
```
domoritz, line 51: Shouldn't this happen after we are done rendering? https://vega.github.io/vega/docs/api/view/#view_finalize

domoritz (approved):

domoritz:
Do you think something like this can also help us fix https://github.com/vega/vega-lite/issues/9543?

domoritz:
@jonathanzong @arvind could you help wrap this up?

packages/vega-cli/src/render.js
jonathanzong: my change preserves the order of operations that was there before, and it's currently working, so i'm inclined to leave this

jonathanzong:
tests pass locally, and i think they were passing in april. any advice?

domoritz:
@jheer @arvind there seems to be a range condition for selection initialization. Vega initializes the signals mostly right (before the first render) but the dependent datasets (for selection testing) are not always initialized before the first render. That made me wonder, do you recall why `vlSelectionTest` uses a dataset as input rather than a signal?

joelostblom:
I asked Arvind about this on slack and got the following response:

> Interesting. Yes, there are a handful of times the existing selections code experienced race conditions—I'll need to sweep the (significant) cobwebs to remind myself how I got around that; I think it was often via some gating signal.
>
>Wrt to the specific question re:dataset, it's basically to performantly handle when the representation of the selection involves an array (i.e., multiple things are selected) which happens quite often: selections in multiview (eg multiple brushes in a SPLOM); brushes on maps (which hit test to select the underlying points) and then doing the dataflow + rendering recomputation performantly as eg points at the margins fall into/out of the selection.
> 
> This would be much hairier to implement via signals alone because—at least "back in my day"—signals were not really designed to take array-type data; it was always meant for scalar/object data. And it would certainly be less performant since it couldn't do any changeset/diff-based updates.

@domoritz @jonathanzong Is that enough to help us get unstuck here?

domoritz:
I was talking with @yuvvantalreja about this issue this morning. Good timing and thanks for the additional detail.

jonathanzong:
ok, i threw opus 5 at it and think we're good. the remaining CI failure seems to be an unrelated test about basemap.

the tl;dr — the Vega-Lite compiler needs to emit `update` instead of `on` because `on` doesn't run on the first tick. (**This will require a small PR in Vega-Lite.** It won't affect anything other than animation-related code.)Also, I made the race condition worse by moving the `finalize()` call. It looks like render.js didn't need to be changed at all in the end, so let me know if you want me to revert the cosmetic changes made by claude (i do think leaving a new comment there telling the next person not to move finalize is worth it)

some slop explanation:

> **Root cause**. It isn't really a race in Vega's dataflow — it's that in the committed fixture, animation_frame_tuple and animation_frame_modify are compiled as on: event handlers rather than update: expressions:
> 
> {"name": "animation_frame_tuple",  "on": [{"events": [{"signal": "eased_anim_clock"}, ...], ...}]}
> {"name": "animation_frame_modify", "on": [{"events": {"signal": "animation_frame_tuple"}, ...}]}
> 
> on: handlers are event-driven — by design they do not fire during the initial pulse. So after the first run, animation_frame_tuple === undefined, modify() is never called, and animation_frame_store is empty. Trace at clock=1:
> 
> SIGNAL animation_frame_tuple  = undefined
> SIGNAL animation_frame_modify = undefined
> < eval#1 store=[] curr=682          ← all 682 rows, no frame filter
> 
> The store only ever fills because a timer tick happens to land and drive anim_clock 0→1. Whether that tick arrives before serialization is luck — and it depends on how long the remote gapminder fetch took, which is why it looked flaky and load-dependent. With timers disabled it never initializes.
> 
> The update: variant you pasted is correct at clock=1, deterministically:
> 
> SIGNAL animation_frame_tuple  = {unit:"", fields:[...], values:[1955]}
> SIGNAL animation_frame_modify = 1
>   > eval#2   DATA animation_frame_store = [[1955]]   ← modify()'s runAfter pulse
> < eval#1 store=[[1955]] curr=62
> 
> I confirmed this is the only difference by converting just those two signals in your fixture, changing nothing else — it went from 692 paths to 72.
> 
> **On modify() being one pulse behind**: that part is real but benign. modify() queues its changeset via df.runAfter(..., true, 1), so the store is written by a follow-up pulse. That pulse propagates synchronously inside the first run's postrun, so it lands before runAsync() resolves. It's a genuine second-order hazard, not your bug.
> 
> **The branch's render.js change made it worse**: moving finalize() to after runAsync() opened a window where timers run during the data load.
> 


Could someone help me figure out the failing basemap test thing before I go any further in case there's a deeper issue? (EDIT: https://github.com/vega/vega/pull/4315)

jonathanzong:
ready for review

domoritz (commented):
Do we still need this after https://github.com/vega/vega-lite/pull/9900? https://github.com/vega/vega/pull/4333 seems to imply it's not really needed.

jonathanzong:
Right, the only thing this PR adds at this point is tests for gapminder, which #4333 seems to do equivalently, so I'm fine with closing this

PR #4063, opened 2025-04-23
Pbi david patch 1

PBI-David:
Previous version did not account for changes to outer radius. 

Apologies @domoritz .


PBI-David:
Ignore

PR #4104, opened 2025-08-09
Quantize like CSS justify-between

mkdynamic:
/**
 * Returns `count` evenly spaced samples from an interpolator.
 * - count <= 0 → []
 * - count === 1 → [interpolator(0.5)]
 * - count >= 2 → includes both endpoints (justify-between) */

### DIFF
--- a/packages/vega-scale/src/interpolate.js
+++ b/packages/vega-scale/src/interpolate.js
@@ -15,9 +15,13 @@ export function interpolateColors(colors, type, gamma) {
 }
 
 export function quantizeInterpolator(interpolator, count) {
-  const samples = new Array(count),
-        n = count + 1;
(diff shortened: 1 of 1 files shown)


hydrosquall:
Hey @mkdynamic , thanks for reaching out with this change, I didn't know CSS behaves this way. 

Could you share a bit more about the graphing use-case where you would like the interpolator to behave this way? 

I don't know if it was intentional for the old behavior to avoid touching the boundaries of (0 and 1)

If you're up for demonstrating the new behavior programmatically, this is the [folder](https://github.com/vega/vega/tree/f69c3facf5906e34ffaa35b156b42390085356f4/packages/vega-scale/test) where we would store test cases for the scaling function.

PR #4137, opened 2025-09-22
ci: test on main, not master

domoritz:


hydrosquall (approved):
Makes sense since master redirects to main: https://github.com/vega/vega/tree/master

PR #4147, opened 2025-10-01
fix: improve expressions

markov00:


hydrosquall:
Basemap test was erroring on the first go for node 21, but it passed after 2 retries.

Do you think we should update the package-lock.json as well (since that's what should ensure consistent installs when cloning/running locally), or just updating the package.json are enough? cc @domoritz

domoritz:
* [ ] package-lock.json should be updated. If we run `npm ci` it should error when the locale is not up to date
* [ ] remove any kind of deployment scripts for the website so that we don't accidentally override the current version.

domoritz:
Before making a release, we need to check the scripts to delete anything that would override release artifacts like the schema or website. 

But happy to merge this pr as is.

PR #4150, opened 2025-10-03
fix: updated stdevp and variancep for population 1 case

### LINKED ISSUE #4149: stdevp and variancep ops return undefined when population size is 1
Both operators should return 0 when the population size is 1 as seen in the code below:
https://github.com/vega/vega/blob/28bd0fe257e3444e754a66599cda858f8a32d505/packages/vega-transforms/src/util/AggregateOps.js#L63-L66 line 64 should be ```m => m.dev / m.valid```
https://github.com/vega/vega/blob/28bd0fe257e3444e754a66599cda858f8a32d505/packages/vega-transforms/src/util/AggregateOps.js#L71-L74 and line 72 should be ``m => `Math.sqrt(m.dev / m.valid)``

soldenstoll:
fixes #4149

### DIFF
--- a/packages/vega-transforms/src/util/AggregateOps.js
+++ b/packages/vega-transforms/src/util/AggregateOps.js
@@ -61,7 +61,7 @@ export const AggregateOps = {
     req:  ['mean'], idx: 1
   },
   variancep: {
-    value: m => m.valid > 1 ? m.dev / m.valid : undefined,
+    value: m => m.dev / m.valid,
     req:  ['variance'], idx: 2
   },
   stdev: {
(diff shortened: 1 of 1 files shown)


packages/vega-transforms/src/util/AggregateOps.js
```diff
@@ -61,15 +61,15 @@ export const AggregateOps = {
     req:  ['mean'], idx: 1
   },
   variancep: {
-    value: m => m.valid > 1 ? m.dev / m.valid : undefined,
+    value: m => m.dev / m.valid,
     req:  ['variance'], idx: 2
   },
   stdev: {
     value: m => m.valid > 1 ? Math.sqrt(m.dev / (m.valid - 1)) : undefined,
```
domoritz, line 68: Why not change this as well?

domoritz (commented):
It feels like this might have been intentional. Did you try to find out why it was like that?

packages/vega-transforms/src/util/AggregateOps.js
soldenstoll: stdev is the sample standard deviation which will have a division by 0 error if ```m.valid``` is 1. The population standard deviation divides by ```m.valid``` instead of ```m.valid - 1```, so this isn't an issue there.

Issue #4151, opened 2025-10-02
Diverging color scale with domainMid not in the domain is awkward


kanitw:
## Bug Description

Example: if the data range from 19 to 91, but the domainMid is -20, it's producing a funny scale 

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v6.json",
  "description": "A simple bar chart with embedded data.",
  "data": {
    "values": [
      {"a": "A", "b": 28}, {"a": "B", "b": 55}, {"a": "C", "b": 43},
      {"a": "D", "b": 91}, {"a": "E", "b": 81}, {"a": "F", "b": 53},
      {"a": "G", "b": 19}, {"a": "H", "b": 87}, {"a": "I", "b": 52}
    ]
  },


    "mark": "bar",
    "encoding": {
      "x": {"field": "a", "type": "nominal", "axis": {"labelAngle": 0}},
      "y": {"field": "b", "type": "quantitative"},
      "color": {
        "field": "b", "type": "quantitative",
        "scale": {"type": "linear", "scheme": "redblue", "domainMid": -20}
      }
    }
}

```

<img width="297" height="361" alt="Image" src="https://github.com/user-attachments/assets/cd63ce33-c08d-4759-b210-6baef0c39d53" />

Expected: 

the mid should be mapped to the middle of the ramp

<img width="287" height="346" alt="Image" src="https://github.com/user-attachments/assets/7c6dfe9b-6866-4b37-bf6f-377a2e902967" />

Note that we if I add another layer with datum = -20 

```
{
  "$schema": "https://vega.github.io/schema/vega-lite/v6.json",
  "description": "A simple bar chart with embedded data.",
  "data": {
    "values": [
      {"a": "A", "b": 28}, {"a": "B", "b": 55}, {"a": "C", "b": 43},
      {"a": "D", "b": 91}, {"a": "E", "b": 81}, {"a": "F", "b": 53},
      {"a": "G", "b": 19}, {"a": "H", "b": 87}, {"a": "I", "b": 52}
    ]
  },
  "layer": [{

    "mark": "bar",
    "encoding": {
      "x": {"field": "a", "type": "nominal", "axis": {"labelAngle": 0}},
      "y": {"field": "b", "type": "quantitative"},
      "color": {
        "field": "b", "type": "quantitative",
        "scale": {"type": "linear", "scheme": "redblue", "domainMid": -20}
      }
    }
  }, {
    "mark": "text",
    "encoding": {
[...]

PR #4126, opened 2025-09-04
chore: v6.2.0

lsh:
Changes since v6.1.2

**monorepo**:
* Use blazediff to speed up visual diffs (thanks @teimurjan!) (via #4125)
* Update version in specs (thanks @domoritz!) (via #4045)

**vega**:
* Export schema from Vega package (thanks @domoritz!) (via #4046)

**vega-loader**:
* Correct mode handling in loader (thanks @domoritz!) (via #4048)

**vega-util**:
* Fix handling of null values in stringValue (thanks @domoritz!) (via #4127)
* Replace all problematic Unicode characters (thanks @basbroek!) (via #4130)

**docs**:
* Update links to Prefuse and Protovis (thanks @star1327p!) (via #4095)
* Replace version in specs (thanks @domoritz!) (via #4085)
* Update file name in weather example (thanks @domoritz!) (via #4083)
* Update versions in docs (thanks @domoritz!) (via #4050)
* Update of the labelled donut example (thanks @PBI-David!) (via #4064)
* Addition of the labelled donut example (thanks @PBI-David!) (via #4062)

PR #4152, opened 2025-10-03
#4151 -- improve domainMid when it's less or more than max

kanitw:
See #4151


docs/vega-core.js
```diff
@@ -22579,8 +22584,6 @@
       globalvar = opt.globalvar,
       fieldvar = opt.fieldvar,
       outputGlobal = isFunction(globalvar) ? globalvar : id => `${globalvar}["${id}"]`;
-    // JSON authors are not allowed to set properties with these names, as these are built-in to the JS Object Prototype.
-    new Set([...Object.getOwnPropertyNames(Object.prototype).filter(name => typeof Object.prototype[name] === 'function'), '__proto__']);
```
mattijn, line 22583: This seems an unrelated deletion. If I remember right this was purposely introduced recently #4126

PR #4156, opened 2025-10-21
fix: replace dots with underscores in aggregated fields

basbroek:
## Summary
This PR fixes sorting on aggregated fields with (escaped) dots in the target field name by replacing these dots with underscores in aggregated field names.

## Problem
Sorting on aggregated fields doesn't work or results in errors if the target field contains (escaped) dots or is truly nested, as these examples show: field with dot ([open in editor](https://vega.github.io/editor/#/url/vega/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykSAbJQBWENgDsQAGhAB3GgBN6aAMwAGNTPg0yWNAFZNIHEgUKa4sgZlIGmNhBoAvOGhAAzOtJAKkmFKgA2qDiSAiu6P5MADauMsRI0QxwEGjBIFB+bhAwbAzRCgAETHDFcNFsst7y9JQKbJhoAIxqAL5SoJmN6Dl5BWWFSExspNV0MHUNqm0Auq0zMsgATgDWqUGgmACeOBEgS3BQjTLuS2wIaKC+-m5RsSDtIHDiUGwKEaAMONcfIAAelxA0ESewB8USyTQbRkfwATIDgfd0GCPDRygo3DUYAAdbGTRqPLYIzJIkBEk5ogpuLoPLRwHR6VCgRF7ckgJhIcQY1BNVp8+YyFnrdKhcJuFHbXZuaIWOBIJbeerICyA64BEB3OKo9GY8a4-G0-acsh7eRKGDecQ0KARTBLZLtEJhVneSV7DlcxXnJAqpk+PzqzXeTw69A0wVsJbdUBsHDZBgXCmhuR6vH1Y4gSPvBXod7QZ7mSwPR5LY17bS6DMmMwWKyoNSUXkLEBIP4pNLMkkumSRtHibogWLuAlzIA)), nested field ([open in editor](https://vega.github.io/editor/#/url/vega/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykSAbJQBWENgDsQAGhAB3GgBN6aAMwAGNTPg0yWNAFZNIHEgUKa4sgZlIGmNhBoAvOGhAAzOtJAKkmFKgA2qDiSAiu6P5MADauMsRI0QxwEGjBIFB+bhAwbAzRCgAETHDFcNFsst7iKZhwCmigCUkRAIxqAL4dUqCZmNm5+UUlhUhMbKTVtfWNIM3Jqp0dALorMsgATgDWqUGgmACeOBEgG3BQ-TLuG2wIs77+blGxIN0gcOJQbAoRoAw4D1+IAAHrNoIkTqD4okFqhOjJgQAmMGZF7oKEeGjlBroGoQOoKSjzVxvA4oiFuMlXLEFNx9V5aOA6PSoUDgtEgKkgJhIcQ41pdLrLGTslJpEJhSHeQ7HNzRCxwJAbbwKW5ICz3PwBEDPOKY7FuPEEokwkkyDa8sgneRKGDVGhQCKYDbJboS8KU6VHE48vkqtUa1k+LVPMYvakG9D0kVsDb9INsHBuZAYzyRkBG+omlreWM-ZXoH7QD7mSyvN4WywnbS6S7GUylqxwygC4UgJDAsV7ECiz0yWNY8TxkCxdz9FYdIA)).

## Solution
In the function ```aggrField``` in ```util.js``` all (escaped) dots are replaced with underscores, so for example
```"op": "sum", "field": "[a.b]"``` becomes 'sum_a_b', 
```"op": "mean", "field": "a\\.b"``` becomes 'mean_a_b,
```"op": "max", "field": "a.b"``` becomes 'max_a_b.

## Fixes
[Vega-Lite issue](https://github.com/vega/vega-lite/issues/7546): Sort by aggregate sum does not work with nested object

### DIFF
--- a/packages/vega-parser/src/util.js
+++ b/packages/vega-parser/src/util.js
@@ -65,7 +65,9 @@ export function aggrField(op, field) {
         ? '$' + field.signal
         // Replace non-alphanumeric character sequences with underscores and trim leading/trailing underscores
         // to prevent incorrect path extraction for nested target fields or target fields with (escaped) dots. 
-        // example: 'a\\.b[c.d]' => 'a_b_c_d'
+        // Example: 'a\\.b[c.d]' => 'a_b_c_d'. 
+        // Note: aggregating both a nested field and a field with a dot could lead to conflicting names: 
+        // with data like [{ a: {b: 1}, 'a.b': 1 }], summing 'a.b' and 'a\\.b' would both result in a field 'sum_a_b'   
         : field?.replace(/\W+/g, '_').replace(/^_+|_+$/g, '') || '');
 }
 

### CODE REVIEW

File: packages/vega-parser/src/util.js [hunk 1]
```diff
@@ -61,7 +61,13 @@ export function sortKey(sort) {
...
   return (op && op.signal ? '$' + op.signal : op || '')
     + (op && field ? '_' : '')
-    + (field && field.signal ? '$' + field.signal : field || '');
+    + (field && field.signal 
+        ? '$' + field.signal 
+        // replace dots with underscores in aggregated fields to prevent incorrect path extraction
+        // "op": "sum", "field": "[a.b]" => 'sum_a_b' 
+        // "op": "mean", "field": "a\\.b" => 'mean_a_b
+        // "op": "max", "field": "a.b" => 'max_a_b
+        : field?.replace(/[.[\]\\]/g, '_') || '');
```
domoritz: Should we use `varName` from utils instead?

basbroek: Isn't `varName` a VL util? That function would also replace brackets and backslashes with underscores, resulting in `sum__a_b_` instead of `sum_a_b` and `mean_a__b` instead of `mean_a_b`.

domoritz: Ups, This is vega, not Vega-Lite. I just wanted to see whether we have a method that works already but looks like not so this is good. 

Let's add a unit test and then we can merge.

basbroek:
Thanks for reviewing. I refined the regex based on VL's `varName` util to catch some edge cases and added a unit test.

packages/vega-parser/test/transform-test.js
```diff
@@ -52,3 +53,17 @@ tape('Parser parses Vega specs with data transforms', t => {
 
   t.end();
 });
+
+tape('Parser replaces non-alphanumeric characters with underscores in aggregated fields', t => {
+  // example data: [{ a: { b: 1 }, 'c.d': 1, 'e.f': { g: 1, 'h.i': 1} }]
+  t.deepEqual(aggrField('sum', 'a.b'), 'sum_a_b');
+  t.deepEqual(aggrField('mean', 'c\\.d'), 'mean_c_d');
```
domoritz, line 60: What I'd we have a nested d inside an object c as well? Wouldn't we get a conflict?

domoritz: This may be okay but we should at least be aware when we get collapsing properties.

basbroek: You're right, with data like this `[{ a: {b: 1}, 'a.b': 1 }]` aggregating both `'a.b'` and `'a\\.b'` or `[a.b]` would result in the same field name `sum_a_b`. A possible solution for this edge case could be using double underscores for escaped dots. The function below would do that, but feels a bit over-engineered. What do you think?

This code...

```
function normalizeField(field) {
  if (!field) return '';

  return field
    // escape all dots within brackets
    .replace(/\[([^\]]*?)\]/g, (_, inner) => `[${inner.replace(/\./g, '\\.')}]`)
    // replace all escaped dots with a double underscore
    .replaceAll('\\.', '__')
    // replace all regular dots with a single underscore
    .replaceAll('.', '_')
    // remove quotes
    .replaceAll('\'', '')
    // replace brackets with underscores
    .replaceAll('][', '_')
    .replaceAll('[', '_')
    .replaceAll(']', '_')
    // trim leading/trailing underscores
    .replace(/^_+|_+$/g, '');
}
```

... passes these tests:

```
// example data: [{ a: { b: 1 }, c: {d: 1} 'c.d': 1, 'e.f': { g: 1, 'h.i': 1} }]
t.deepEqual(aggrField('sum', 'a.b'), 'sum_a_b');
t.deepEqual(aggrField('mean', 'c.d'), 'mean_c_d');
t.deepEqual(aggrField('median', 'c\\.d'), 'median_c__d');
t.deepEqual(aggrField('average', '[c.d]'), 'average_c__d');
t.deepEqual(aggrField('min', 'e\\.f.g'), 'min_e__f_g'); 
t.deepEqual(aggrField('max', "e\\.f['g']"), 'max_e__f_g'); 
t.deepEqual(aggrField('mean', "[e.f]['g']"), 'mean_e__f_g'); 
t.deepEqual(aggrField('average', '[e.f][h.i]'), 'average_e__f_h__i'); 
t.deepEqual(aggrField('median', 'e\\.f.h\\.i'), 'median_e__f_h__i'); 
t.deepEqual(aggrField('distinct', 'e\\.f[h.i]'), 'distinct_e__f_h__i'); 
```

domoritz: Let's keep it simple but document the limitation for example with a commented out test case. 

I'd say it's okay since nested aggregates don't work right anyway and having both nested and data with "." in the name would be unlikely and honestly I'd be surprised if it didn't cause issues in other places.

packages/vega-parser/src/util.js
```diff
@@ -61,7 +61,12 @@ export function sortKey(sort) {
 export function aggrField(op, field) {
   return (op && op.signal ? '$' + op.signal : op || '')
     + (op && field ? '_' : '')
-    + (field && field.signal ? '$' + field.signal : field || '');
+    + (field && field.signal
+        ? '$' + field.signal
+        // Replace non-alphanumeric character sequences with underscores and trim leading/trailing underscores
```
domoritz, line 66: What's wrong with trailing or leading _, btw?

packages/vega-parser/test/transform-test.js
basbroek: Alright, I documented the potential conflict in the existing comment.

packages/vega-parser/src/util.js
basbroek: Without trimming the name of the aggregated field would differ depending on the notation of the target field: `'c\\.d'` =would become `'sum_c_d'` and  `'[c.d]'` would become  `'sum___c_d_'`, which I think is undesirable.

domoritz: I see. Yeah, that makes sense. Although, I'd say that it's up to the user to aim for consistent references to nested fields so we may not consistently support the different forms.

PR #4172, opened 2025-11-09
docs: Correct grammar issues especially on a/an usage

star1327p:
Correct grammar issues especially on a/an usage


joshpoll (approved):
lgtm!

Issue #3971, opened 2024-09-08
convert packages to Typescript


hydrosquall:
## Motivation

- Vega was written in JS before the wide adoption of Typescript (1.0 in 2014). As a result, separate `d.ts` are manually created and sync'd with the type of files. Occasionally the types get out of sync with the types ( #3954 )

## Proposal

We could converted some of vega sub-packages to TS. This way we won't have to sync .d.ts files with JS source code / can delete the `d.ts` files entirely . Many of the newer vega projects are written in typescript (including Vega-Lite). 

## Contraints

- We need to make sure doing this doesn't degrade the developer experience. For example 
  - if I change code in a linked Vega package, the Vega editor (in local dev mode) should update. (via @domoritz in https://github.com/vega/vega/issues/3957#issuecomment-2334737308 )
  - Unit tests should still work
- It might be unrealistic to migrate the entire vega repository at once. We could try this a small package at a time.

domoritz:
We should do this incrementally and could start with Vega utils since it already has type declarations. I originally did this a while ago but it's a good time to do it again: https://github.com/vega/vega/pull/1566.

hydrosquall:
Some of the smaller packages (vega-time, vega-format) could also be decent candidates for pilot conversion, given their relative smallness / status as "pure functions".

hydrosquall:
Umbrella thread for all `vega-util` migration project

I set this in motion after doing a poc migrating everything in 1 pass to see if the result builds, before splitting the diff for closer scrutiny/safety

https://github.com/vega/vega/pull/4198

### Infra / Validation

Used JSDoc to validate the approach

- https://github.com/vega/vega/pull/4184
- https://github.com/vega/vega/pull/4185
- https://github.com/vega/vega/pull/4187
- https://github.com/vega/vega/pull/4188
- https://github.com/vega/vega/pull/4189

## Actual migrations 

Easy cases

- https://github.com/vega/vega/pull/4200
- https://github.com/vega/vega/pull/4202

Subtle behavior changes / generic types , added tests along the way

- https://github.com/vega/vega/pull/4203
- https://github.com/vega/vega/pull/4218
- https://github.com/vega/vega/pull/4219

PR #4184, opened 2025-11-24
chore: update node / typescript configurations at vega root

hydrosquall:
## Motivation

- Recent security issues raised the risk of having types, comments, and implementation out of date (a param described as string actually accepting objects)
- Let's reduce the frequency shifting some of the simpler parts (no special types, just the ones operating on standard JS number/string etc) over to TS to pave the way for future larger scale typings
- #3971 

## Changes

- Node 18 lost support in April 2025, using older versions is not playing well with latest typescript. Upgrading to the currently supported version of node. See the [beware](https://nodejs.org/en/blog/announcements/node-18-eol-support) notice
- Updated test matrix to test version 20 and 23 instead of 20 / 22 
- Updated the root tsconfig w/ newer properties 

## Testing

- CI should continue to pass. Running `npm run typecheck` will be a no-op for now (it completes in 4s), but it will become useful as some packages gradually adopt native types, similar to vega-lite .

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "serve": "serve packages/vega/ -l 8080",
     "test": "npm run test:no-lint && npm run lint",
     "test:no-lint": "lerna run test --ignore vega-typings && lerna run test --scope vega-typings",
-    "typecheck": "tsc --noEmit",
+    "typecheck": "tsc",
     "lint": "eslint -c eslint.config.js .",
     "format": "npm run lint -- --fix",
     "postinstall": "scripts/postinstall.sh"


domoritz (approved):

package.json
```diff
@@ -18,6 +18,7 @@
     "serve": "serve packages/vega/ -l 8080",
     "test": "npm run test:no-lint && npm run lint",
     "test:no-lint": "lerna run test --ignore vega-typings && lerna run test --scope vega-typings",
+    "typecheck": "tsc --noEmit",
```
domoritz:
```suggestion
    "typecheck": "tsc",
```

It's redundant

tsconfig.json
```diff
@@ -0,0 +1,26 @@
+{
+  "compilerOptions": {
+    "target": "ES2019",
```
domoritz: We can probably target something newer than that.

hydrosquall: We do 2022 for vega-lite, how do you feel about having that for consistency?

https://github.com/vega/vega-lite/blob/e7f55118e7c149c42436c800d053423cdd10bb1a/tsconfig.json#L3

Ok to not decide right away since we're not compiling anything yet, just running the typecheck part.

domoritz: Makes sense.

package.json
domoritz:
```suggestion
    "typecheck": "tsc",
```

PR #4185, opened 2025-11-24
chore(types): JSDoc types for vega-util string processing

hydrosquall:
## Motivation

- Adding JSDoc param types allows us to get some type safety without having to switch to the typescript compiler (.ts files) yet
- Builds on https://github.com/vega/vega/pull/4184

## Changes

- Adds type signatures to the 5 string processing utilities, including 3 sub-utilities they depended on

## Testing

- If you open this locally and edit one of the files remove the types, then run `npm run typecheck`, the job should fail

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -3,7 +3,7 @@
     "target": "ES2022",
     "module": "ESNext",
     "lib": [
-      "es6",
+      "es2021",
       "dom"
     ],
     "moduleResolution": "Node",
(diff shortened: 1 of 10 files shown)


tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "target": "ES2019",
     "module": "ESNext",
     "lib": [
-      "es6",
+      "es2021",
```
hydrosquall, line 6: `vega-util`'s stringValue has `replaceAll` , which requires es2021 as the baseline: 

https://www.w3schools.com/js/js_2021.asp

domoritz:
@hydrosquall feel free to merge via the cli instead.

hydrosquall:
I amended the last commit and force pushed to get us out of a bad state, which seems to have worked !

I try to not to merge locally via CLI since it bypasses the branch protections on `main`

PR #4187, opened 2025-11-30
chore(vega-util): type all type checking functions

hydrosquall:
## Motivation

- Adding JSDoc param types allows us to ge type safety without having to switch to the typescript compiler (.ts files) yet
- Builds on https://github.com/vega/vega/pull/4185 / https://github.com/vega/vega/issues/3971

## Changes

- Documented every `is*` helper in `packages/vega-util/src` with JSDoc guards and kept the `Array.isArray` wrapper for the templated signature.
- Extended `packages/vega-util/tsconfig.json` to include `src/is*.js`, regenerated `build/types`, and rewired `packages/vega-util/index.d.ts` to export all generated declarations.

## Testing

- Verified `npm run generate-types` and `npm run typecheck`; the rewritten helpers now use the same pipeline as the string utilities.

### DIFF
--- a/packages/vega-util/index.d.ts
+++ b/packages/vega-util/index.d.ts
@@ -31,16 +31,15 @@ export function falsy(): false;
 
 // Type Checkers
 
-export function isArray<T>(a: any | T[]): a is T[];
-export function isArray<T>(a: any | readonly T[]): a is readonly T[];
-export function isBoolean(a: any): a is boolean;
-export function isDate(a: any): a is Date;
-export function isFunction(a: any): a is Function;
-export function isIterable(a: any): boolean;
-export function isNumber(a: any): a is number;
-export function isObject(a: any): a is object;
-export function isRegExp(a: any): a is RegExp;
-export function isString(a: any): a is string;
+export { default as isArray } from './build/types/isArray.js';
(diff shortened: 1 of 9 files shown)


hydrosquall:
> at which point is it easier

If we're willing to fully replace the [rollup build with tsconfig](https://github.com/vega/vega/blob/1722e0db4356fdf72d995b655ec2b9c928a20104/packages/vega-util/package.json#L25) then I think it should be safe to cutover . I haven't investigated using typescript compiler directly on its own vs using the typescript plugin for rollup.

I saw @jheer had some concerns about cutting over Vegaa a whole over to TS some [time ago](https://github.com/vega/vega/pull/1706#issuecomment-480540202) , but if it's not an issue for `vega-util`, I think it would be a good package to pilot a direct typescript conversion .

hydrosquall:
To clarify - the js files JSdoc signatures are not picked up by
Vega-typings, only the contents of index.d.ts are. There’s no programmatic
tie between the index.d.ts written by hand  and the actual runtime JS code,
it’s only if an author remembers to run both spots.

The type params in the Jsdoc are a bridging step, the full goal is to
convert the source files to TS, and delete the handwritten index.d.ts in
favor of one that is generated from TS source files.

PR #4188, opened 2025-12-01
feat(vega-util): add signatures for all remaining type coercion to* utils

hydrosquall:
## Motivation

- Adding JSDoc param types allows us to ge type safety without having to switch to the typescript compiler (.ts files) yet
- Builds on https://github.com/vega/vega/pull/4185 / https://github.com/vega/vega/pull/4187  / https://github.com/vega/vega/issues/3971

## Changes

- Types/documents every `to*` helper in `packages/vega-util/src`
- Changes _some_ runtime behaviors + documents some surprises (i.e. toBoolean does not always return a Boolean)

## Testing

- Verified `npm run generate-types` and `npm run typecheck`; the rewritten helpers now use the same pipeline as the string utilities.

### DIFF
--- a/packages/vega-util/index.d.ts
+++ b/packages/vega-util/index.d.ts
@@ -43,10 +43,10 @@ export { default as isString } from './build/types/isString.js';
 
 // Type Coercion
 
-export function toBoolean(a: any): boolean;
-export function toDate(a: any, parser?: (_: any) => number): number;
-export function toNumber(a: any): number;
-export function toString(a: any): string;
+export { default as toBoolean } from './build/types/toBoolean.js';
+export { default as toDate } from './build/types/toDate.js';
+export { default as toNumber } from './build/types/toNumber.js';
+export { default as toString } from './build/types/toString.js';
 
 // Objects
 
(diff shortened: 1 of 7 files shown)


packages/vega-util/src/toSet.js
```diff
@@ -1,6 +1,11 @@
+/** @template T
+ * @param {readonly T[] | ArrayLike<T>} _ - Collection to convert.
+ * @returns {Record<string, true>}
+ */
 export default function(_) {
+  /** @type {Record<string, true>} */
   const s = {},
         n = _.length;
-  for (let i=0; i<n; ++i) s[_[i]] = true;
+  for (let i=0; i<n; ++i) s[_[i] + ''] = true;
```
hydrosquall, line 9: Without this line, conversion of object keys to strings is implicit rather than explicit.

REF: Js properties are strings or symbols: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Data_structures#properties

packages/vega-util/index.d.ts
```diff
@@ -43,10 +43,10 @@ export { default as isString } from './build/types/isString.js';
 
 // Type Coercion
 
-export function toBoolean(a: any): boolean;
```
hydrosquall, line 46: These old types were not correct, all these functions were capable of returning `null` and other types

Given that going through this exercise is exposing these typing inconsistencies, I'm increasingly convinced it may be worthwhile to directly convert the whole package to typescript instead of going through these intermediary types (after thinking about Dom's suggestion https://github.com/vega/vega/pull/4187#pullrequestreview-3522293158 more)

PR #4189, opened 2025-12-01
chore: use typescript source / replace rollup in `vega-util`

hydrosquall:
## Motivation

- Writing the utils directly in typescript is simpler than keeping a mix of JSDoc types + typescript
- https://github.com/vega/vega/pull/4187#pullrequestreview-3522293158

### Converted JS to TS

Moved all the `vega-util` functions with doc types from `.js` to `.ts` (already had JSDoc strings)
  - Type checking: `isArray`, `isBoolean`, `isDate`, `isFunction`, `isIterable`, `isNumber`, `isObject`, `isRegExp`, `isString`
  - Type coercion: `toBoolean`, `toDate`, `toNumber`, `toSet`, `toString`
  - String utilities: `pad`, `repeat`, `splitAccessPath`, `stringValue`, `truncate`

### Build Changes

- Added a `src/index.ts` as an entrypoint for TS, and changed the JS entrypoint to point to the `build` folder
- Updated tsconfig, .npmignore , and other build config to allow replacing Rollup with Typescript compiler
- Testing should be unaffected by the use of TS, since we already had to run the build step before running tests. However, I change the import paths to reference the build folder instead of source since tape can't reference ts directly.

## Testing

- CI Passing should be enough, no intended behavior changes

### DIFF
--- a/packages/vega-util/index.d.ts
+++ b/packages/vega-util/index.d.ts
@@ -31,7 +31,7 @@ export function falsy(): false;
 
 // Type Checkers
 
-export { default as isArray } from './build/isArray.js';
+export { isArray } from './build/isArray.js';
 export { default as isBoolean } from './build/isBoolean.js';
 export { default as isDate } from './build/isDate.js';
 export { default as isFunction } from './build/isFunction.js';
--- a/packages/vega-util/tsconfig.json
+++ b/packages/vega-util/tsconfig.json
@@ -10,5 +10,5 @@
     "declarationMap": true,
     "sourceMap": true
   },
-  "include": ["src/**/*.ts", "src/**/*.js"]
+  "include": ["src/index.ts"]
 }


packages/vega-util/package.json
```diff
@@ -21,12 +21,11 @@
     "url": "git+https://github.com/vega/vega.git"
   },
   "scripts": {
-    "prebuild": "del-cli build && npm run generate-types",
-    "build": "rollup -c rollup.config.js",
+    "prebuild": "del-cli build",
+    "build": "tsc -p tsconfig.json && rollup -c rollup.config.js",
```
domoritz: Do we actually need rollup? We could also have the export point directly at the output of tsc maybe.

hydrosquall: We could replace rollup if there's no extra post-processing steps we want rollup to do after typescript completes.

domoritz: I don't think it does anything important for most packages. We do need it to bundle the main Vega bundle but we can keep rollup for that.

packages/vega-util/src/index.ts
```diff
@@ -0,0 +1,90 @@
+export {
```
hydrosquall, line 1: This replaces the previous `src/index.js` . It's a common ts package pattern that everything in the `src` directory will be compiled.

packages/vega-util/index.d.ts
```diff
@@ -31,22 +31,22 @@ export function falsy(): false;
 
 // Type Checkers
 
-export { default as isArray } from './build/types/isArray.js';
```
hydrosquall, line 34: This file will be deleted after everything is converted to TS.

packages/vega-util/index.js
```diff
@@ -1,90 +0,0 @@
-export {
```
hydrosquall, line 1: This became `src/index.ts`

packages/vega-util/rollup.config.js
```diff
@@ -1 +0,0 @@
-export { default } from '../../rollup.config.js';
```
hydrosquall, line 1: No need for rollup, it's all compiled by `tsc`

packages/vega-util/src/isArray.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Return whether the provided value is an array.
+ */
+const isArray: <T = unknown>(value: unknown) => value is readonly T[] = Array.isArray;
+
+export default isArray;
```
domoritz: It might be cleaner not to do a default export and then we don't need to rename in the import either.

packages/vega-util/tsconfig.json
```diff
@@ -1,24 +1,14 @@
...
-    "src/toString.js",
-    "src/toDate.js",
-    "src/toSet.js",
-    "src/truncate.js",
-    "src/pad.js",
-    "src/repeat.js",
-    "src/splitAccessPath.js",
-    "src/stringValue.js"
-  ]
+  "include": ["src/**/*.ts", "src/**/*.js"]
```
domoritz: I usually prefer to only import the index.ts file and then let ts figure out the dependencies. This way I make sure not to forget an import.

packages/vega-util/tsconfig.json
```diff
@@ -1,24 +1,14 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "checkJs": true,
-    "rootDir": "src",
+    "checkJs": false,
+    "allowJs": true,
```
domoritz, line 5: What's this for now?

hydrosquall: 1. `allowJs` allows JS files to be imported. Without it, we would need the hybrid rollup/TSC combination for all files to end up in the build folder. ([src](https://www.typescriptlang.org/tsconfig/#allowJs))
2. `checkJs` determines if the imported files are typechecked ( [src](https://www.typescriptlang.org/tsconfig/#checkJs) ) . There would be a host of type errors if we imported all the JS files as is without adding type param comments first. 

This combo is a bridge step during this brief window during which there's both JS and TS in vega-util, we can remove it once the whole folder is TS (hopefully in the next PR)

packages/vega-util/src/isArray.ts
hydrosquall: I'm in favor of avoiding the default export in favor of using named exports 👍

packages/vega-util/tsconfig.json
domoritz: Ah, makes sense. It might be easier to just bite the bullet and convert everything in one pull request.

hydrosquall: Given the size of the diff (1000+ lines) I figured it would be safer to deliver incrementally (i.e. get rid of rollup in 1 step with the easy cases, then finish typing the rest of vega-util in the next PR with the tricky cases): https://github.com/vega/vega/pull/4198

I can combine the two PRs if you'd prefer to avoid the middle state, but I believe it'll be safer to go through this middle state as it's unlikely others will come in to add changes to vega-util in the middle of the transition.

domoritz: Fair. I'm happy to help get it in either way.

hydrosquall: Appreciate it - I'd recommend starting with this PR first then. The second may end up being split up further.

packages/vega-util/src/isBoolean.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Determine if the value is a boolean primitive.
+ */
+export default function isBoolean(value: unknown): value is boolean {
```
domoritz, line 4: Do we export as default or not? Seems inconsistent now.

domoritz: I'm okay either way. 

For default
* we only have one export usually per file

Against
* we would have to rename the import and can't just do a * import in the index file

hydrosquall: > Do we export as default or not? Seems inconsistent now.

I tried not to modify files that weren't part of this typescript shift, but after everything is in Typescript I'm recommending named default exports going forward.

The previous state of anonymous default exports before was that function sources were not searchable, and when errors appear in stack traces, the function would be anonymous instead of named.

hydrosquall: I don't feel strongly about naming the export and not providing it as default vs naming the export _AND_ doing it as a default export, but if it were just me I would do export named only (no default). 

In the context of this specific utility library, I think exporting as default is OK since it makes it clear what the most important function is when that exist for some packages. This is already not the case for the multi-function utils files  like

- https://github.com/vega/vega/blob/main/packages/vega-util/src/transform.js 
- https://github.com/vega/vega/blob/main/packages/vega-util/src/mergeConfig.js

so it seems OK for not every file to be handled this way.

domoritz: Happy to defer to your judgement here. Thanks for the explanation.

PR #4200, opened 2025-12-09
chore: migrate low-risk vega-util to TypeScript (Wave 1 of 3)

hydrosquall:
## Motivation

- Safely package all of vega-utils (Sequel to https://github.com/vega/vega/pull/4189 )
- Later changes are for files where converting required more than just adding the function signature or checking for empty arrays

## Changes

- Migrated 20 files
  - array, clampRange, constant, debounce, flush, hasOwnProperty, inherits, inrange, lerp, lruCache, peek, quarter, span
- In a few cases, I had to change the runtime behavior to handle invalid data (empty arrays or missing inputs) - flush, lerp, inherits - this seemed preferable to making non-null assertions. 

## Testing

- CI passing should cover most of it, check if conversions make sense

### DIFF
--- a/packages/vega-util/src/flush.ts
+++ b/packages/vega-util/src/flush.ts
@@ -20,7 +20,6 @@ export default function<T>(
       b = peek(range),
       l: number;
 
-  // TODO: Check in PRV: behavior change to handle empty b
   if (b === undefined) return center;
 
   // swap endpoints if range is reversed
--- a/packages/vega-util/src/hasOwnProperty.ts
+++ b/packages/vega-util/src/hasOwnProperty.ts
@@ -1,6 +1,3 @@
 export default function hasOwnProperty(object: object, property: PropertyKey): boolean {
-  // @ts-ignore - Object.hasOwn is available at runtime (ES2022) but not in lib target (es2021)
-  // This is safe because Object.hasOwn(obj, prop) is equivalent to Object.prototype.hasOwnProperty.call(obj, prop)
-  // and properly handles objects without a hasOwnProperty method (like Object.create(null))
   return Object.hasOwn(object, property);
 }
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -3,7 +3,7 @@
     "target": "ES2022",
     "module": "ESNext",
     "lib": [
-      "es2021",
+      "es2022",
       "dom"
     ],
     "moduleResolution": "Node",


packages/vega-util/src/flush.ts
```diff
@@ -0,0 +1,37 @@
...
+  center: T
+): T {
+  if (!threshold && threshold !== 0) return center;
+
+  const t = +threshold;
+  let a = range[0],
+      b = peek(range),
+      l: number;
+
+  // TODO: Check in PRV: behavior change to handle empty b
```
hydrosquall: This is a behavior change from before. It shouldn't happen if range contains at least 1 member, but it will happen if range is an empty array.

packages/vega-util/src/hasOwnProperty.ts
```diff
@@ -0,0 +1,6 @@
+export default function hasOwnProperty(object: object, property: PropertyKey): boolean {
+  // @ts-ignore - Object.hasOwn is available at runtime (ES2022) but not in lib target (es2021)
```
hydrosquall: We can dismiss this comment if we change our lib target to es2022

packages/vega-util/src/lerp.ts
```diff
@@ -1,8 +1,11 @@
 import peek from './peek.js';
 
-export default function(array, frac) {
+export default function(array: readonly number[], frac: number | string): number {
   const lo = array[0],
         hi = peek(array),
         f = +frac;
+
+  if (hi === undefined) return lo;
```
hydrosquall, line 8: This is also a behavior change for cases where the array is empty. LMK if the previous behavior (allow subtracting undefined - undefined = NaN ) is preferred

packages/vega-util/src/peek.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Returns the last element of an array.
```
hydrosquall, line 2: Added a docstring since the old name wasn't self-explanatory to me

packages/vega-util/src/hasOwnProperty.ts
domoritz: Let's change to es2022 then.

hydrosquall: Cool, it will mean we match vega-lite too.

https://github.com/vega/vega-lite/blob/bf84e14e3b029562aa86d961751d30fb69da5afa/tsconfig.json#L16

packages/vega-util/src/flush.ts
hydrosquall: Removing the comment assuming we're OK with this

PR #4198, opened 2025-12-06
chore: full vega-util TS conversion

hydrosquall:
> [!WARNING]
> Not a clean diff yet - WIP 

## Motivation

- Improve DX for new contributors + help with maintaining source safely ~
- https://github.com/vega/vega/issues/3971

## Changes

- Continue https://github.com/vega/vega/pull/4189

PR #4201, opened 2025-12-10
chore: vega-util wave 2 migration

hydrosquall:
## Motivation

- Migrate in waves so we can comment the more complicated functions

## Changes

- Split from https://github.com/vega/vega/pull/4198


hydrosquall:
Replaced by https://github.com/vega/vega/pull/4200 and https://github.com/vega/vega/pull/4202

PR #4202, opened 2025-12-10
chore(vega-util): migrate type-only files to TypeScript (Wave 2a)

hydrosquall:
## Motivation

- Continue migrating vega-utils to typescript, making runtime and docstring changes if needed

## Changes

- accessor.ts : Switch to using Object.assign instead of mutating a function
- accessors.ts: Type annotations only
- extent.ts: Type annotations and overloads
- extentIndex.ts: Non-null assertions (compile-time only)
- interpreter.ts: Type assertions in filter callback
- visitArray.ts: Added overload signatures

Plus one significant runtime change to protect against bad inputs

- merge.ts: **Contains a runtime change to protect against objects impersonating an array**

## Testing

- CI Passing should have us covered, though feedback on docstrings or signatures are welcome too.


packages/vega-util/src/extent.ts
```diff
@@ -0,0 +1,59 @@
...
+ * Example: extent([1, 5, 3]) returns [1, 5] with type [number | undefined, number | undefined]
+ */
+export default function extent<T>(array: ArrayLike<T>): [T | undefined, T | undefined];
+/**
+ * Overload 2: With accessor function that transforms T to U, returns extent of transformed values.
+ * Example: extent([{a:1}, {a:5}], d => d.a) returns [1, 5] with type [number | undefined, number | undefined]
+ * The returned extent has type U (the accessor's return type), not T (the array element type).
+ */
+export default function extent<T, U>(array: ArrayLike<T>, f: (value: T) => U): [U | undefined, U | undefined];
+/**
```
hydrosquall, line 16: Commented these verbosely with examples in hopes that someone who hasn't worked with generics before would find this helpful

I can remove the overload definitions if they are more confusing than helpful.

packages/vega-util/src/accessor.ts
```diff
@@ -0,0 +1,20 @@
+export interface Accessor<T = unknown> {
+  (obj: unknown, ...args: unknown[]): T;
+  fields?: readonly string[];
+  fname?: string;
+}
+
+export default function accessor<T = unknown>(fn: (obj: unknown, ...args: unknown[]) => T, fields?: readonly string[], name?: string): Accessor<T> {
+  return Object.assign(fn, {
```
hydrosquall, line 8: Mutating `fn` is a bit unconventional for typescript, using `Object.assign` made the intent a bit clearer here.

Note that Object.assign is not returning the input fn, it _is_ still mutating the input `fn`, so there shouldn't be an observable functional change as a result of switching from setting` fn.fields` directly

packages/vega-util/src/merge.ts
```diff
@@ -0,0 +1,42 @@
...
+): T[] {
+  const n0 = array0.length,
+        n1 = array1.length;
+
+  if (!n1) return array0;
+  if (!n0) return array1;
+
+  // Use array constructor to preserve type (Array or TypedArray), fallback to Array if invalid
+  const ctor = array0.constructor;
+  const merged = output || (
```
hydrosquall, line 23: The old code looked like this: 

```
const merged = output || new array0.constructor(n0 + n1);
```

Unfortunately if the input array was _not_ actually an array but an object `{ constructor: someBadFunction }`, this path could be used for bad purposes. Ensuring the array is either a valid Array or TypedArray prevents abuse from happening through this function.

packages/vega-util/src/merge.ts
```diff
@@ -0,0 +1,42 @@
+// Type guard for Array or TypedArray constructors
+function isArrayConstructor(ctor: Function): ctor is ArrayConstructor {
+  return ctor === Array || (typeof ctor === 'function' && ctor.prototype && ArrayBuffer.isView(ctor.prototype));
```
hydrosquall, line 6: Cheap way to check if the constructor is a typed array

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/isView

packages/vega-util/src/toDate.ts
```diff
@@ -2,6 +2,8 @@ import isDate from './isDate.js';
 import isNumber from './isNumber.js';
 
 const defaultParser = (_: unknown): Date | number =>
+  // Date.parse accepts string but will implicitly coerce the input value.
+  // Type assertion allows Date.parse to handle its own coercion behavior.
```
hydrosquall, line 6: I don't like to put `as` without explaining why

packages/vega-util/src/merge.ts
domoritz: Maybe add that link as a comment.

domoritz (approved):

packages/vega-util/src/merge.ts
hydrosquall: 👍  I try not to over-comment but given isView is not a frequently used API, it seems appropriate to add here

PR #4203, opened 2025-12-12
chore(vega-util): migrate files with subtle behavior changes to Typescript

hydrosquall:
## Motivation

- Continue migrating vega-utils to typescript, making runtime and docstring changes if needed

## Changes

Migrate 6 files with runtime behavior changes:

**Refactoring (no functional impact):**

- field.ts: Uses fieldName variable instead of reassigning parameter
- getter.ts: Uses result variable instead of mutating obj parameter
- key.ts: Uses processedFields/arrayFields instead of reassigning parameter
- extend.ts: Changed from arguments object to rest parameters
- logger.ts: Changed from arguments object to rest parameters
- fastmap.ts: Added isValue() guard in get() method to prevent returning NULL sentinel value

Also added a signature for `inherits.ts` to make it behave more like the [pre-migration JS behavior](https://github.com/vega/vega/blob/aa34a29e48798bf9f99c304ddd847764df0f40b4/packages/vega-util/src/inherits.js) (where undefined is a valid input / we don't fill in an empty object)

## Testing

- Leaning on existing tests

### DIFF
--- a/packages/vega-util/src/extend.ts
+++ b/packages/vega-util/src/extend.ts
@@ -1,14 +1,9 @@
 export type Writable = Record<string, unknown>;
 
-export default function extend<T extends Writable>(
-  _: T,
-  ...args: Writable[]
-): T {
-  // Use arguments object directly to avoid rest parameter array allocation overhead
-  for (let x, k, i = 1, len = arguments.length; i < len; ++i) {
-    x = arguments[i] as Writable;
-    for (k in x) {
-      (_ as Writable)[k] = x[k];
+export default function extend<T extends Writable>(_: T, ...args: Writable[]): T {
+  for (const arg of args) {
+    for (const k in arg) {
+      (_ as Writable)[k] = arg[k];
     }
   }
 
--- a/packages/vega-util/src/fastmap.ts
+++ b/packages/vega-util/src/fastmap.ts
@@ -1,4 +1,4 @@
-import hasOwnProperty from "./hasOwnProperty.js";
+import hasOwnProperty from './hasOwnProperty.js';
 
 const NULL = {};
 
@@ -11,13 +11,15 @@ export interface FastMap<T = unknown> {
   set(key: string, value: T): FastMap<T>;
   delete(key: string): FastMap<T>;
   clear(): void;
+  /** Use the test */
   test(): ((value: T) => boolean) | undefined;
-  test(fn: (value: unknown) => boolean): FastMap<T>;
+  /** Set test function */
+  test(fn: (value: T) => boolean): FastMap<T> | undefined;
   clean(): void;
 }
 
 export default function fastmap<T = unknown>(
-  input?: Record<string, T>,
+  input?: Record<string, T>
 ): FastMap<T> {
   let obj: Record<string, T | typeof NULL> = {},
     test: ((value: T) => boolean) | undefined;
@@ -54,7 +56,8 @@ export default function fastmap<T = unknown>(
       map.size = map.empty = 0;
       map.object = obj = {};
     },
-    test(fn?: (value: unknown) => boolean): any {
+    /** See the FastMap interface for the expected return type, which depends on whether a function is provided */
+    test(fn?: (value: T) => boolean): any {
       if (arguments.length) {
         test = fn;
         return map;
@@ -75,7 +78,7 @@ export default function fastmap<T = unknown>(
       map.size = size;
       map.empty = 0;
       map.object = obj = next;
-    },
+    }
   };
 
   if (input)


packages/vega-util/src/extend.js
```diff
@@ -1,7 +0,0 @@
-export default function(_) {
-  for (let x, k, i=1, len=arguments.length; i<len; ++i) {
```
hydrosquall, line 2: In modern code, `arguments` is discouraged compared to using a `rest` parameter

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments#description

packages/vega-util/src/logger.js
```diff
@@ -1,40 +0,0 @@
...
-export const Error = 1;
-export const Warn  = 2;
-export const Info  = 3;
-export const Debug = 4;
-
-export default function(_, method, handler = log) {
-  let level = _ || None;
-  return {
-    level(_) {
-      if (arguments.length) {
```
hydrosquall, line 16: In modern code, `arguments` is discouraged in favor of using a `rest` parameter

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments#description

packages/vega-util/src/extend.ts
```diff
@@ -0,0 +1,12 @@
+export type Writable = Record<string, unknown>;
+
+export default function extend<T extends Writable>(_: T, ...args: Writable[]): T {
+  for (let i = 0; i < args.length; ++i) {
+    const x = args[i];
+    for (const k in x) {
+      (_ as Writable)[k] = x[k];
```
hydrosquall: `as any` was also an option here since we don't actualy check if _ is an object, but this avoids a runtime behavior change so I figured start here

About the type, technically the return type is T + more properties, but we don't know in advance what properties the extended objects are going to add to the base.

packages/vega-util/src/fastmap.ts
```diff
@@ -0,0 +1,83 @@
...
+    return hasOwnProperty(obj, key) && obj[key] !== NULL;
+  }
+
+  const map: FastMap<T> = {
+    size: 0,
+    empty: 0,
+    object: obj,
+    has: has,
+    get(key: string): T | undefined {
+      return has(key) ? obj[key] as T : undefined;
```
hydrosquall: Noting the type cast is intentional

packages/vega-util/src/fastmap.ts
```diff
@@ -0,0 +1,83 @@
...
+      } else {
+        return test;
+      }
+    },
+    clean(): void {
+      const next: Record<string, T | typeof NULL> = {};
+      let size = 0;
+      for (const key in obj) {
+        const value = obj[key];
+        if (value !== NULL && (!test || !test(value as T))) {
```
hydrosquall, line 73: Noting the type cast is intentional

packages/vega-util/src/inherits.ts
```diff
@@ -14,5 +15,5 @@ export default function<C extends Constructor, P extends Constructor>(
     enumerable: true,
     configurable: true
   });
-  return extend(proto, members || {});
+  return extend(proto, members as Writable);
```
hydrosquall, line 18: Removed  the `{}` initializer  to be more like the JS era version

https://github.com/vega/vega/blob/aa34a29e48798bf9f99c304ddd847764df0f40b4/packages/vega-util/src/inherits.js

packages/vega-util/src/key.ts
```diff
@@ -0,0 +1,37 @@
+import accessor, { Accessor } from './accessor.js';
+import array from './array.js';
+import getter from './getter.js';
+import splitAccessPath from './splitAccessPath.js';
+
+export interface KeyOptions {
```
hydrosquall, line 6: Noting these options have the same signature as `FieldOptions` in field.ts

packages/vega-util/src/extend.ts
```diff
@@ -0,0 +1,12 @@
+export type Writable = Record<string, unknown>;
+
+export default function extend<T extends Writable>(_: T, ...args: Writable[]): T {
+  for (let i = 0; i < args.length; ++i) {
```
domoritz: Can't this be a for ... of loop?

hydrosquall: It could be, I was trying to keep the runtime implementation as close to how it was before

However rewriting this to `for arg in args` seems like a reasonable rewrite to me , I'll change it

domoritz: Sounds good. Should be cleaner.

PR #4218, opened 2026-01-15
refactor(vega-util): Migrate mergeConfig from JS to TS

hydrosquall:
## Motivation

- Improve DX for new contributors + help with maintaining source safely
- I'm converting the functions that either had less test coverage or required some runtime changes separately, so this wave is taking longer then the first two.

## Changes

- Rewrites `mergeConfig.js` to `mergeConfig.ts` , adding (human reviewed) types and comments along the way.

## Testing

- CI should continue to pass, I added new tests for the behavior that wasn't covered before.

## Notes

- Continues https://github.com/vega/vega/pull/4189

### DIFF
--- a/packages/vega-util/src/mergeConfig.ts
+++ b/packages/vega-util/src/mergeConfig.ts
@@ -1,40 +1,82 @@
 import {isArray} from './isArray.js';
 import isObject from './isObject.js';
 
+
 const isLegalKey = (key: string) => key !== '__proto__';
 
-export function mergeConfig(...configs: any[]): any {
+/** Represents any valid configuration value (primitives, arrays, or nested config objects). */
+type ConfigValue =
+  | string
+  | number
+  | boolean
+  | null
+  | undefined
+  | ConfigValue[]
+  | ConfigObject;
+
+interface ConfigObject {
+  [key: string]: ConfigValue;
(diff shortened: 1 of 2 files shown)


packages/vega-util/src/mergeConfig.ts
```diff
@@ -0,0 +1,108 @@
...
+  }
+}
+
+/** Merges named object arrays, deduplicating by name. (b takes precedence). */
+function mergeNamed(
+  a: NamedSignal[] | null | undefined,
+  b: NamedSignal[] | undefined
+): NamedSignal[] | undefined {
+  if (a == null) return b;
+  if (b == null) return a;
```
hydrosquall, line 93: This is the only new line. Before, if b was nullish, we would crash at line 59.

image

packages/vega-util/test/mergeConfig-test.js
```diff
@@ -90,3 +90,99 @@ tape('mergeConfig must not allow prototype pollution', t => {
 
   t.end();
 });
+
+tape('mergeConfig recursively merges style properties', t => {
```
hydrosquall, line 94: I added these 3 tests since we were not testing the recursive merging behavior before.

PR #4219, opened 2026-01-18
refactor(vega-util): Make `transform` functions typesafe

hydrosquall:
## Motivation

- Incrementally typing `vega-util`
- Related to https://github.com/vega/vega/issues/3971

## Changes

- Adds a runtime error when users try to operate on a domain that doesn't have at least 1 value
- Everything else should have same runtime behavior as before, I just added types

## Testing

```vasg
npm test --workspace=vega-util -- test/transform-test
```


packages/vega-util/src/transform.ts
```diff
@@ -0,0 +1,88 @@
+import error from './error.js';
+import peek from './peek.js';
+import toNumber from './toNumber.js';
+
+type NonEmptyArray<T> = [T, ...T[]];
+type TransformFn = (x: number) => number;
+type Domain = [number, number];
+type Anchor = number | null | undefined;
+
+function isNonEmpty<T>(arr: T[]): arr is NonEmptyArray<T> {
```
hydrosquall, line 10: This type guard is new vs the old version, but I felt it was worthwhile, otherwise `(x: number) => number` signature would have to handle `null/undefined` .  It could be rolled back if we don't want to provide this kind of defensive programming overhead

packages/vega-util/src/transform.ts
```diff
@@ -0,0 +1,90 @@
...
+
+const symexp = (c: number): TransformFn =>
+  x => Math.sign(x) * Math.expm1(Math.abs(x)) * c;
+
+const pow = (exponent: number): TransformFn =>
+  x => x < 0 ? -Math.pow(-x, exponent) : Math.pow(x, exponent);
+
+// Internal wrappers to adapt utility functions to TransformFn signature
+// Unlike toNumber, coerceNumber always returns a number (never null)
+const coerceNumber: TransformFn = x => toNumber(x) ?? 0;
```
hydrosquall, line 31: In the context of this file, returning 0 should be safe (everywhere that `lift()` is called)

If we don't make this change, we'll have to relax the types on all the other functions to indicate that `d0` / `d1` could sometimes be `undefined/null` instead of numbers.

packages/vega-util/src/transform.ts
```diff
@@ -0,0 +1,90 @@
...
+const identity: TransformFn = x => x;
```
hydrosquall, line 32: Technically this exists in a form that involves patching the function with a property, but this code doesn't need that extra property, and that function has a looser type that what I can provide inline here.

https://github.com/vega/vega/blob/50dc5f56f18a435529a8154d41e2db2408650c21/packages/vega-util/src/accessors.ts#L6

packages/vega-util/src/transform.ts
```diff
@@ -0,0 +1,90 @@
...
+
+function pan(domain: number[], delta: number, lift: TransformFn, ground: TransformFn): Domain {
+  if (!isNonEmpty(domain)) error('Domain array must not be empty');
```
hydrosquall, line 35: The code will give bad results if domain does not have at least 1 element, so I opted to add a runtime error here.

packages/vega-util/src/transform.ts
```diff
@@ -0,0 +1,90 @@
...
+export function panPow(domain: number[], delta: number, exponent: number): Domain {
+  return pan(domain, delta, pow(exponent), pow(1/exponent));
+}
+
+export function panSymlog(domain: number[], delta: number, constant: number): Domain {
+  return pan(domain, delta, symlog(constant), symexp(constant));
+}
+
+function zoom(domain: number[], anchor: Anchor, scale: number, lift: TransformFn, ground: TransformFn): Domain {
+  if (!isNonEmpty(domain)) error('Domain array must not be empty');
```
hydrosquall, line 64: The code will give bad results if domain does not have at least 1 element, so I opted to add a runtime error here.

PR #4220, opened 2026-01-29
chore(vega-util): Migrate compare.ts to typescript

hydrosquall:
## Motivation

- Final chapter in the convert "to TS" sequence to make contributing / maintaining Vega code easier and safer

## Changes

- Added tests for the missing data/boolean cases, + converted the last vega-util file to typescript 
  - I looked at each potential type issue, and decided to keep the existing behaviors rather than add type guards since compare is called so frequently for core sorting operations
- Added comments/renamed some variables but tried to preserve the previous style as much as possible
- After this is done, we can simplify the build files for vega-util 🎉 

## QA

- Commit 1 adds [tests](https://github.com/vega/vega/actions/runs/21465215011/job/61825695180) for unspecified behavior: 
- Commit [2](https://github.com/vega/vega/pull/4220/commits/c3d3105e9e3bfef8af56251687663377d36a7073) makes sure they still pass after migrating to TS : 
- Any feedback about unclear naming or comments are welcome - I tried to keep the runtime behavior identical to what we had before, even though that meant putting in some non-null assertions.


packages/vega-util/src/compare.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  return get.length === 0
+    ? null
+    : accessor(gen(get, ord), Object.keys(fmap));
+}
+
+export const ascending = (_u: Comparable, _v: Comparable): -1 | 0 | 1 => {
+  let u: Comparable = _u;
+  let v: Comparable = _v;
+  // Non-null assertions (!) are compile-time only and erased at runtime; JS safely compares null/undefined
```
hydrosquall: I considered adding runtime checks for this isntead, but given that the previous code behaved this way and that ascending is called many times (every pairwise row comparison), I decided to leave this alone

packages/vega-util/src/compare.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  return get.length === 0
+    ? null
+    : accessor(gen(get, ord), Object.keys(fmap));
+}
+
+export const ascending = (_u: Comparable, _v: Comparable): -1 | 0 | 1 => {
+  let u: Comparable = _u;
+  let v: Comparable = _v;
+  return (u! < v! || u == null) && v != null ? -1
```
hydrosquall, line 55: I considered adding runtime checks for this isntead, but given that the previous code behaved this way and that ascending is called many times (every pairwise row comparison), I decided to leave this alone

packages/vega-util/test/compare-test.js
```diff
@@ -99,3 +99,55 @@ tape('compare supports accessor functions', t => {
...
+  t.equal(c({x:false}, {x:false}), 0);
+  t.equal(c({x:false}, {x:null}), 1);
+  t.equal(c({x:null}, {x:false}), -1);
+  t.equal(c({x:true}, {x:null}), 1);
+  t.equal(c({x:null}, {x:true}), -1);
+  t.end();
+});
+
+tape('compare supports raw function accessors', t => {
+  const c = vega.compare([d => d.x, d => d.y], ['ascending', 'descending']);
```
hydrosquall, line 126: This suite is here confirm that this function is compatible without using the `accessor.js` utils library, otherwise it may seem odd that this suite looks very similar to the existing `compare supports accessor functions`

PR #4222, opened 2026-01-31
build(vega-util): drop JS infra from vega-util (typescript cleanup)

hydrosquall:
## Motivation

- Simplify build infra now that we're fully in TS land
- Avoid drift between a hand-coded index.d.ts and actual typescript source files

## Changes

- Fixed import for LoggerInterface that were exposed after deleting the `index.d.ts` in CI

## QA

- No behavioral changes, since these are only types, CI should be enough

### DIFF
--- a/packages/vega-util/package.json
+++ b/packages/vega-util/package.json
@@ -8,6 +8,7 @@
   ],
   "license": "BSD-3-Clause",
   "type": "module",
+  "types": "./build/index.d.ts",
   "author": {
     "name": "Vega",
     "url": "https://vega.github.io"
--- a/packages/vega-util/src/logger.ts
+++ b/packages/vega-util/src/logger.ts
@@ -13,18 +13,18 @@ export const Warn  = 2;
 export const Info  = 3;
 export const Debug = 4;
 
-export interface Logger {
-  level(_?: number): number | Logger;
+export interface LoggerInterface {
+  level(_?: number): number | LoggerInterface;
   error(...args: unknown[]): this;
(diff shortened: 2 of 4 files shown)


domoritz (approved):
🎉

packages/vega-util/src/logger.ts
```diff
@@ -13,18 +13,18 @@ export const Warn  = 2;
 export const Info  = 3;
 export const Debug = 4;
 
-export interface Logger {
-  level(_?: number): number | Logger;
+export interface LoggerInterface {
```
hydrosquall, line 16: I had to rename this since that was the type defined in the `index.d.ts`

packages/vega-util/package.json
```diff
@@ -8,6 +8,7 @@
   ],
   "license": "BSD-3-Clause",
   "type": "module",
+  "types": "./build/index.d.ts",
```
hydrosquall, line 11: Needed for backwards compatibility

packages/vega-util/index.d.ts
```diff
@@ -1,158 +0,0 @@
...
-// Strings
-export { default as pad } from './build/pad.js';
-export { default as repeat } from './build/repeat.js';
-export { default as splitAccessPath } from './build/splitAccessPath.js';
-export { default as stringValue } from './build/stringValue.js';
-export { default as truncate } from './build/truncate.js';
-
-// Logging
-
-export interface LoggerInterface {
```
hydrosquall, line 141: Old interface

PR #4230, opened 2026-02-01
chore: lint js code before commits w/ husky + lint-staged

hydrosquall:
## Motivation

- Resolve style issues locally instead of in CI for a faster feedback loop / less wasted CI minutes
- Relates to #4237
- See all projects that do this: [vega-editor](https://github.com/search?q=org%3Avega%20lint-staged&type=code) 

## Changes

- Added `husky` for pre-commit hooks
- Added `lint-staged` to format just the modified files
- Updated gitignore to also ignore local config files from agents (Claude)

## Testing

- I modified order of member inputs to something non alphabetical, and confirmed the linter fixed it

## Notes

- Realized we're not using `prettier` even though it's a dependency, but we could add it separately

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,14 +1,22 @@
-.sass-cache
-.jekyll-metadata
+# OS/IDE
 .DS_Store
 /.idea
+
+# Build
+/node_modules
+/packages/*/node_modules
+/packages/*/build
+/packages/*/LICENSE
+
+# Docs
+.sass-cache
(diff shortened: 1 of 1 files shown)


domoritz (commented):
How much does this slow down commits? If we add it here, I feel it would make sense to do it across all Vega projects so we don't have an inconsistent experience. I know that's more work but I feel that consistency would be important.

PR #4232, opened 2026-02-01
chore: remove unused vega release JS files from main

hydrosquall:
## Motivation

image

Map made by [this](https://githubnext.com/projects/repo-visualization) tool by @wattenberger

- Free up 8 MB of js files that are no longer updated or referenced, keeps the repo focused on current source code
- Improve performance when running searches in vega directory , whether in local IDE or in the github UI

## Changes

- Removed `/docs/releases` directory
- Switched to publishing the legacy js files using an old hash. This way we can keep the old files published under the vega.github.io domain without breaking the workflows of users who need the old domain.

## Notes

- Github pages example action: https://raw.githubusercontent.com/actions/starter-workflows/refs/heads/main/pages/jekyll.yml

### DIFF
--- a/.github/workflows/deploy-pages.yml
+++ b/.github/workflows/deploy-pages.yml
@@ -1,0 +1,52 @@
+name: Deploy Jekyll site to Pages
+
+on:
+  push:
+    branches: ["main"]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+concurrency:
+  group: "pages"
+  cancel-in-progress: false
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0  # Need full history to checkout from old commit
+
+      - name: Restore legacy releases from git history
+        run: |
+          # Checkout the docs/releases folder from the last commit that had it
+          git checkout 40795c2f -- docs/releases/
+
+      - name: Setup Pages
+        uses: actions/configure-pages@v5
+
+      - name: Build with Jekyll
+        uses: actions/jekyll-build-pages@v1
+        with:
+          source: ./docs
+          destination: ./_site
+
+      - name: Upload artifact
+        uses: actions/upload-pages-artifact@v3
+
+  deploy:
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    runs-on: ubuntu-latest
+    needs: build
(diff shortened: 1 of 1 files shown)


hydrosquall:
Discussed on slack - for this to work, we'd have to find a way to keep these files published even after we delete them. They can be found via github permalinks, so let's try to download those files at build time instead.

.github/workflows/deploy-pages.yml
```diff
@@ -0,0 +1,52 @@
+name: Deploy Jekyll site to Pages
+
+on:
+  push:
+    branches: ["main"]
```
hydrosquall, line 5: Loosely based on the official example:

https://raw.githubusercontent.com/actions/starter-workflows/refs/heads/main/pages/jekyll.yml

.github/workflows/deploy-pages.yml
```diff
@@ -0,0 +1,52 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0  # Need full history to checkout from old commit
+
+      - name: Restore legacy releases from git history
+        run: |
+          # Checkout the docs/releases folder from the last commit that had it
+          git checkout 40795c2f -- docs/releases/
```
domoritz, line 29: Are future releases still going to be on the docs page? I mean, most people should use a CDN but just for consistency would be nice to keep this working.

domoritz: Actually, I am probably okay with not doing this for future releases. We should just check what our release and build processes do.

hydrosquall: As far as I can tell, we don't have actions that are helping with releasing or building, so this folder was only populated manually 

https://github.com/vega/vega/tree/main/.github

I didn't see this (creating releases) discussed in Contributing MD, so I'm viewing https://github.com/vega/vega/discussions/4171 as the current source of truth for what our build/release process should involve. 

We can turn this into Git documentation similar to what I did [here](https://github.com/hydrosquall/tiingo-python/blob/master/RELEASE.rst) , possibly with an addition for when to run `npm run release` ( [here](https://github.com/vega/vega/blob/a778b4edbf19563cbf088e6931a5cf980b741360/package.json#L17) ), which I believe only applies if the runner is an admin with publish access to all the Vega ecosystem packages

domoritz: Ohh, we stopped doing this in 2.6.5. Never mind. This is good. 
Screenshot 2026-02-04 at 11 22 21

PR #4269, opened 2026-04-16
build: switch tsconfig module resolution to Bundler

joelostblom:
I had to do this for things to work locally, because of this error message:

```
tsconfig.json:3:3 - error TS5107: Option 'moduleResolution=node10' is deprecated and will stop functioning in TypeScript 7.0. Specify compilerOption '"ignoreDeprecations": "6.0"' to silence this error.
```

It seems possible to also silence it, so feel free to change to that if it is more appropriate.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -6,7 +6,7 @@
       "es2022",
       "dom"
     ],
-    "moduleResolution": "Node",
+    "moduleResolution": "Bundler",
     "noImplicitAny": true,
     "noImplicitThis": true,
     "strictNullChecks": true,


hydrosquall:
Thanks for contributing this 🙇  To confirm, which local command led to this error message appearing? Want to check if it reproduces locally / with specific dependency versions, since I don't see this error message appearing in CI (yet).

joelostblom:
Thanks for the comment @hydrosquall ! I believe it was happening when I was running `npm install` and digging a bit deeper into this, I think it is due to [breaking changes in Typescript 6.0](https://devblogs.microsoft.com/typescript/announcing-typescript-6-0-rc/#deprecated:---moduleresolution-node-(a.k.a.---moduleresolution-node10)). However, TS6.0 is still just in release candidate status, so I likely had a beta version of that installed when I encountered the error. I just tried with TS5.9.3 and the current `.tsconfig.json` is working fine.

An alternative to merging this PR individually, could be to review the other breaking changes and see if there is anything else we need to change to support TS6 (and in the meantime maybe specify a maximum TS version?). Although for me previously, the only change I did was the one in this PR and it seems like it made things work with TS6, so maybe there aren't other notable changes needed in Vega to be compatible with TS6?

I'm happy with either direction and don't feel like this is urgent since TS6 is still only a release candidate, but I also didn't run into any issues so far with using the change in this PR together with TS5.9.3 so maybe it is a harmless change. The caveat to all of this is that my general TS knowledge is quite limited.

hydrosquall:
Thanks for having a look @joelostblom 🙇 . I haven't tried using TS6 in the Vega ecosystem yet.

 If we make this change, let's bundle it with a TS6 upgrade when it's stable, since our capacity to investigate surprise regressions is low (especially given that the majority of the `vega` repo is not in typescript yet ). . Would it be OK to close this PR for now, and come back to it as part of a bigger TS6 upgrade?

joelostblom:
Totally ok!

Issue #3959, opened 2024-08-09
Marks do not show when `domain()` is used with an ordinal scale range


joelostblom:
[...]

PR #4270, opened 2026-04-16
fix: provide ordinal scheme fallback for empty initial domains

joelostblom:
This addresses https://github.com/vega/vega/issues/3959, details about the motivation can be found there.


packages/vega-encode/src/Scale.js
```diff
@@ -325,6 +325,7 @@ function configureScheme(type, _, count) {
   // determine size for potential discrete range
   count = (type === Threshold) ? count + 1
     : (type === BinOrdinal) ? count - 1
+    : (type === Ordinal) ? (+_.schemeCount || count || DEFAULT_COUNT)
```
hydrosquall, line 328: In cases where an explicit `count` is provided, should that case take precedence over `+_.schemeCount` ?

packages/vega-encode/test/scale-test.js
```diff
@@ -218,6 +218,16 @@ tape('Scale respects range color schemes', t => {
   t.equal(v[1], u(2/4));
   t.equal(v[2], u(3/4));
 
+  // fallback ordinal scheme count for initially empty implicit domain
+  s = scale({type: 'ordinal', scheme: 'blues', domain: [], domainImplicit: true});
+  v = s.range();
+  t.ok(v.length > 0);
```
hydrosquall: Is the idea that the length of `v` should be number of colors in the `blues` color scheme? When would we want the length to be the size of the scheme vs the DEFAULT_COUNT (5)?

hydrosquall (commented):
This looks like a promising direction, please bear with my questions as this is a new area of code for me too.  Thanks for pushing on this promising direction of making interactive legends work 🙇

packages/vega-encode/src/Scale.js
joelostblom: Great question, I agree that it depends on what we want to take precedence in a spec. To way I understood it, is that `+_.schemeCount` corresponds closer to the case of an "explicit count" being given, e.g. in the following example I believe that `"count": 3` inside `range` is passed via `+_.schemeCount` whereas `count` would be holding the number of values in the domain array `["a", "b"]`:

```json
{
  "$schema": "https://vega.github.io/schema/vega/v6.json",
  "width": 220,
  "height": 100,
  "padding": 5,
  "data": [
    {
      "name": "table",
      "values": [
        {"x": 0, "y": 0, "cat": "a"},
        {"x": 1, "y": 0, "cat": "b"},
        {"x": 0, "y": 1, "cat": "a"},
        {"x": 1, "y": 1, "cat": "b"}
      ]
    }
  ],
  "scales": [
    {
      "name": "x",
      "type": "band",
      "domain": [0, 1],
      "range": [0, 100],
      "padding": 0.1
    },
    {
      "name": "y",
      "type": "band",
      "domain": [0, 1],
      "range": [0, 80],
      "padding": 0.2
    },
    {
      "name": "withoutRangeCount",
      "type": "ordinal",
      "domain": ["a", "b"],
      "range": {"scheme": "blues"}
    },
    {
      "name": "withRangeCount",
      "type": "ordinal",
      "domain": ["a", "b"],
      "range": {"scheme": "blues", "count": 3}
    }
  ],
  "marks": [
    {
      "type": "rect",
      "from": {"data": "table"},
      "encode": {
        "enter": {
          "width": {"scale": "x", "band": 1},
          "height": {"scale": "y", "band": 1},
          "fill": [
            {
              "test": "datum.y === 0",
              "scale": "withoutRangeCount",
              "field": "cat"
            },
            {
              "scale": "withRangeCount",
              "field": "cat"
            }
          ]
        }
      }
    },
    {
      "type": "text",
      "encode": {
        "enter": {
          "x": {"value": 115},
          "y": {"value": 25},
          "text": {"value": "no count: 2 colors"},
          "baseline": {"value": "middle"},
          "fontSize": {"value": 12}
        }
      }
    },
    {
      "type": "text",
      "encode": {
        "enter": {
          "x": {"value": 115},
          "y": {"value": 65},
          "text": {"value": "count: 3 colors"},
          "baseline": {"value": "middle"},
          "fontSize": {"value": 12}
        }
      }
    }
  ]
}
```

Image

Since the range count gets precedence over the number of values in the domain array, the second color is lighter in the second row since the darkest of the three colors is not shown as there are only two values in the array.

If we put `count` first in the conditional, then both rows would have the same number of colors (two):

Image

packages/vega-encode/test/scale-test.js
joelostblom: I think that Vega stores `blues` as a continuous/interpolating scheme, so there is no discrete "scheme length" to preserve. For ordinal scales with a non-empty domain, we quantize to the domain cardinality. The DEFAULT_COUNT fallback only applies when the initial domain is empty, because otherwise the count would be 0 and the generated range would be empty. I tightened the test to assert the fallback length is 5 and added a non-empty domain case.

joelostblom (commented):
Thanks for the comments @hydrosquall ! See my responses inline and let me know what you think

packages/vega-encode/src/Scale.js
joelostblom: @hydrosquall Just a ping on this to see if you think this explanation is sufficient to merge in the PR .

domoritz: Makes sense to me. I'll merge.

domoritz (approved):

packages/vega-encode/src/Scale.js
joelostblom: Thank you!

Issue #4278, opened 2026-05-22
bug(types): mergeConfig’s `signals` type is incompatible with vega-typings


flying-sheep:
https://github.com/vega/vega/blob/f436c6727018c7bfb41dce846b5746f7a494f7cd/packages/vega-util/src/mergeConfig.ts#L21-L25

Attempting the following errors:

```ts
import { mergeConfig } from 'vega-util'
import type { Config } from 'vega-typings'

const cfg: Config = …

mergeConfig(cfg, {})
```

```
The argument of type “Config” can’t be assigned to parameter of type “Partial<VegaConfig>” with “exactOptionalPropertyTypes: true”. […].
  The types of the property "signals" are incompatible.
    The type "(InitSignal | NewSignal)[]" can’t be assigned to type "NamedSignal[]".
      The type "InitSignal | NewSignal" can’t be assigned to type "NamedSignal".
        The type "InitSignal" can’t be assigned to type "NamedSignal".
          The index signature for type "string" is missing in type "InitSignal".
```

shouldn’t `mergeConfig` have type `(...configs: Partial<Config>[]) => Config` (with `Config` from `vega-typings`)?

domoritz:
https://github.com/vega/vega/blob/f436c6727018c7bfb41dce846b5746f7a494f7cd/packages/vega-typings/types/spec/config.d.ts#L45 is a lot morę specific. If you have a more generic fix, let's do that instead.

flying-sheep:
The actual merged type is this, and some of these properties aren’t `ConfigValue`.

```ts
export interface NewOrInitSignal {
  name: string;
  init: Expr;
  description?: string;
  on?: OnEvent[];
  value?: SignalValue;
  react?: boolean;
  update?: Expr;
  bind?: Binding;
}
```

Also I think even if you fixed that, your `NamedSignal` definition would still be *more* specific than the actual one, as it says that all possible keys on the object have to conform to the index signature whereas a `InitSignal | NewSignal` can have any additional properties as long as the above types match.

hydrosquall:
Thanks for raising this @flying-sheep , I didn't import the existing type as it would have created a circular dependency.

Will have a look at how to fix this 👍

hydrosquall:
@flying-sheep I've prepped a draft with unit tests, would this work for you?

https://github.com/vega/vega/pull/4279

hydrosquall:
Should be fixed in the latest release: https://github.com/vega/vega/pull/4312

PR #4279, opened 2026-05-24
fix(vega-typings, vega-util): tighten NumericValueRef and fix mergeConfig signature

hydrosquall:
## Motivation

- `vega-util` and `vega-typings` types for the Config object are currently incompatible
- Draft in #4278

### Changes

Two related type correctness fixes:

1. vega-typings: Replace '| {}' in NumericValueRef with '| { offset: number | NumericValueRef }'

   The '| {}' escape hatch made NumericValueRef accept any object (including '{}',
   '{ mult: 2 }', etc.) which the JSON schema would reject. The schema's anyOf has
   a 'required: ["offset"]' branch for standalone offset-only refs; this replaces
   the overly broad '{}' with the precise shape that branch represents.

   Valid patterns still accepted: { field }, { signal }, { value }, { scale, band },
   { offset: 5 }, { field, mult, offset } etc.
   Now correctly rejected: {}, { mult: 2 }, { round: true } (modifier-only, no base)

2. vega-util: Make mergeConfig generic and decouple VegaConfig from ConfigObject

   The previous VegaConfig extended ConfigObject ([k: string]: ConfigValue), which
   forced callers to provide index-signature-bearing types. vega-typings Config does
   not have a top-level index signature, so mergeConfig(configA, configB) failed TS.

   The root issues (per @flying-sheep's feedback in https://github.com/vega/vega/issues/4278
   - NamedSignal had [k: string]: ConfigValue, but NewSignal/InitSignal properties like 'on: OnEvent[]' and 'bind: Binding' contain types outside ConfigValue (e.g. EventTarget, Stream), so the constraint was both wrong and unsatisfiable
   - VegaConfig extending ConfigObject propagated the index signature requirement to callers

## Testing

- CI passing should be enough, this only modifies types.

### DIFF
--- a/packages/vega-typings/tests/spec/invalid/encode.ts
+++ b/packages/vega-typings/tests/spec/invalid/encode.ts
@@ -1,0 +1,17 @@
+import { NumericValueRef } from 'vega';
+
+let numRef: NumericValueRef;
+
+// An empty object is not a valid numeric value ref — there is no base value
+// @ts-expect-error
+numRef = {};
+
+// Modifier-only objects are not valid — a base value (field/signal/value/offset/etc.) is required
+// @ts-expect-error
+numRef = { mult: 2 };
+
+// @ts-expect-error
+numRef = { exponent: 2 };
+
+// @ts-expect-error
+numRef = { round: true };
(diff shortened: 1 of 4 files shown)


flying-sheep:
Looks good to me. Is `vega-lite`’s Config the same as `vega`’s? Or should they be both tested?

hydrosquall:
Good question, per the [docs](https://vega.github.io/vega-lite/docs/config.html)

> (This config object in Vega-Lite is a superset of [Vega config](https://vega.github.io/vega/docs/config/).)

hydrosquall:
This is made unnecessary by https://github.com/vega/vega/pull/4312 (thanks @domoritz )

Issue #1372, opened 2018-07-19
Add support for `pattern` mark


GCheung55:
MDN Reference: https://developer.mozilla.org/en-US/docs/Web/SVG/Element/pattern

An alternative to having a pattern would be to use CSS background rules, but that wouldn't work on Canvas.

domoritz:
Just wondering, can you use an SVG image with a reference?

GCheung55:
@domoritz Could you clarify? I'm not sure what you mean.

domoritz:
In

```svg
<svg viewBox="0 0 230 100" xmlns="http://www.w3.org/2000/svg">
  <defs>
    <pattern id="star" viewBox="0,0,10,10" width="10%" height="10%">
      <polygon points="0,0 2,5 0,10 5,8 10,10 8,5 10,0 5,2"/>
    </pattern>
  </defs>

  <circle cx="50"  cy="50" r="50" fill="url(#star)"/>
  <circle cx="180" cy="50" r="40" fill="none" stroke-width="20" stroke="url(#star)"/>
</svg>
```

`fill="url(#star)"` references the pattern. I wonder whether you could use this in Vega and define the pattern outside.

GCheung55:
Looks like that is possible but I'm unsure if/how this would work for Canvas rendering.

```html
<svg viewBox="0 0 0 0" width="0" height="0" xmlns="http://www.w3.org/2000/svg">
  <defs>
    <pattern id="star" viewBox="0,0,10,10" width="10%" height="10%">
      <polygon points="0,0 2,5 0,10 5,8 10,10 8,5 10,0 5,2"/>
    </pattern>
  </defs>
</svg>

<svg viewBox="0 0 230 100" xmlns="http://www.w3.org/2000/svg">
  <circle cx="50"  cy="50" r="50" fill="url(#star)"/>
  <circle cx="180" cy="50" r="40" fill="none" stroke-width="20" stroke="url(#star)"/>
</svg>
```

domoritz:
It won't work unfortunately.

liuq:
Adding the possibility to specify filling patterns would be great!

jheer:
I'm open to PRs on vega-scenegraph for adding pattern fill/stroke support (as opposed to a new mark type). However, I'm not sure if there are equivalent methods shared between SVG and Canvas for achieving this...

kanitw:
One possibility is to make image mark support repeat/repeat-x/repeat-y ? 

[...]

PR #4290, opened 2026-07-13
feat: Pattern Fills (#1372)

dm-p:
> Attribution Disclaimer: I have some huge knowledge gaps when it comes to canvas vs. SVG. There were many nits in the implementation I had shown in #1372 that I was hoping to tidy up and submit at the time, but due to life getting in the way, I wasn't able to dedicate enough time and resources to finishing them as I would have liked. As such, my original branch had become stale, but given the recent re-enablement of Claude Fable, I felt this was a good time to try and smooth out the burrs in the original approach. I used the opportunity to review and rewrite it. The submitted work centers on this original branch but has been replanned and redeveloped using Claude Code rather than by hand. As such, I have reviewed and UAT'd this extensively before submitting for review.

## Overview

Adds pattern/texture fills and strokes to Vega marks, rendering consistently in both the SVG and Canvas renderers, per the discussions on #1372:

- Patterns are **object-valued fills/strokes in the gradient family** ("proposal B"), usable anywhere a color is: mark `fill`/`stroke` encode values and **scale range literals**.
- A **named-pattern registry** (new `vega-pattern` package, **authored in TypeScript**, as some packages seem to be undergoing migration, e.g., vega-util).
- This exports 13 built-ins reproducing the common patternfills/textures vocabulary (`diagonal-stripe`, `horizontal-stripe`, `vertical-stripe`, `crosshatch`, `circles`, `crosses`, `caps`, `grid`, `squares`, `nylon`, `waves`, `woven`, `houndstooth`), extensible at runtime via `vega.pattern(name, def)` — the same shape as `scheme()`.
- **`repeat: true | false | "x" | "y"`** as agreed.
- A **`pattern()` expression function** (the `gradient()` analog) builds pattern values in signal expressions - the idiom for composing a texture scale with a separate color scale on one mark: `pattern(scale('tex', datum.type), {foreground: scale('color', datum.group)})`.
- Two built-in schemes for ordinal ranges: **`"patterns"`** (texture-only, for redundant color+pattern encoding) and **`"monochrome"`** (solid grayscale values interleaved with textures - the classic value-plus-texture palette for print/grayscale charts).
- **Legend swatches render patterns** (contain-fit, both renderers).
- Works headless: `vg2png` / `vg2svg` / `View.toCanvas()` / `View.toSVG()` all produce complete pattern output.

## Spec Syntax

Four variants, discriminated by key:

```jsonc
// named (registry): foreground/background recolor; geometry + tileSize locked at registration
"fill": {
  "value": {
    "pattern": {
      "name": "diagonal-stripe",
      "foreground": "#4c78a8"
    }
  }
}

// inline symbol: SVG path in tile coords; strokeWidth > 0 declares stroked geometry
"fill": {
  "value": {
    "pattern": {
      "shape": "M2,2 L8,8 M8,2 L2,8",
      "tileSize": 10,
      "strokeWidth": 1.5,
      "foreground": "teal"
    }
  }
}

// rule generator: angled line sets expanded analytically to a tile path
// (spacing defaults per angle to a seamlessly tessellating value)
"fill": {
  "value": {
    "pattern": {
      "rule": {
        "angle": [45, 135]
      },
      "foreground": "purple"
    }
  }
}

// image tile: aspect-preserving; tileSize = tile width; partial repeat is mark-anchored by default
"fill": {
  "value": {
    "pattern": {
      "url": "data:image/png;...",
      "tileSize": 16,
      "repeat": "x"
    }
  }
}
```

Common properties: 

- `foreground`
- `background`
- `strokeWidth`
- `origin: "view" | "mark"` (defaults to `view` for fully repeating patterns and `mark` for partial/`false` repeat; a single strip or tile only makes sense relative to the mark it fills)
- `scale`
- `shapeRendering`
 
`shape` also accepts an inline `{type: "lines", angle, spacing, bleed, phase}` generator. Scale ranges accept wrappers directly, mixed with plain colors:

```jsonc
"scales": [{"name": "c", "type": "ordinal", "domain": {...},
  "range": [{"pattern": {"name": "crosshatch", "foreground": "#4c78a8"}}, "#f58518"]}]
// or the one-liners: "range": {"scheme": "patterns"}   // texture-only
//                    "range": {"scheme": "monochrome"} // greys + textures, print-friendly
```

Signal expressions can build or rewrite pattern values using the `pattern()` function: `pattern(spec[, overrides])`, which takes a name, an existing pattern value, or a definition object and returns the canonical wrapper.

## Design Decisions that May Warrant Further Discussion

1. **Text and rule marks are always mark-anchored, by construction.**

    - These are the only marks whose SVG emission uses per-item element transforms, which makes SVG patterns on them inherently mark-anchored.
    - Rather than fighting that (measurement/compensation cannot converge across font engines), Canvas matches the frame: when (and only when) a text/rule paint is a pattern, drawing wraps in `context.translate(x, y)` (the code path rotated text already takes). 
    - Non-pattern rendering is byte-identical to before (verified pixel-for-pixel against pre-change code). `origin` is ignored for text/rule.

2. **Renderer state never touches user spec objects.**

    - Normalization is pure; all renderer state lives in `WeakMap`s.
    - SVG defs are content-keyed (shared across items with equal canonical specs)
    - Legend swatch defs (`objectBoundingBox`, contain-fit) are keyed separately from mark defs (`userSpaceOnUse`) so they never collide.

3. **Canvas positioning is raster-baked (no `DOMMatrix`).**

    - node-canvas has no global `DOMMatrix` and its `repeat-x`/`repeat-y` `createPattern` modes are broken (fills off the origin strip paint nothing).
    - Positioning is therefore baked into the tile raster (canonical phase-0 cell rasterized once per wrapper, phased layouts composed via wrapped `drawImage` - translation-equivalent, probe-verified at maxDelta 0).
    - Browser and Node output are identical by construction.

4. **Named patterns lock geometry AND `tileSize` at registration** (`scale` is the resize control)

    - This is so stripe continuity can't be broken by overrides.
    - `foreground` recolors declared paints but preserves `transparent`/`none` fills (hollow patterns stay hollow).

5. **Rule/lines spacing defaults are tessellation-aware.**

    - A line set tiles seamlessly only when the tile translations projected onto the line normal are integer multiples of the spacing. 
    - The default spacing is therefore resolved per angle (`tileSize / 2` axis-aligned, `tileSize / √2` for the 45-degree family).
    - Explicit spacing is honored verbatim, with the constraint documented.

6. **Continuous scales reject pattern schemes with a comprehensible error.**

    -  _"Scale type sequential-linear does not support interpolating color schemes"_ instead of the silent black interpolation vega-encode previously produced for any object scheme.
    - The guard inspects every scheme entry, so mixed schemes (e.g., `"monochrome"`, which leads with a color string) cannot slip past.

7. **The `origin` default follows `repeat`.**

    - A fully repeating pattern defaults to `origin: "view"` (a view-wide field, continuous across marks)
    - Partial (`"x"`/`"y"`) or non-repeating patterns default to `"mark"` - a single strip or tile anchored at the view origin would usually miss the mark entirely.
    - An explicit `origin` always wins.

8. **The `"monochrome"` scheme is value-plus-texture by design.**

    - I wanted to provide a scheme that had an intuitive interpolation of solids and patterns, not just patterns-only.
    - Four solid greys are the ceiling for reliably distinguishable flat fills on white.
    - Pure white is excluded (invisible without a mark stroke)
    - Solids never sit adjacent, and neighboring textures come from different geometry families.
    - The composition and order are pinned by contract tests.

9. **The registry patterns (and `patterns` scheme order).**

    - These were everything that made sense to include at this time, but I'm happy to make changes as needed to make these as ideal as possible for accessible designs.
    - The extensibility framework (and registry approach) was designed to try and make future amendments as easy as possible, but still allow devs to register their own patterns with as little friction as possible.
    - I didn't want to fill the registry with schemes that hadn't been vetted (or gray variants) without further discussion first, just in case we needed to course correct any decisions.

## Testing

- Testing coverage in vega-pattern, plus extension points to vega-parser, vega-encode, vega-schema, vega-scenegraph, vega-functions and vega end-to-end tests (see gallery examples below for these).
- Due to the initial nits with SVG/Canvas synergy, there are also cross-renderer parity tests in vega-scenegraph to ensure that these match.

#### `pattern-marks.vg.json`

All four variants across rect/symbol/arc/**text**/path/area/trail, pattern **stroke**, mixed pattern/color range, interactive signals.

image

#### `pattern-scale-redundant.vg.json`

**Flagship accessibility chart**: color + pattern redundant encoding; legibility in grayscale is machine-verified in tests.

image

#### `pattern-legend.vg.json`

Mixed pattern/color ordinal range with correct legend swatches.

image

#### `pattern-textures-showcase.vg.json`

All 13 built-ins, data-driven via `{pattern: {name: datum.name}}` expressions.

image

#### `pattern-image-legend.vg.json`

Image tiles (data URIs) with bound repeat/tileSize signals + legend.

image

#### `pattern-registry.vg.json` 

The four wrapper variants side by side; runtime registration covered in `pattern-registry-test.js`.

image

## Documentation

Doc-site sources are included:

- A **Pattern** parameter-type reference (`docs/types/#Pattern`, structured like the existing Gradient type - all four variants, the lines generator, and common properties).
- **Pattern fills in scale ranges** (`docs/scales/#range-patterns`).
- The built-in **pattern schemes** (`docs/schemes/#patterns` - both `"patterns"` and `"monochrome"`, each with a live rendered swatch example), a legend swatch note, and updated `fill`/`stroke` mark property rows.
- The **`vega.pattern`** registry accessor in the Extensibility API (`docs/api/extensibility/#patterns`). The **`pattern()` expression function** is documented alongside `gradient()` in the expression language reference (`docs/expressions/#pattern`).

**I don't know the target version** (assuming this gets approved 😅), so any `includes` clauses have been left as `"TBD"`. I can fix these up if you can advise a suitable named version.

Also, I've included two examples:

#### **Monochrome Bar Chart**

Bar Charts category; the "monochrome" scheme with outlined bars and pattern legend swatches.
   
image

#### **Pattern Area Chart**

Line & Area Charts category; a diagonal-stripe shaded area beneath a solid black border line, with the pattern bound to a signal selector.
    
image

## Known Limitations (or Possible Follow-Ups)

- `position` property for non-repeating tiles (proposed on #1372) is not implemented. If this is critical to the feature, I can spend some time looking into it.

- Rule/lines patterns at angles whose tangent is irrational (e.g., 30 degrees) cannot tessellate in a square tile mathematically. The defaults cover the axis-aligned and 45-degree families. Other angles show seams at tile boundaries.

- Free rotation (`angle`) is not a common property for patterns, but line angles live in the `rule`/`lines` generators.

- Spec scheme *combining* (pattern scheme + color scheme element-wise) would need a `scheme()` expression function  - as everything in the PR should cover known cases for now, I didn't want to conflate.

- Stroke-channel legends, gradient-legend misuse with pattern ranges, and custom legend `symbolType` swatches are untested paths.

- SVG def ids (`pattern_N`) are globally sequential - two views inlined into one page could collide (same as existing gradient ids).

- Dynamic scale-range updates re-emit defs each render pass. Def GC across re-renders relies on the existing defs index cleanup (gradient-identical).

- Fractional `scale` products:
    - Canvas rounds the tile cell.
    - SVG keeps the exact value (~sub-pixel phase drift per tile; exact at integer products).

- The `Pattern*` spec types exist in two places by design for now:
    - `vega-typings/types/spec/encode.d.ts` (feeds `ts-json-schema-generator`, schema source of truth)
    - `vega-pattern/src/types.ts` (runtime source of truth) - field-identical, each cross-referenced with a keep-in-sync note.
    
    Consolidating onto the vega-pattern (the `LoggerInterface` precedent) would be a follow-up gated on verification of schema-generator output.

- `vega-typings` gained runtime accessor + spec types for patterns.
    - Vega-Lite wiring is follow-up work (as discussed on the issue).
    - I already have a proposal for the Vega-Lite syntax, but didn't want to write up unless material changes were needed. If you can advise where this is best placed (an issue or draft PR in the V-L repo?), I can document this.


hydrosquall:
Hey @dm-p , wanted to acknowledge that we've received this and look forward to figuring out how to bring this in safely / maintainably . I figure this [comment](https://github.com/vega/vega/issues/1372#issuecomment-3283453195) covers the main motivation, but if any new graphs have emerged since then, I'd be interested to know about them.

 If merged, would this this be the first time a feature explored in Deneb made its way back into Vega core?

dm-p:
Hi @hydrosquall, and thanks for the acknowledgment :) 

Yes, that linked comment is the main motivation. I think that if vega-themes counts, then this may be the second contribution; I contributed a Power BI theme to that package a long time ago. I would believe that the pattern fill functionality I patched into Deneb would be the only real divergence from core, aside from the Power BI-specific things, which are all added via the conventional extensibility APIs, and the feature in Deneb was always a hacky way of making it work (plus no canvas support or real extensibility). So there's probably not much to offer back in terms of what Deneb does differently to the core, except for some workarounds I'm patching for vega-embed.

I've long wanted to be able to spend more of my time learning and contributing more to the core, but Deneb has been a large demand on my time, but hopefully this will ramp down very soon. I'm finishing up a major version, and (hopefully) this should be the last of the overall vision I had for it when I started it ~6 years ago, except for a few more additions, and assuming MS doesn't do anything majorly new in this space that would necessitate a good look at what we could do to integrate it.

As such, I'm happy to be available (as much as OSS contribution time allows) to maintain this feature, including adding proper Vega-Lite support if you see it as a candidate for the core. Likewise, happy to tackle any remediations you see here as necessary to land it successfully.

hydrosquall:
Hey @dm-p , thanks for the thoughtful reply and ongoing interest in supporting the ecosystem 🙇 . The themes contribution definitely counts as a contribution to the ecosystem, but I think a debut contribution to the `vega/vega` package is a unique and excellent milestone to make note of!

Regarding feedback for this specific PR - since this would modify the vega spec (as opposed to being a purely Javascript based contribution), we may have to handle this a bit more slowly but I think we can still support it. Would it be a useful first pass to add the patterns in 1 pr, and hold off on adding the extensibility API, or is the extensibility API needed for this to be useful in Deneb/PowerBI?

dm-p:
Thanks @hydrosquall - it's genuinely exciting to be able to potentially contribute to a tool that has made a huge difference to how I can do my work, and that it's so accessible for others to do the same :)

To provide a quick turnaround on your questions:

### Holding off on the extensibility API 

- The extensibility is a by-product of the implementation: I designed the solution to use its own APIs to construct the OOTB registry, so exposing it publicly was almost free, and as such I thought it made sense to include. I understand why doing it now might be too early, though, and it would be more sensible to introduce it later.

- The registry mechanism has to stay (it's how the built-ins exist internally), but the public surface is just three things:
    - The `pattern` re-export on the `vega` namespace
    - The accessor declaration in vega-typings
    - The extensibility docs section.

-  I can trim all three in this PR, and we can expose it in a follow-up if/when there's demand.
- One clarification so we're scoping the same thing: this is only the JS registry accessor (`vega.pattern(name, def)`). The `pattern()` expression function is spec-side (it's how JSON-only authors compose a texture scale with a color scale) and would need to stay.

### Requirement for Deneb

No, it's not needed.

- Deneb is JSON-only, similar to vega-editor in how folks author things (we can't allow JS execution by users due to MS restrictions), so Deneb authors could never call the registry API anyway. They're fully served by the spec-side surface: the built-in names, the two schemes, and the inline pattern variants. For custom patterns, they'd be able to use the inline symbol (`shape`), `rule` generator, or image (`url`) methods.
- The registry API would only ever benefit JS-embedding hosts - e.g., Deneb itself (or similar tooling) registering curated patterns on behalf of users - which is a nice-to-have, not a prerequisite. I don't have any plans at the moment to diverge for Deneb vs. Vega.
- For context on motivation: Deneb's implementation has always been a hack around Vega not supporting this natively. As patterns/textures are a big part of visual encodings, I've wanted to see Vega support them centrally rather than have Deneb be exclusive here. It'd be nice to eventually deprecate the Deneb implementation (it'll stick around for backward compatibility), and have Deneb be less bespoke in its approach going forward.

dm-p:
Hey @hydrosquall - just checking in to see if any of the above were definite actions on my end to sort out? I know this is a substantive PR, and I want to make sure I haven't missed anything I should have done to make things better/easier for the team before it can be looked at.

PR #4309, opened 2026-07-20
feat: add opt-in cooperative scheduling for evaluation and rendering

ianduvall:
Adds an opt-in `scheduling` option to the View constructor (`boolean` or `{signal?: AbortSignal}`) for cooperative scheduling of dataflow evaluation and canvas rendering:

- Long-running `runAsync` calls yield to the browser event loop every 10ms, using `scheduler.yield()` where available with a `setTimeout` fallback, so large datasets don't block the main thread. Yields inherit the caller's task priority; Vega never posts tasks with an explicit priority.
- Evaluation yields between transforms, and also within the per-item loops of the data join (`DataJoin`), encoding (`Encode`), and bounds (`Bound`) transforms, so a single large mark does not hold the main thread for the whole of its own work. Line and area marks are still bounded as a unit, so one such mark with very many points remains a single uninterrupted task.
- Canvas rendering also yields, drawing to an offscreen buffer that is displayed once rendering completes, so no partial frames are shown.
- An aborted `AbortSignal` rejects the in-flight `runAsync` promise with the signal's reason; the view should then be finalized. `finalize()` also cancels in-flight scheduled work.
- While scheduling is enabled, synchronous `run()` calls queue behind in-flight evaluation and rendering instead of mutating state mid-render.
- Runs issued from `runAfter` callbacks (for example the second layout pass of `autosize: "fit"`) are awaited before `runAsync` resolves, matching synchronous evaluation, where those runs complete inline before the promise settles.
- Fixes `Transform` so an async evaluate that resolves to `StopPropagation` is not stored as the operator's pulse.
- Default behavior (`scheduling: false`) is unchanged: evaluation and rendering run synchronously exactly as before.


hydrosquall:
Hi @ianduvall , thanks for reaching out with this feature idea. Do you have an example large spec that we could use to test the effect of this change for the end user, or context about the situation that motivated opening

The idea looks interesting, I am wary of adding a significant volume of code to  `vega/vega` for this optional feature that would ripple across many packages . if we adopt this change, I'd like to make sure that the benefits are clearly worth it from an end-user's perspective before going into the code.

ianduvall:
Hi @hydrosquall, thanks for taking the time to review.

> Do you have an example large spec that we could use to test the effect of this change for the end user, or context about the situation that motivated opening

The motivation is to prevent complex visualizations from blocking the main thread for large periods of time. This problem is not solvable in user land because Vega's synchronous architecture runs everything in a single task.

I added [scheduling.html](https://github.com/vega/vega/pull/4309/changes#diff-61260f1a1db4899f675aebe6e9a23c4940f46c44cb050f74dbf753941a481f68) with three visualizations that you can use to profile locally to see the difference between scheduling and without. 

Happy to explain more if anything is unclear.

ianduvall:
@hydrosquall I split this PR into a stacked PR with 4 commits. I'm not able to point it at the upstream repo because it's a fork but happy to land them one at a time if that's easier.

ianduvall:
Going to close this for now

PR #4305, opened 2026-07-18
test: compare scenegraphs against goldens by default

domoritz:
## Problem

`packages/vega/test/scene-test.js` had `const GENERATE_SCENES = true;` committed on main. In that mode the scene test **writes** the golden fixtures instead of comparing against them, so every run (including CI) trivially passed and the scenegraph suite guarded nothing.

This hid two issues:

- **Stale crossfilter goldens**: `crossfilter.json` and `crossfilter-multi.json` had drifted from current output (sub-pixel numeric differences, stale since ~v6.0.0).
- **Missing golden**: `bar-time-auto.json` was never committed for the spec added in #4299, so comparison mode failed with ENOENT.

## Changes

- Default `GENERATE_SCENES` to `false`; regeneration is now opt-in via an environment variable:
  ```
  GENERATE_SCENES=1 npx tape test/scene-test.js
  ```
  (run with `TZ=America/Los_Angeles` to match `npm test`, since some scenes are timezone-sensitive)
- Refresh the stale `crossfilter.json` and `crossfilter-multi.json` goldens.
- Add the missing `bar-time-auto.json` golden (generated under `TZ=America/Los_Angeles`, matching the package test script).

## Verification

- Comparison mode before fixture refresh: 115 tests, 112 pass, 3 fail (bar-time-auto, crossfilter, crossfilter-multi).
- Comparison mode after: 115/115 scenes pass.
- Full `vega` package suite (`npm test`): 309/309 pass.

## Ordering note

Open PR #4302 (stroke bounds) regenerates many of the same fixtures. Whichever of these two PRs merges second must merge main and re-run `GENERATE_SCENES=1` (with `TZ=America/Los_Angeles`) to refresh the goldens. Recommendation: merge #4302 first, then rebase/refresh this one.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3767, opened 2023-06-12
Add geoScale vega function to support altair_tiles

binste:
In [altair_tiles](https://github.com/altair-viz/altair_tiles) we are exploring how we can include (geo)tile-layers, such as openstreetmap basemaps, within Altair. The development is progressing well and the biggest road block we have so far is that we cannot access the `scale` value of a geo projection. Closes #3758 <- This issue explains why we need this and contains some example specs (cc @mattijn).

This PR introduces a `geoScale` vega function which fetches the named projection and calls `.scale()` on it. I named it `geoScale` to differentiate from the existing `scale` function.

I added a spec to the `test` folder which showcases how it can be used. This is somewhat the spec that we will produce with `altair_tiles` (run `yarn serve` to see it):

image

This is the first time I'm working on Vega so it's likely that I missed something. Happy to work more on this and push it over the finish line with some guidance.

Issue #3758, opened 2023-05-22
Support accessing projection scale function in top-level scope for (geo)-tiles integration


mattijn:
Context: In the package [altair_tiles](https://github.com/altair-viz/altair_tiles) we are exploring how we can include (geo)tile-layers, such as openstreetmap basemaps, within Altair. 
Currently this is possible if there is a `projection-scale` set upfront in the specification. 

But a more common case within the context of Vega-Lite is to use a `projection-fit` object. 
One of the question now is*, how to get the `projection-scale` value if this scale is internally derived within the projection? And especially, how to get this scale value as parameter/signal using Vega Expressions? 

An example:
Given the following Vega-Lite specification [Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAdxoATemgDMABh1L4NMlm16lapJhSpQDAE6C0IC1YLK2DtfwCM3nQikZeSUAMw9kTDRQTABPHDgnTDYcNmk5RRAQuEt7BPQoNgZZTDsaOAgQAF9KpRw7VLgoTHZ5G0y6KJBY+KcAMWzMXIyyODYEOBKYzu68kAAFNkEYsnSlAo81GllLcrQAbQOxbwUtAHYAXQU9o4V+LQAWS72fU9uHp5eTi6ubs-P-mogOrJOB2Zq7GzVJQzJzjOxQSweKq1JB2JAICqoPagbbjJwAWQAmgB9OYAJQA8gApDJwAAedScBRwMQAFAByYGSRrNOTsgCUVUuIGQdgA1tM4rMRmwIDAkD1QjRBI50MJjJgyGipkoICU2GLpdqMlBhDg0CUGHBqkA):
```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v5.json",
  "width": 300,
  "height": 300,
  "data": {
    "url": "data/world-110m.json",
    "format": {"type": "topojson", "feature": "countries"}
  },
  "projection": {
    "fit": {
      "type": "Feature",
      "geometry": {
        "type": "Polygon",
        "coordinates": [[[51, 37], [51, -34], [-17, -34], [-17, 37], [51, 37]]]
      },
      "properties": {}
    },
    "type": "mercator"
  },
[...]

PR #4315, opened 2026-07-27
test(vega): load basemap's topojson from local test data

jonathanzong:
fixing test flakiness in basemap, which was messing up CI in https://github.com/vega/vega/pull/4061.

claude says:

> basemap was the only one of the 116 valid specs that fetched its data from a third-party CDN at test time. When that request fails on a CI runner -- rate limiting, a transient outage -- vega-dataflow logs a warning and pulses an empty dataset rather than throwing, so the spec still renders, just with none of its 177 country shapes. The scene test then fails at the sceneEqual comparison with nothing in the output pointing at the network.
> 
> The suite only started comparing scenegraphs in #4305; before that it regenerated them, so this dependency never had a chance to bite.
> 
> vega-datasets' world-110m.json is already vendored in test/data and is byte for byte identical to the pinned CDN copy, so just load it from there like every other spec does. The golden is unchanged.
> 
> The OpenStreetMap tile URLs in the spec are left alone: they are built by a formula into image mark urls, and image marks only load through a renderer, so they are never fetched under renderer: 'none'.
> 
> With this the scene suite passes 115/115 with all outbound network calls blocked; without it, basemap fails.
>
> **Note**: A failed load stays invisible in these tests. If it bites again, the clean fix is probably raising the default log level for the whole suite rather than filtering individual messages.

### DIFF
--- a/packages/vega/test/specs-valid/basemap.vg.json
+++ b/packages/vega/test/specs-valid/basemap.vg.json
@@ -31,7 +31,7 @@
     },
     {
       "name": "source_0",
-      "url": "https://cdn.jsdelivr.net/npm/vega-datasets@v1.29.0/data/world-110m.json",
+      "url": "data/world-110m.json",
       "format": {"feature": "countries", "type": "topojson"},
       "transform": [
         {


joelostblom (approved):
Thank you! This looks goods to me so I'm going to to merge to unblock #4061.

I also asked codex to have a look at the git history and it couldn't find any particular reason for the introduction of the CDN instead of just the path:

- PR [#3767](https://github.com/vega/vega/pull/3767) initially added a primitive `geoScale` fixture using `data/world-110m.json`.
- Commit [`3c4ed688`](https://github.com/vega/vega/commit/3c4ed688ae69fe333e0fd4c1a400828a220f953f) replaced it with the “better test spec,” introducing the CDN URL.
- Historical `altair_tiles` (which was the motivating factor for the original PR) code constructs the source as:
  ```python
  alt.topo_feature(data.world_110m.url, "countries")
  ```
  `vega_datasets` expands `data.world_110m.url` to the pinned jsDelivr URL. The Vega fixture was effectively copied from that generated example.
- The motivating issue [#3758](https://github.com/vega/vega/issues/3758) itself used the local `data/world-110m.json` path.
- No review comment or commit message indicates that remote loading was required or deliberate.

domoritz:
Thank you

PR #4316, opened 2026-07-28
feat(vega-functions): add easing functions and interpolateLinear

jonathanzong:
This PR adds easing functions and interpolation to the expression language. These are needed for Animated Vega-Lite.

**Claude below:**

Adds easing functions and `interpolateLinear` to the expression language. Animated visualizations need an eased clock and piecewise interpolation, and no existing expression computes either operation.

### Easing functions

This PR registers the [d3-ease](https://github.com/d3/d3-ease) functions under their d3 names: `easeLinear`, `easeQuadIn/Out/InOut`, `easeCubic*`, `easePoly*`, `easeSin*`, `easeExp*`, `easeCircle*`, `easeBounce*`, `easeBack*`, `easeElastic*`.

Each function maps normalized time in `[0, 1]` onto an eased position in `[0, 1]`, so a spec can reshape an animation clock inside the expression language:

```js
easeCubicInOut(anim_clock / duration) * duration
```

The registration uses d3's defaults for the parametric families (`easePoly`, `easeBack`, `easeElastic`), because `.exponent()`, `.overshoot()`, and `.amplitude()` have no expression-language equivalent. The bare names alias d3's own defaults: `easeBounce` is `easeBounceOut` and `easeQuad` is `easeQuadInOut`. The aliases match d3 rather than inventing a convention.

Adds `d3-ease` (~2 kB, BSD-3-Clause) as a dependency of `vega-functions`.

### `interpolateLinear(array, fraction)`

Interpolates piecewise across every entry of an array rather than across its endpoints.

`lerp` already covers the endpoint case, but collapsing intermediate control points gives the wrong answer when the array describes a curve. `lerp([0, 100, 10], 0.5)` returns `5`, while the curve is at `100`. `interpolateLinear([0, 100, 10], 0.5)` returns `100`.

The two functions together let a spec define a custom easing curve from its own control points, and let a spec read any sampled series at an arbitrary position.

### Tests

`packages/vega-functions/test/ease-test.js` and `interpolate-test.js` — 207 assertions pass. The easing tests assert that every registered function maps the unit interval onto itself and is reachable from the function context. The interpolation tests cover segment interpolation, the divergence from `lerp`, clamping, and degenerate input.

### Notes

- Purely additive. No existing behavior changes.
- A companion Vega-Lite change adds `easing` to animated selections and consumes both functions. The interpolation function lives here rather than in Vega-Lite because it must run inside a signal expression.

### DIFF
--- a/packages/vega-functions/src/functions/ease.js
+++ b/packages/vega-functions/src/functions/ease.js
@@ -1,0 +1,61 @@
+import * as d3 from 'd3-ease';
+
+/**
+ * The d3-ease easing functions, exposed to the expression language under their
+ * d3 names. Each maps a normalized time in [0, 1] to an eased position in
+ * [0, 1], letting animations vary their playback rate over the time domain.
+ *
+ * The parametric families (`easePoly`, `easeBack`, `easeElastic`) are exposed
+ * at their default parameters only; d3's `.exponent()` / `.overshoot()` /
+ * `.amplitude()` configuration has no expression-language equivalent.
+ */
+export const easeFunctions = {
+  easeLinear: d3.easeLinear,
+
+  easeQuad: d3.easeQuad,
+  easeQuadIn: d3.easeQuadIn,
+  easeQuadOut: d3.easeQuadOut,
+  easeQuadInOut: d3.easeQuadInOut,
+
+  easeCubic: d3.easeCubic,
+  easeCubicIn: d3.easeCubicIn,
+  easeCubicOut: d3.easeCubicOut,
+  easeCubicInOut: d3.easeCubicInOut,
+
+  easePoly: d3.easePoly,
+  easePolyIn: d3.easePolyIn,
+  easePolyOut: d3.easePolyOut,
+  easePolyInOut: d3.easePolyInOut,
+
+  easeSin: d3.easeSin,
+  easeSinIn: d3.easeSinIn,
+  easeSinOut: d3.easeSinOut,
+  easeSinInOut: d3.easeSinInOut,
+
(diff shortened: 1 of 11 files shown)


jonathanzong:
~~failing unrelated basemap test~~ so will depend on https://github.com/vega/vega/pull/4315

EDIT: ready for review

packages/vega-functions/src/functions/ease.js
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * The d3-ease easing functions, exposed to the expression language under their
+ * d3 names. Each maps a normalized time in [0, 1] to an eased position in
+ * [0, 1], letting animations vary their playback rate over the time domain.
+ *
+ * The parametric families (`easePoly`, `easeBack`, `easeElastic`) are exposed
+ * at their default parameters only; d3's `.exponent()` / `.overshoot()` /
+ * `.amplitude()` configuration has no expression-language equivalent.
+ */
+export const easeFunctions = {
```
domoritz, line 12: It's a lot of new methods so I considered a single method with a string parameter. But that would make it inconsistent with other namespaces in Vega so let's go with the flat version.

domoritz (approved):

rollup.config.js
```diff
@@ -13,6 +13,7 @@ const d3Deps = [
   'd3-color',
   'd3-dispatch',
   'd3-dsv',
+  'd3-ease',
```
domoritz, line 16: This is needed to keep the bundle small when we have d3 loaded already.

PR #4317, opened 2026-07-28
fix: apply DisallowedObjectProperties to quoted object keys in codegen

EchoSkorJjj:
The ObjectExpression deny-list check read the key from prop.key.name, which is undefined for a quoted (Literal) key, so keys like {"toString": 1} skipped the DisallowedObjectProperties check that the unquoted {toString: 1} hits. Resolve the name from prop.key.value for Literal keys, which matches the interpreter
(interpret.js already evaluates the key). String() keeps numeric keys like {0: 1} working.

### DIFF
--- a/packages/vega-expression/src/codegen.js
+++ b/packages/vega-expression/src/codegen.js
@@ -100,7 +100,9 @@ export default function(opt) {
     ObjectExpression: n => {
       // If any keys would override Object prototype methods, throw error
       for (const prop of n.properties) {
-        const keyName = prop.key.name;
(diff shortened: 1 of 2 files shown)


hydrosquall:
@EchoSkorJjj thanks for the contribution. Could we please add a unit test ( packages/vega-expression/test ) that would show the benefit of including this condition ? Thank you!

EchoSkorJjj:
I have added the unit test.

domoritz (approved):

PR #4319, opened 2026-07-28
feat(a11y/i18n): Add ability to localize accessibility labels

yelper:
This PR adds the ability for Vega to take in a localization object that is a map of localization keys to translated phrases.  This allows Vega to emit aria-labels and aria-roleDescription content that is localized for the current users' locale.  

I anticipate this locale to be driven by the `lang` tag set in the parent HTML tag (usually how webpages initially realize what the home locale is for a user). The locale then drives the "languageTag" set on the `ariaLocale` object ([ISO 639](https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes) set 1; two letter) and would signify to the application to load the appropriate translations.   This is left as an exercise to the dev author of the calling Vega instance.  

Because this feels like a big interface change, I totally welcome feedback on the approach.  It feels like it would be hard to change post-hoc, so discussion would be great :)  If it's easier to chat synchronously about this, let me know.

### Entry point
I thought that adding localization to the initialization of the view made the most sense, so here are the entry points:

Set globally at the Vega level, like existing value formatting locale:

```javascript
vega.ariaLocale(spanishAriaLocale); // Record<string, string>
```

Set locally at the `View` level:
```javascript
// At view construction
const view = new vega.View(runtime, {
  ariaLocale: spanishAriaLocale 
});

// Or update dynamically
view.ariaLocale(spanishAriaLocale);
```

Vega already makes English the default language.  If no localization hints are provided, Vega will continue to emit a11y labels in English.  The `ariaLocale` property of Vega inherits the `DEFAULT_ARIA_LOCALE` set of translations.

### Translation strings

I tried to isolate common phrasing and language list comprehension into meaningful loc keys.  "listJoiner" and "listFinalJoiner" lets users customize how lists of items (e.g., how axis labels list out items) are composed.  Role-specific descriptions are prefixed with "role."  It may be possible to compose things further, but I'll leave it here for now...

The given localization key-value string map extends the default English translation.  If a translation is missing, then the original English translation will be used.

The changes try to take advantage of [Unicode CLDR](https://cldr.unicode.org/)), so I try to take advantage of [Intl.PluralRules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/PluralRules) for determining translations of zero/one/two/few/many/other.  These numerosity-sensitive translations will be suffixed as appropriate for that language.  If the number of items is greater than one, it'll try to use the appropriate key, but fall back to the "other" suffix if it fails.  If `Intl.PluralRules()` does not exist in the user-agent, the code will just use the "other" suffix key as appropriate

### Testing

I tried to follow the existing testing infrastructure to author unit tests.  I verified locally with a test harness (harness/index.html) where I translated all strings to Turkish and it seems to be okay.

## Other things

AI note: I used assistance from Copilot CLI, using Opus 4.6 to write the initial version of the code.  I then myself went through the proposed code and cleaned up some interfaces, styling, comments, and refactored a bit to make the code more concise.    The PR description is written wholly by me.


docs/_includes/properties.md
```diff
@@ -35,7 +35,7 @@ When using multiple spatial properties along the same dimension (_x_ and _x2_, o
 
 ### Accessibility Properties {% include tag ver="5.11" %}
 
-Accessibility properties are used to determine [ARIA (Accessible Rich Internet Applications) attributes](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA) when using Vega to render SVG output. Vega will automatically generate ARIA attributes for container SVG `<g>` elements for each mark definition; this can be configured using [mark-level accessibility properties](/docs/marks/#accessibility). Use the properties below in encode sets to add ARIA attributes for individual mark items.
+Accessibility properties are used to determine [ARIA (Accessible Rich Internet Applications) attributes](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA) when using Vega to render SVG output. Vega will automatically generate ARIA attributes for container SVG `<g>` elements for each mark definition; this can be configured using [mark-level accessibility properties]({{ site.baseurl }}/docs/marks/#accessibility). Use the properties below in encode sets to add ARIA attributes for individual mark items.
```
yelper, line 38: fixes a linking issue.  I tried `{% link <url> %}`, but nothing seemed to work other than explicitly including it...

PR #4334, opened 2026-08-10
fix(parser): ignore unknown scale spec keys

### LINKED ISSUE #4329: parseScale forwards unknown spec keys into the dataflow, causing spurious 'Unsupported scale property' warnings
`packages/vega-parser/src/parsers/scale.js` (around lines 50-53) copies *every* unrecognized key from a scale spec into the Scale operator's params. `packages/vega-encode/src/Scale.js` (lines 84-91) then warns for any param that is neither in its `SKIP` set nor a scale method.

The result is that any annotation key on a scale produces a runtime warning:

```
WARN: Unsupported scale property: description
WARN: Unsupported scale property: _description
```

This matters because the `_`-prefixed convention is the workaround currently recommended to users who want to annotate their specs (see discussion #4234) — and it is not runtime-clean on scales. Other object types are unaffected: transform parsing iterates only `def.params`, and data parsing uses fixed key reads.

Adding individual names to the `SKIP` set treats the symptom. The underlying question is whether `parseScale` should forward unknown keys into the dataflow at all, or filter to known scale properties at parse time.

Found while reviewing #4314.

cyphercodes:
Fixes #4329.

## Description

Filters scale specifications against known scale properties before forwarding parameters into the runtime scale operator. This prevents extension or unknown spec keys from reaching `Scale` and emitting spurious `Unsupported scale property` warnings.

## Testing

- `npx tape packages/vega-parser/test/scale-test.js`
- `npx lerna run build --scope vega-transforms --include-dependencies`
- `npm test --workspace vega-parser`
- `git diff --check`

### DIFF
--- a/packages/vega-parser/src/parsers/scale.js
+++ b/packages/vega-parser/src/parsers/scale.js
@@ -5,12 +5,19 @@ import {aggrField, isSignal, keyFieldRef, ref} from '../util.js';
 
 import {isDiscrete, isQuantile, isValidScaleType} from 'vega-scale';
 import {
-  error, extend, hasOwnProperty, isArray, isObject, isString, stringValue
+  error, extend, hasOwnProperty, isArray, isObject, isString, stringValue, toSet
 } from 'vega-util';
 
 let FIELD_REF_ID = 0;
 
 const MULTIDOMAIN_SORT_OPS  = {min: 'min', max: 'max', count: 'sum'};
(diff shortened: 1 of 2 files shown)


domoritz:
Thanks for the pull request. I was in the middle of a larger triaging session and not all info was in the issue. I'll need to approach this differently.

PR #4345, opened 2026-09-06
perf: Optimize Symbol Bounds Calculation

KaNaDaAT:
While developing the WebGPU renderer ([vega-webgpu](https://github.com/KaNaDaAT/vega-webgpu)), I noticed that in some benchmarks a large share of the computing cost comes from calculating symbol bounds rather than the actual rendering.

Every symbol draws the same design, just scaled by `sqrt(size) / 2`. This means we can measure the design's bounding box once and scale it for each symbol, instead of replaying the path for every item on every frame.

Here are some of the results from the benchmarks:

| Spec | Symbols | Before | After | Improvement |
|---|---:|---:|---:|---:|
| splom-outer-50k | 800,003 | 144 ms | 16 ms | **9.0x** |
| splom-outer | 5,475 | 1.10 ms | 0.22 ms | **5.0x** |
| scatter-plot-panzoom | 396 | 0.14 ms | 0.04 ms | — |
| contour-scatter | 392 | 0.14 ms | 0.08 ms | — |

I also tested this with the [benchmark animation](https://kanadaat.github.io/vega-webgpu/test/?spec=benchmark&renderer=webgpu), where the number of symbols can be changed interactively:

| Symbols | Frame | Vega dataflow | Our `renderAsync` |
|---:|---:|---:|---:|
| 10k | 4.5 ms | 3.5 ms | 1.0 ms |
| 50k | 17.7 ms | 15.8 ms | 1.9 ms |
| 100k | 34.4 ms | 31.9 ms | 2.5 ms |
| 300k | 96.4 ms | 89.4 ms | 7.0 ms |

Rotated symbols are a little more complicated since their axis-aligned bounding box changes with the angle, so their bounds still need to be recalculated when the rotation changes.
Maybe we can find a way to improve this as well?

### DIFF
--- a/packages/vega-scenegraph/src/marks/symbol.js
+++ b/packages/vega-scenegraph/src/marks/symbol.js
@@ -1,5 +1,28 @@
-import {symbol} from '../path/shapes.js';
-import {intersectPoint} from '../util/intersect.js';
-import markItemPath from './markItemPath.js';
+const unitBounds = Object.create(null);
 
-export default markItemPath('symbol', symbol, intersectPoint);
+function unitBox(shape) {
+  let box = unitBounds[shape];
+
+  if (box === undefined) {
+    // Size 4 gives a scale of 1.
+    const b = new Bounds();
+    symbol(boundContext(b, 0), {shape, size: 4});
+    box = unitBounds[shape] = [b.x1, b.y1, b.x2, b.y2];
+  }
+
+  return box;
+}
+
+function bound(bounds, item) {
+  if (item.angle) {
+    // Rotated shapes need their bounds recalculated.
+    symbol(boundContext(bounds, item.angle), item);
+  } else {
+    const u = unitBox(item.shape || 'circle'),
+          r = Math.sqrt(item.size != null ? item.size : 64) / 2;
+
+    bounds.set(u[0] * r, u[1] * r, u[2] * r, u[3] * r);
+  }
+
+  return boundStroke(bounds, item, true).translate(item.x || 0, item.y || 0);
+}
--- a/packages/vega-scenegraph/src/marks/symbol.js
+++ b/packages/vega-scenegraph/src/marks/symbol.js
@@ -1,12 +1,21 @@
+import { symbol } from '../path/shapes.js';
+import { intersectPoint } from '../util/intersect.js';
+import markItemPath from './markItemPath.js';
+
 const unitBounds = Object.create(null);
+
+/** Size used to construct unit bounds, resulting in a scale factor of 1. */
+const UNIT_SYMBOL_SIZE = 4;
+
+/** Fallback symbol size used when no size is specified. Results in radius of 4. */
+const FALLBACK_SYMBOL_SIZE = 64;
 
 function unitBox(shape) {
   let box = unitBounds[shape];
 
   if (box === undefined) {
-    // Size 4 gives a scale of 1.
     const b = new Bounds();
-    symbol(boundContext(b, 0), {shape, size: 4});
+    symbol(boundContext(b, 0), { shape, size: UNIT_SYMBOL_SIZE });
     box = unitBounds[shape] = [b.x1, b.y1, b.x2, b.y2];
   }
 
@@ -18,8 +27,9 @@ function bound(bounds, item) {
     // Rotated shapes need their bounds recalculated.
     symbol(boundContext(bounds, item.angle), item);
   } else {
-    const u = unitBox(item.shape || 'circle'),
-          r = Math.sqrt(item.size != null ? item.size : 64) / 2;
+    const u = unitBox(item.shape || 'circle');
+    const size = item.size ?? FALLBACK_SYMBOL_SIZE;
+    const r = Math.sqrt(size) / 2;
 
     bounds.set(u[0] * r, u[1] * r, u[2] * r, u[3] * r);
   }
--- a/packages/vega-scenegraph/src/marks/symbol.js
+++ b/packages/vega-scenegraph/src/marks/symbol.js
@@ -7,15 +7,12 @@ import markItemPath from './markItemPath.js';
 
 const unitBounds = Object.create(null);
 
-/** Size used to construct unit bounds, resulting in a scale factor of 1. */
-const UNIT_SYMBOL_SIZE = 4;
-
 function unitBox(shape) {
   let box = unitBounds[shape];
 
   if (box === undefined) {
     const b = new Bounds();
-    symbol(boundContext(b, 0), {shape, size: UNIT_SYMBOL_SIZE});
+    symbol(boundContext(b, 0), {shape, size: 4}); // sqrt(4) / 2 = 1, a unit box
     box = unitBounds[shape] = [b.x1, b.y1, b.x2, b.y2];
   }
 
--- a/packages/vega-scenegraph/src/marks/symbol.js
+++ b/packages/vega-scenegraph/src/marks/symbol.js
@@ -1,7 +1,7 @@
 import Bounds from '../Bounds.js';
 import boundContext from '../bound/boundContext.js';
 import boundStroke from '../bound/boundStroke.js';
-import {symbol} from '../path/shapes.js';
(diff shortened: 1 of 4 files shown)


PBI-David:
My comment has nothing to do with the PR itself but that benchmark page is very cool. Just in case you weren't aware, the mouse over events for the Wordcloud webgpu 2.0.0.0 example are very delayed in Firefox (less so on Chrome). Looking forward to seeing this merged.

KaNaDaAT:
> My comment has nothing to do with the PR itself but that benchmark page is very cool. Just in case you weren't aware, the mouse over events for the Wordcloud webgpu 2.0.0.0 example are very delayed in Firefox (less so on Chrome). Looking forward to seeing this merged.

Thank you very much and yes thats a general problem with the texts rendered to offscreen textures.
I laready changed it to text glyphs and it got way better. All local still but working on it.

packages/vega-scenegraph/src/marks/symbol.js
```diff
@@ -1,5 +1,28 @@
...
+  return box;
+}
+
+function bound(bounds, item) {
+  if (item.angle) {
+    // Rotated shapes need their bounds recalculated.
+    symbol(boundContext(bounds, item.angle), item);
+  } else {
+    const u = unitBox(item.shape || 'circle'),
+          r = Math.sqrt(item.size != null ? item.size : 64) / 2;
```
lsh: nit: can we document why 64 is here?

domoritz:
This is awesome. Just checking, does this also work well when the size changes across frames?

lsh:
This is very cool! Great work! I pinged @domoritz for review since I can't review today.

packages/vega-scenegraph/src/marks/symbol.js
KaNaDaAT: Fair point.

KaNaDaAT:
> This is very cool! Great work! I pinged @domoritz for review since I can't review today.

Should work fine as this only caches for shape and then the bounding box is scaled up based on the items size.
Only problem I could imagine if the underlying shape of a cached entry changes over time but I dont think that should ever happen? 
If so this performance gain would be difficult to achive

packages/vega-scenegraph/src/marks/symbol.js
KaNaDaAT: Is there a way to get the Config.symbol.size? 
I only found https://vega.github.io/vega/docs/config which is why I took 64.
But I am unsure wether I can access the config from here?

packages/vega-scenegraph/src/marks/symbol.js
```diff
@@ -1,5 +1,46 @@
...
 import markItemPath from './markItemPath.js';
 
-export default markItemPath('symbol', symbol, intersectPoint);
+const unitBounds = Object.create(null);
+
+/** Size used to construct unit bounds, resulting in a scale factor of 1. */
+const UNIT_SYMBOL_SIZE = 4;
+
+/** Fallback symbol size used when no size is specified. Results in radius of 4. */
+const FALLBACK_SYMBOL_SIZE = 64;
```
domoritz: We need to re-use the existing size. `defaultSymbolSize` is 30 (defined here https://github.com/vega/vega/blob/31abf45a179035193263693c82399b3484bc21b5/packages/vega-parser/src/config.js#L8)

packages/vega-scenegraph/src/marks/symbol.js
```diff
@@ -1,5 +1,46 @@
...
+  return box;
+}
+
+function bound(bounds, item) {
+  if (item.angle) {
+    // Rotated shapes need their bounds recalculated.
+    symbol(boundContext(bounds, item.angle), item);
+  } else {
+    const u = unitBox(item.shape || 'circle');
+    const size = item.size ?? FALLBACK_SYMBOL_SIZE;
```
domoritz: I don't think this is correct. The symbol size falls back to the config, no?

packages/vega-scenegraph/src/marks/symbol.js
```diff
@@ -1,5 +1,46 @@
...
+
+    bounds.set(u[0] * r, u[1] * r, u[2] * r, u[3] * r);
+  }
+
+  return boundStroke(bounds, item, true).translate(item.x || 0, item.y || 0);
+}
+
+export default {
+  ...markItemPath('symbol', symbol, intersectPoint),
+  bound
```
domoritz: Does this feel like a hack? What if we improved `markItemPath`? https://github.com/vega/vega/blob/31abf45a179035193263693c82399b3484bc21b5/packages/vega-scenegraph/src/marks/markItemPath.js#L16

domoritz (changes requested):
Perf improvements are great but I think the code needs at least another pass.

domoritz:
Actually, I was wrong. We don't need to pass in the config. The symbol size should already be set by the parser. But we should re-use the fallback. 

In packages/vega-scenegraph/src/path/shapes.js:

```ts
export const DefaultSymbolSize = 64;

const ...
      sz = item => value(item.size, DefaultSymbolSize),
```

In packages/vega-scenegraph/src/marks/symbol.js:

```ts
import {DefaultSymbolSize, symbol} from '../path/shapes.js';

// delete the FALLBACK_SYMBOL_SIZE

const size = item.size ?? DefaultSymbolSize;
```

packages/vega-scenegraph/src/marks/symbol.js
```diff
@@ -1,5 +1,46 @@
...
+    symbol(boundContext(b, 0), { shape, size: UNIT_SYMBOL_SIZE });
+    box = unitBounds[shape] = [b.x1, b.y1, b.x2, b.y2];
+  }
+
+  return box;
+}
+
+function bound(bounds, item) {
+  if (item.angle) {
+    // Rotated shapes need their bounds recalculated.
```
domoritz: That feels like an obvious comment describing the code.

packages/vega-scenegraph/src/marks/symbol.js
```diff
@@ -1,5 +1,46 @@
...
+    const b = new Bounds();
+    symbol(boundContext(b, 0), { shape, size: UNIT_SYMBOL_SIZE });
+    box = unitBounds[shape] = [b.x1, b.y1, b.x2, b.y2];
+  }
+
+  return box;
+}
+
+function bound(bounds, item) {
+  if (item.angle) {
```
domoritz: Could probably not go into this condition for circles.

packages/vega-scenegraph/src/marks/symbol.js
domoritz: Ignore. See below.

packages/vega-scenegraph/src/marks/symbol.js
KaNaDaAT: Agree. I was unsure about the guidlines. Will remove

packages/vega-scenegraph/src/marks/symbol.js
KaNaDaAT: Ye I though so as well. Then I figured if a user spins circles, that's kind of their own business, but with animations and nested elements, that can probably happen quickly without their knowledge. So something to work on as its an easy fix

KaNaDaAT:
I think another way to improve performance for the rotated versions would be to cache the shape and rotate it based on the vertices. That way, the shape doesn't have to be recalculated, which should be a bit faster. If all the symbols had the same rotation, that would be even better, since we could then create a small circular buffer (if two angles alternate, for example).

Measured locally on 200k items:

| case | before | after |
|---|---:|---:|
| constant angle | 11.9 ms | 1.8 ms (6.7x) |
| cross, angle per item | 31.2 ms | 8.6 ms (3.6x) |
| all distinct angles | 13.3 ms | 6.2 ms (2.1x) |

May be worth to commit as well at some point but for now I think that would get a little out of hand, and if a user really needs that performance, the could just rotate the symbol themselves as custom symbols? (wouldnt work for animations ofcourse but who animates thousands of symbols anyways)

domoritz (approved):

domoritz:
Thank you!

